PR #3, opened 2015-05-06
add blacklist function

antop:


lucaderi:
@antop This code cannot be merged automatically anymore and it is still missing some cleanup. Please do that and send me a new merge request.

PR #6, opened 2015-05-19
Added QUIC protocol

backer-and:


lucaderi:
@backer-and See https://github.com/ntop/nDPI/commit/8f3b8ff4dfb5b7674b6d87c78c1bdb275f178d4d as your code was generating false positives. Looking at the specs we can reduce the QUIC scope to UDP ports 80/443

PR #9, opened 2015-05-21
Introduced incremental versioning from git tag and -r option to ndpiReader to print revision

lucagiove:
Version is now extracted using `git describe --tags --long --dirty --always` this uses the last git tag as version then an incremental number that reflects the number of commits (like svn revision) plus the commit hash:

```
1.5.2-43-gb3a292d-dirty
^     ^  ^        ^
|     |  |        |
|     |  |        if there are uncommitted changes
|     |  SHA of HEAD
|     number of commits since last tag
last tag
```
This allows to have incremental versioning changing at each build like it was before with svn.

Also added a `-r` to print ndpiReader / nDPI version and revision.
(I would have rather preferred `-V` or `--version` that is standard but I didn't want to change too much the code) 

```
ntop@bs-ndpi:~/target/bin$ ./ndpiReader -r
ndpiReader, nDPI (1.5.2-45-gb3b0ce7)
```

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -10,32 +10,31 @@ AC_PROG_CC
 AX_PTHREAD
 
 if test -d ".git"; then :
-GIT_TAG=`git rev-parse HEAD`
-GIT_DATE=`date +%Y%m%d`
-GIT_RELEASE="${PACKAGE_VERSION} (${GIT_TAG}:${GIT_DATE})"
+     GIT_TAG=`git log -1 --format=%H`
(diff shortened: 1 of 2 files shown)


lucagiove:
Great ;) consider that you need to add at least one tag.
On my branch I did this:
```
git tag 1.5.2 2e5ceac844c32fb52f4f3042be5b872f8b0b4ff0
git push --tags
```

lucaderi:
I am patching the code. Hold on.

PR #26, opened 2015-06-14
added initial brand-new api rndpi

rcarbone:
I have an initial implementation for a new API to nDPI because the existing does not match my preferences.

So I started hacking at code to have a new access to the nDPI protocol dissectors. I put the code in the the rndpi/ directory at the top level as a stand-alone contrib not yet full integrated with the rest (above all in terms of Makefiles).

My code compiles on Linux running Debian/RedHat and I not currently interested in porting on other platforms until the API has been accepted.

If you think it is acceptable please push in the dev  nDPI base.

/rocco


lucaderi:
Unfortunately this merge request can't be merged automatically due to conflicts. This said I believe that we should merge changes only relative to nDPI. Things on top of the library should stay elsewhere in order to keep projects limited in scope.

PR #29, opened 2015-06-17
Allow Host IP based subprotocols to be added in protos.txt

commsquare:
These changes allow to dynamically add new subprotocols that depend on the
host IP, similarly to the host-name based subprotocols.
```
#  IP based Subprotocols
#  Format:
#  ip:<value>,ip:<value>,.....@<subproto>

ip:213.75.170.11@CustomProtocolName
```


lucaderi:
Thank you!

PR #30, opened 2015-06-17
Adding more Twitch.tv support

sparkon:
Now able to recognize when someone is streaming by matching packet endpoints with Twitch.tv ingestion servers. Added two hosts that are used for static content ( JustinTV CDN ) such as chat images ( emoticons ) and scripts.


lucaderi:
Thank you!

PR #55, opened 2015-07-03
Fixed double version bug for ndpiReader -r

lucagiove:
Removed the branch name from the ndpiReader version otherwise this is what happen if you build from branch 1.6-stable:

ntop@bs-ndpi:~$ ndpiReader -r
ndpiReader - nDPI (1.6.0-1.6-stable-62-a31b0a3)

now it's:
ntop@bs-ndpi:~$ ndpiReader -r
ndpiReader - nDPI (1.6.0-63-7943e6e)

This is also e problem for buildsystems that checkout a specific commit not the whole branch, this is what happens:

root@e-30-ndpi:~# ndpiReader -r
ndpiReader - nDPI (1.6.0-HEAD-63-7943e6e)

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -17,8 +17,7 @@ if test -d ".git"; then :
      # 
      #
      GIT_NUM=`git log --pretty=oneline | wc -l`
-     GIT_BRANCH=`git rev-parse --abbrev-ref HEAD`
-     GIT_RELEASE="${PACKAGE_VERSION}-${GIT_BRANCH}-${GIT_NUM}-${GIT_TAG}"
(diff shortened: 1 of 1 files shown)


lucaderi:
Unfortunately this information is required by additional application that depend on it. I would like to thus keep the branch name

lucagiove:
Okay it would work if the breach name is not the version.
On our side we could always check out the commit in order to have HEAD.
Thanks for the answer.

Il giorno mer 8 lug 2015 19:26 Luca Deri notifications@github.com ha
scritto:

PR #58, opened 2015-07-07
Remove wrong space on Whatsapp Voice protocol name

pr0gg3d:
Whitespace break iptables protocol name so you won't able to specify it as protocol since WhatsApp already exists.


lucaderi:
This problem has already been fixed by a previous fix.

PR #66, opened 2015-07-17
Starcraft 2/Battle.net support

c4stan:
In this request:
-Battle.net hostnames added to the list
-Starcraft 2 specific dissector added to the supported protocols
and everything else is related, in order to make these two work inside nDPI.


lucaderi:
Thanks

PR #72, opened 2015-07-29
Enhance NTP support, add protocol version identification; Add pcap ex…

pavel-odintsov:
Hello, ntop team!

Please look on my patch and review it. I have removed check for packet size because monlist packets exceed this size. 

Also I have added information about protocol version and request type.


lucaderi:
Thank you

pavel-odintsov:
Thank you! :)

PR #73, opened 2015-07-30
battle.net removed

c4stan:
battle.net host identification removed, fixed a typo in ndpi_protocols.h about starcraft


lucaderi:
Thank you

PR #91, opened 2015-08-28
Make sure packet->user_agent_line.ptr is up to date before using it.

rs-natano:
Otherwise we might dereference a stale pointer, resulting in a segfault.

1.7 is affected. I didn't test 1.6, but from reading the code it looks like it is affected too.


lucaderi:
Thank you

PR #96, opened 2015-09-22
Remove duplicated function declarations

mgigante:
Double declaration of two functions at lines 85-90 and 146-151


lucaderi:
Thank you

PR #98, opened 2015-10-05
Add Ubiquity AirControl 2 protocol detector

Tomasu:
Supports detecting AirControl 2 (at the very least, no specs for any of the ubiquity protocols that I could find, and I only have dumps for one specific announce/broadcast packet) packets.


lucaderi:
Thank you. I will fix a few things (e.g. you cloned dhcp.c and not cleaned up)

Tomasu:
Ah, yeah, I forgot to change the copyright.. and left in an extra log. sorry.

PR #101, opened 2015-10-08
NDPI_PROTOCOL_KAKAOTALK_VOICE was used instead of NDPI_SERVICE_KAKAOT…

Tomasu:
kakaotalk_voice dissector uses the wrong define. It's in a LOG function so it isn't seen unless NDPI_ENABLE_DEBUG_MESSAGES is set.


kYroL01:
You are right. Thank you.

PR #105, opened 2015-10-09
ndpi_search_dns: fix potential buffer overflow

vel21ripn:
If j == sizeof(flow->host_server_name) then snprintf() generates the warning that the size is greater MAX_INT.


lucaderi:
Thank you

Issue #22, opened 2015-06-08
Some of nDPI filters do not match traffic


Mile-Lile:
I am ddwrt user and some time ago ddwrt developers drop using l7 filters and started using nDPI with iptables (for marking-matching traffic for QoS and Access Restriction purposes). I am not developer nor the network guru. Just simple SOHO user with very basic understanding of routers.
ddwrt has big community and now we have problem because some nDPI protocols (filters) do not work. ddwrt devs says it is not ddwrt problem. Youtube, Steam do not work... my findings about this problem are here in bug report on ddwrt TRAC http://svn.dd-wrt.com/ticket/4117#comment:7
If I can provide more informations I would be happy, just to solve this issue. Thx for your work on nDPI it is already great!

lucaderi:
@Mile-Lile Can you please attach a pcap file that I can use to reproduce the bug?

Mile-Lile:
I can run Wireshark on Windows machine and get you pcap but I think it wont help much because nDPI is implemented and running as module on ddwrt router (my win machine is behind that router)... it is embedded system and there is no flashspace for tcpdump (at least on my router, no space for optware - cheap low end router)... Recently I found out that ddwrt nDPI work is based on this "distro" of nDPI http://devel.aanet.ru/ndpi/

Mile-Lile:
ok. here is pcap. You will see (line 660) that I successfully connected to youtube (tls1) but skype after upgrade of app failed to connect and did retransmision all the time... file is 12,4MB so it is on gdrive for download https://drive.google.com/file/d/0B0SCwy1irn3qR0VKVlJRMGRBMUE/view?usp=sharing
here is ddwrt firewall where you can see that skype and youtube filters are loaded but aperently only skype is working http://www.dodaj.rs/f/3q/74/VkSNqDL/ddwrtfirewall.png

Mile-Lile:
File is to big for upload 12.4MB and only 10MB is allowed on git so here is my googledrive where you can download it  <https://drive.google.com/file/d/0B0SCwy1irn3qR0VKVlJRMGRBMUE/view?usp=sharing> https://drive.google.com/file/d/0B0SCwy1irn3qR0VKVlJRMGRBMUE/view?usp=sharing

lucaderi:
[...]

PR #114, opened 2015-11-12
Refactoring nDPI

kYroL01:
nDPI changes:
- unique ndpi_typedef.h for the typedef used;
- deleted kernel support
- deleted useless defines

### DIFF
--- a/example/ndpiReader.c
+++ b/example/ndpiReader.c
@@ -42,16 +42,14 @@
 #include <pcap.h>
 #include <signal.h>
 #include <pthread.h>
+#include <sys/socket.h>
 
 #include "../config.h"
+#include "ndpi_api.h"
 
 #ifdef HAVE_JSON_C
 #include <json.h>
 #endif
-
-#include "ndpi_api.h"
-
-#include <sys/socket.h>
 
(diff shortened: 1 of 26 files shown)


vel21ripn:
Removing kernel mode support is bad idea!
ndpi-netfilter is working.

lucaderi:
I understand your concern but in https://github.com/ntop/nDPI/issues/22 we asked many times to send us a merge request for the net filter code. This didn't happen so we prefer to keep the code tidy and clean. However if you want to share the netfilter code with us, we can accept your contribution.

vel21ripn:
I have branch ndpi-netfilter https://github.com/vel21ripn/nDPI/tree/netfilter (based on 1.7-stable).
Please look at it and express the opinion about the possibility of including it in the main project.
I made a lot of changes.
In July and early August, I was on vacation and could not respond to the request #22

lucaderi:
Please update with respect to our current tree, and send me a diff so we can check this work

vel21ripn:
OK, the difference between the 1.7-stable branches and the current is not very big. I'll try to do it in two days.  Please read the README in the directory ndpi-netfilter my branch.

PR #132, opened 2015-12-28
Made ndpi memory allocator type compatible with C++ on x686 platforms

pavel-odintsov:
Hello!

nDPI uses unsigned long as allocation functions parameter. It's not recommended way because it's not platform independent. ISO / C++ standard uses size_t for this case.

So I decided to change all "unsigned long" to size_t. 

For C++ case it could help critical error about type mismatch on 32 bit platforms (because we could not use 8 byte counter here):

``` bash

g++ ndpi_example.cpp  -I/opt/ndpi/include/libndpi-1.7.1  -L/opt/ndpi/lib -lndpi  
ndpi_example.cpp: In function ‘ndpi_detection_module_struct* init_ndpi()’:
ndpi_example.cpp:40:89: error: invalid conversion from ‘void* (*)(size_t)throw () {aka void* (*)(unsigned int)throw ()}’ to ‘void* (*)(long unsigned int)’ [-fpermissive]
In file included from ndpi_example.cpp:3:0:
/opt/ndpi/include/libndpi-1.7.1/libndpi/ndpi_api.h:77:40: error:   initializing argument 2 of ‘ndpi_detection_module_struct* ndpi_init_detection_module(u_int32_t, void* (*)(long unsigned int), void (*)(void*), ndpi_debug_function_ptr)’ [-fpermissive]
```

Thanks!

### DIFF
--- a/example/ndpiReader.c
+++ b/example/ndpiReader.c
@@ -407,7 +407,7 @@ static void debug_printf(u_int32_t protocol, void *id_struct,
 
 /* ***************************************************** */
 
-static void *malloc_wrapper(unsigned long size) {
+static void *malloc_wrapper(size_t size) {
   current_ndpi_memory += size;
 
   if(current_ndpi_memory > max_ndpi_memory)
--- a/src/include/ndpi_api.h
+++ b/src/include/ndpi_api.h
@@ -45,8 +45,8 @@ extern "C" {
   u_int32_t ndpi_detection_get_sizeof_ndpi_id_struct(void);
 
   /* Public malloc/free */
-  void* ndpi_malloc(unsigned long size);
-  void* ndpi_calloc(unsigned long count, unsigned long size);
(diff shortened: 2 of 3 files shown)


kYroL01:
@lucaderi I think is correct: `size_t` is much more portable and avoid any exceptions caused by the different size of types.
For example: http://stackoverflow.com/questions/15637228/what-is-the-downside-of-replacing-size-t-with-unsigned-long

What do you think ?

pavel-odintsov:
I'm also like this idea because for non standard platforms PPS, ARM it could make sense.

kYroL01:
I'm agree with you but I prefer to wait @lucaderi opinion before merge it ;)

lucaderi:
@kYroL01 @pavel-odintsov I think the change is a good idea. As far as I am concerned you can merge it Michele.

pavel-odintsov:
Thanks a lot!

PR #153, opened 2016-03-02
protocols: sip CANCEL packet detection fix

Timothyyy:
I noticed, that you have an error in sip CANCEL packet detection: you need to compare 4 chars instead 7 in 124 line "memcmp(&packet_payload[4], "SIP:", 7)".


kYroL01:
You're right.
Thank you!

PR #161, opened 2016-03-20
Add CoAP and Mqtt (IoT specific protocols) dissectors

srzamfir:
Hi,
I've included two new protocol dissectors specific for IoT:
- CoAP based on UDP (very similar with HTTP)
- Mqtt based on TCP (lightweight message queing protocol)

I've also included a test pcap file in /tests/pcap

Please let me know what you think.


kYroL01:
Interesting work. Thank you!

Unfortunately I notice some bugs in Mqtt dissector because now, after your commit, many http and skype flows are detected as Mqtt. I'll fix this for the moment. A check is needed.

lucaderi:
@srzamfir Thank you too!

Issue #164, opened 2016-03-26
is there any relation in coAP and BT?


JeevanNailwal:
Hi All,
I am looking to avoid BT data in my network. I was testing latest DEV branch and found the below result:

Detected protocols:
    Unknown              packets: 161           bytes: 16355         flows: 64  
    DNS                  packets: 19            bytes: 1793          flows: 9  
    HTTP                 packets: 75            bytes: 16067         flows: 6  
    SSDP                 packets: 2             bytes: 540           flows: 2  
    COAP                 packets: 7994          bytes: 5861897       flows: 268  
    BitTorrent           packets: 191           bytes: 29158         flows: 68  
    ICMP                 packets: 6             bytes: 510           flows: 5  
    SSL                  packets: 65            bytes: 18640         flows: 3  
    HTTP_Proxy           packets: 3             bytes: 202           flows: 1    

I had uploaded the relevent pcap file at:
https://drive.google.com/open?id=0B4M_moAIGLMSZUVFN0cwUF8yMnc

I am not sure if this recently aadded CoAP protocol and BT are linked, Can some one please provide information on same? for my experiment, Do I need to avoid CoAP packets too? Will it impact my other services?

Regards,
Jeevan

kYroL01:
There isn't direct relation between coap and BT.
As explain here, [COAP](https://en.wikipedia.org/wiki/Constrained_Application_Protocol) "is an application layer protocol that is intended for use in resource-constrained internet devices, such as WSN nodes. CoAP is designed to easily translate to HTTP for simplified integration with the web, while also meeting specialized requirements such as multicast support, very low overhead, and simplicity".

I need to see your pcap (please accept my request to download the pcap) to understand better.

In your session you captured, the main application is BT ? Try to provide a pcap with only a BT session.

JeevanNailwal:
Hi Michele,
The provided pcap has only session with BT. the o/p of ndpiReader output
was pasted in my query. I am pasting the same again, for your reference.

Detected protocols:
Unknown packets: 161 bytes: 16355 flows: 64

DNS packets: 19 bytes: 1793 flows: 9
HTTP packets: 75 bytes: 16067 flows: 6
SSDP packets: 2 bytes: 540 flows: 2
COAP packets: 7994 bytes: 5861897 flows: 268
BitTorrent packets: 191 bytes: 29158 flows: 68
ICMP packets: 6 bytes: 510 flows: 5
SSL packets: 65 bytes: 18640 flows: 3
[...]

PR #170, opened 2016-04-03
Mqtt dissector fix

srzamfir:
Fixed the Mqtt dissector. No more false positives on skype, kakaotalk and http.
However there are some issues with CoAP :

@kYroL01 : You have modified the CoAP dissector and also the Dropbox one in this commit a594273.
Please consider reverting the Dropbox changes since the actual  coap_mqtt.pcap doesn't contain any Dropbox traffic and thus the coap_mqtt.pcap test will fail. I've uploaded the correct test results (the pcap trace contains only traffic between 4 CoAP clients and 4 Mqtt clients - handcrafted by me - that generate traffic according to a predefined scenario so 800 CoAP packets is correct !)

Also, please do not rely solely on Wireshark for correct identification because it only considers the ports (in this specific test case the ports were randomly chosen by me (17500, 17501) to actually test nDPIs deep packet inspection capabilities for identifying CoAP and MQTT traffic that doesn't run on standard ports).

Cheers,
Sorin


kYroL01:
@srzamfir Ok I fix the Dropbox as u explain. The problem is that your CoAP dissector created false positive for BitTorrent as explain in this issue https://github.com/ntop/nDPI/issues/164.
So I changed to fix this bug.
I also did test with a Dropbox pcap sniffing traffic from my host and everything was detected as CoAp, wrongly.
Now I check your pull request, thanks.

srzamfir:
@kYroL01 If the CoAP dissector creates false positives with other traces this means it requires further fine-tuning. I'll also take a look at the mentioned bug.

kYroL01:
The fix I made solved the problem for BT. 
Now I'm testing your pull request before merge.

kYroL01:
I fixed Dropbox, tested your pull request and I think is correct.
Thank you

PR #177, opened 2016-04-15
Support RX protocol (used by AFS).

giomasce:
RX protocol is an RPC layer with transmission control used by AFS (Andrew File System), an old-but-still-used network file system. Here it is detected by comparing the numbers appearing in the first packet for every direction. The (somewhat informal specs of the protocol is here: http://web.mit.edu/kolya/afs/rx/rx-spec).

As noted in the code, in theory one can also recover the particular AFS service being used by looking at the `service_id` number in the header, but I do not know the best way to report it.


kYroL01:
@giomasce going to test and let you know, thanks!

kYroL01:
@giomasce I think is correct but i have to check a thing for a possible minor fix.
For now I merge, thanks!

PR #185, opened 2016-05-03
Fix test suite

alagoutte:
May be need to have .travis for check ? ;)


lucaderi:
@alagoutte I have merged this code even though I am not sure 100% is correct as the IPv6 formatting is different on some platforms and thus the diff (used by tests) won't work. If you want to use a travis fix, please go ahead. Please let me know if you want me to add to the nDPI developers group.

alagoutte:
I have forget to enable .travis for my nDPI repo

Mile-Lile:
capture

it is different formatting on Linux and MS platforms...

kYroL01:
Yes @Mile-Lile , you're right. @lucaderi this is why we have that different output for tests. We have to find a solution, IMHO.
Maybe this can help: https://tools.ietf.org/html/rfc5952#section-2
Looking at some rules, I think we can might catch the IPv6 and create an own function to convert it in a unique standard format for nDPI. What do u think ?

alagoutte:
@kYroL01 sed the output of test suite ?

kYroL01:
@alagoutte sorry, I don't understand what do u mean.

alagoutte:
use sed command (on another tools) for replace :0: by :: for example ?

kYroL01:
Yes, is a possible solution.

alagoutte:
May be open a issue for follow this problem ? (it is possible to create directly a issue from PR ?)

PR #187, opened 2016-05-09
Add symbols needed by ntopng to the shared library export list.

madpilot78:
The export list for the shared library seems to lack a pair of function names which are actually used by ntopng.

This fixes compiling ntopng to use ndpi as a shared library.

Also looks like this is the underlying cause for issues ntop/ntopng#528 and ntop/ntopng#540.


kYroL01:
Thanks

PR #188, opened 2016-05-10
Enhance QUIC support

alagoutte:
First patch to remove trailing whitespace

And second patch for remove wrong QUIC heuritics (make same false positive with NTP for example)
I have also update the test suite (but don't understood, why there is a change on starcraft_battle.pcap.out
(Detected Google...)

i will push soon some other enhance for add support of new QUIC release (change on connect_id length) and also simplified the check on release version


lucaderi:
Thanks for the contribution

kYroL01:
@alagoutte I'd suggest you also provide us a pcap with new packets for quic. In this way we keep test directory updated.
Thanks

kYroL01:
@alagoutte Another thing: are u sure it's a good idea to delete completely the **sequence** check ? I did right now a pcap from my wifi and I continue to see the sequence field.

backer-and:
To avoid false positives and according to the [QUIC Specifications](https://docs.google.com/document/d/1WJvyZflAO2pq77yOLbp9NsGjC1CHetAXV8I0fQe-B_U), for our purpose it seems that nonzero **sequence** checks are safer in all cases, given that it is possible to capture QUIC packets with specifics public flags (for example **without** **Version** and **CID**; `Public Flags: 0x00`) which represents a very weak information to handle.
I think it is necessary to improve only meaningful checks without removing anything that can't cause conflicts, even if it starts using only 0 and 8 bytes CID.

Until now, with Chromium (52.0.2730.0 canary) is still in use _QUIC_VERSION_32_, so we can’t provide a proper audit for _Q033_ and for the latest _Q034_ which adds an optional diversification nonce to packet headers, and eliminates the 2 byte and 4 byte connection ID length.

kYroL01:
I think @backer-and is right. The sequence check is necessary. 
The only case when we have false positive is when an udp pkt is detected as a quic pkt with flags = 0 and seq > 0 (first pcap).
Thus imho we cannot delete the sequence check. Look at this pcap from yesterday (second pcap).
Is there anything we miss to know ? If you can gently explain could be very usefull.
Thanks.

1) [Quic false positive](https://www.dropbox.com/s/138j26xug5hojdo/udp_noquic.pcapng?dl=1)

2) [Quic pcap](https://www.dropbox.com/s/6mezs9jfgkvwozy/quic.pcap?dl=1)

alagoutte:
Hi,

It will push a some pcap with my next patch (with Q033 from QUIC client/server toy https://github.com/google/proto-quic )

About sequence number check (it is now packet number...), each QUIC conversation start with "Client Hello" from client, it is not enough ?

kYroL01:
Yes, now the name is packet number instead of sequence number.
I don't think conversation start is enough. During a flow you are not sure to catch from the beginning. So it's necessary a more control.

I want to say that I continue to see also previous versions of quic packets, so I think we cannot consider versions < 33 deprecated. Is it correct ?

I need a clarification for this. Thanks.

alagoutte:
> Yes, now the name is packet number instead of sequence number.
> I don't think conversation start is enough. During a flow you are not sure to catch from the beginning. So it's necessary a more control.

Ok.. we need to make a better heuristics ! (because only check first 2 bytes, we be have often a lot false postive...)

> I want to say that I continue to see also previous versions of quic packets, so I think we cannot consider versions < 33 deprecated. Is it correct ?
> No, but from discussions on proto-guic mailing, there is never 2 or 4 bytes CID length...
> See my "fix" on Wireshark QUIC Dissector https://code.wireshark.org/review/#/c/15164/

kYroL01:
Ok good point. So I think we have to modify just the CID, not delete the sequence check.

@backer-and finish to give some modification and, after merge his pull req, test if is better or not. Is ok for you ?

PR #190, opened 2016-05-11
enhance/refactor quic.c

backer-and:
UDP _false positive_ exception fix, _version match_ updates. Temporary patch before new heuristics and optimizations (processing conditions according to specifications flowchart, optional diversification nonce, packet numbers and more as reported). #188

### DIFF
--- a/src/lib/protocols/quic.c
+++ b/src/lib/protocols/quic.c
@@ -31,42 +31,67 @@
 #define QUIC_SEQ_MASK 0x30      // 0011 0000
 
 #define CID_LEN_8 0x0C          // 0000 1100
+#define CID_LEN_0 0x00          // 0000 0000
+
 #define CID_LEN_4 0x08          // 0000 1000
 #define CID_LEN_1 0x04          // 0000 0100
-#define CID_LEN_0 0x00          // 0000 0000
 
 #define SEQ_LEN_6 0x30          // 0011 0000
 #define SEQ_LEN_4 0x20          // 0010 0000
 #define SEQ_LEN_2 0x10          // 0001 0000
 #define SEQ_LEN_1 0x00          // 0000 0000
 
-#define SEQ_CONV(ARR) (ARR[0] | ARR[1] | ARR[2] | ARR[3] | ARR[4] | ARR[5] << 8)
-
+#define INT(C) (C - '0')
+#define DIGIT(X, Y, Z) ((isdigit(X) && isdigit(Y) && isdigit(Z)) ? (INT(X) * 100 + INT(Y) * 10 + INT(Z)) : 0)
 
 #ifdef NDPI_PROTOCOL_QUIC
-static void ndpi_int_quic_add_connection(struct ndpi_detection_module_struct
-                                         *ndpi_struct, struct ndpi_flow_struct *flow)
+static void ndpi_int_quic_add_connection(struct ndpi_detection_module_struct *ndpi_struct, struct ndpi_flow_struct *flow)
 {
-  ndpi_set_detected_protocol(ndpi_struct, flow, NDPI_PROTOCOL_QUIC, NDPI_PROTOCOL_UNKNOWN);
+    ndpi_set_detected_protocol(ndpi_struct, flow, NDPI_PROTOCOL_QUIC, NDPI_PROTOCOL_UNKNOWN);
 }
 
 static int connect_id(const unsigned char pflags)
 {
     u_int cid_len;
 
-        // Check CID length.
-        switch (pflags & QUIC_CID_MASK)
-        {
-           case CID_LEN_8: cid_len = 8; break;
-           case CID_LEN_4: cid_len = 4; break;
-           case CID_LEN_1: cid_len = 1; break;
-           case CID_LEN_0: cid_len = 0; break;
-           default:
-               return -1;
+    // Check CID length
+    switch (pflags & QUIC_CID_MASK)
+    {
+    	case CID_LEN_8: cid_len = 8; break;
+    	case CID_LEN_4: cid_len = 4; break;
+    	case CID_LEN_1: cid_len = 1; break;
+    	case CID_LEN_0: cid_len = 0; break;
+    	default:
+            return -1;
+    }
 
+    // Return offset
+    return cid_len + 1;
+}
+
+static int sequence(const unsigned char *payload)
+{
+    char test[6] = {0};
+    int cid_offs = connect_id(payload[0]);
+    int seq_lens;
+
+    // Retrieve SEQ offset.
+    if (cid_offs >= 0)
+    {
+        // Search SEQ bytes length.
+        switch (payload[0] & QUIC_SEQ_MASK)
+        {
+            case SEQ_LEN_6: seq_lens = 6; break;
+            case SEQ_LEN_4: seq_lens = 4; break;
+            case SEQ_LEN_2: seq_lens = 2; break;
+            case SEQ_LEN_1: seq_lens = 1; break;
+            default:
+                return 0;
         }
-        // Return offset.
-        return cid_len + 1;
+    }
+
+    // Return SEQ comp value;
+    return memcmp(payload + cid_offs, test, seq_lens);
 }
 
 void ndpi_search_quic(struct ndpi_detection_module_struct *ndpi_struct, struct ndpi_flow_struct *flow)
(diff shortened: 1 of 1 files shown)


kYroL01:
Thanks

src/lib/protocols/quic.c
```diff
@@ -31,107 +31,135 @@
...
+            // Version presence; CID length
+            if (packet->payload[0] & QUIC_VER_MASK)
+            {
+                // Skip CID length
+                ver_offs = connect_id(packet->payload[0]);
+
+                if (ver_offs >= 0)
+                {
+                    const uint8_t *vers = &packet->payload[ver_offs];
+                    short int res = DIGIT(vers[1], vers[2], vers[3]);
```
alagoutte, line 127: Why don't use atoi() ?

src/lib/protocols/quic.c
```diff
@@ -31,107 +31,135 @@
...
+
+                    if (*vers == 'Q')
+                    {
+                        // QUIC versions. '22', '28' < omitted.
```
alagoutte, line 131: Why remove 22 and 28 ?

src/lib/protocols/quic.c
```diff
@@ -31,107 +31,135 @@
...
-           case CID_LEN_8: cid_len = 8; break;
-           case CID_LEN_4: cid_len = 4; break;
-           case CID_LEN_1: cid_len = 1; break;
-           case CID_LEN_0: cid_len = 0; break;
-           default:
-               return -1;
+    // Check CID length
+    switch (pflags & QUIC_CID_MASK)
+    {
+    	case CID_LEN_8: cid_len = 8; break;
```
alagoutte, line 60: Need to be change for support > Q033

src/lib/protocols/quic.c
```diff
@@ -31,107 +31,135 @@
...
+    int cid_offs = connect_id(payload[0]);
+    int seq_lens;
+
+    // Retrieve SEQ offset.
+    if (cid_offs >= 0)
+    {
+        // Search SEQ bytes length.
+        switch (payload[0] & QUIC_SEQ_MASK)
+        {
+            case SEQ_LEN_6: seq_lens = 6; break;
```
alagoutte, line 84: Rename to PKT_LEN_XX ?

src/lib/protocols/quic.c
```diff
@@ -31,107 +31,135 @@
...
+            case SEQ_LEN_1: seq_lens = 1; break;
+            default:
+                return 0;
         }
-        // Return offset.
-        return cid_len + 1;
+    }
+
+    // Return SEQ comp value;
+    return memcmp(payload + cid_offs, test, seq_lens);
```
alagoutte, line 94: What the idea ? because we only check if there is some "bytes"
It is no possible to compare if there is alreay the CID ?

alagoutte:
Sorry, the PR is already merged but i have some question...

Do you plan to make some other change ? (i will add support of Q033 and fix Sequence -> Packet)

src/lib/protocols/quic.c
backer-and, line 127: _atoi()_ is not nul terminated. You should make a **copy of 3 bytes to a zero terminated array** (would be an assignment), there's no copy here.

src/lib/protocols/quic.c
backer-and, line 131: I did not see the presence of these versions but if you can confirm, can be added.

src/lib/protocols/quic.c
backer-and, line 60: We are discussing about to completely remove this two-bytes control or add a support for older versions. For now, in **Q032** still exist.

src/lib/protocols/quic.c
backer-and, line 84:
```
PKN_LEN_6 0x30          // 0011 0000
PKN_LEN_4 0x20          // 0010 0000
PKN_LEN_2 0x10          // 0001 0000
PKN_LEN_1 0x00          // 0000 0000
```

only for newer versions still in test.

src/lib/protocols/quic.c
backer-and, line 94: _SEQ_ must be greater than 1 this is a very simple check.

backer-and:
@lucaderi, @kYroL01 _Chromium Version 52.0.2735.0 canary_ continues to use **Q032**, this was not a fix for **Q033 and Q034** but only a temporary patch.
I was working on support for new versions with different _public flags_:

```
#define QUIC_FLG_RESET 0x02     // 0000 0010
#define QUIC_FLG_NONCE 0x04     // 0000 0100
#define QUIC_FLG_CIDLT 0x08     // 0000 1000
#define QUIC_FLG_MPATH 0x40     // 0100 0000

#define PKN_LEN_6 0x30          // 0011 0000
#define PKN_LEN_4 0x20          // 0010 0000
#define PKN_LEN_2 0x10          // 0001 0000
#define PKN_LEN_1 0x00          // 0000 0000
```

if it is no necessary a support for older version we can remove:

```
#define CID_LEN_8 0x0C          // 0000 1100
#define CID_LEN_0 0x00          // 0000 0000
```

and start to consider only **QUIC_FLG_CIDLT** in this example, but i have doubts about this choice.

@alagoutte I do not understand certain subtleties on code fragments and not sure why you're such in a rush to make these changes.

src/lib/protocols/quic.c
kYroL01, line 60: @alagoutte for example the last version of _Google Chrome (50)_ **Q033** is not already supported and fo keep portability we must keep also older version than Q033.

src/lib/protocols/quic.c
kYroL01, line 84: @backer-and @alagoutte Ok is a good thing rename this field.

kYroL01:
@alagoutte No problem for other checks. We work always to do the best work so every ideas and contribution is well accepted!
All the questions are about portability, because I see that not all the browser support Q033 and keep also check to older versions is not an overhead.

src/lib/protocols/quic.c
alagoutte, line 60: It is easy to kept the support of old release, there is never  1 and 4 CID length (See https://groups.google.com/a/chromium.org/d/msg/proto-quic/pR6FA7oyeV8/4Ec42-GrMQAJ )

src/lib/protocols/quic.c
alagoutte, line 94: @backer-and 
it will be easy to have same false positive ?

alagoutte:
My review is only for also supported new release (and try also to continue to support old release)
And i have only some doubt about sequence check and false positive

src/lib/protocols/quic.c
kYroL01, line 60: I'm agree with you.

src/lib/protocols/quic.c
alagoutte, line 131: Ok after look QUIC Wire Layout Specification, there is no Q022 or Q028...
but i think we can change to only if ( and don't omitted Q022 and Q028...)

kYroL01, line 131: Ok

alagoutte:
i have upload a pcap with "Client HELLO" Q033 (and Public Reset Flags) on Wireshark bugtracker https://bugs.wireshark.org/bugzilla/show_bug.cgi?id=11494

backer-and:
> **alagoutte** added a note:
> Rename to PKT_LEN_XX

@alagoutte, You are right but are not only simple text changes?

```
#define PKN_LEN_6 0x30          // 0011 0000
#define PKN_LEN_4 0x20          // 0010 0000
#define PKN_LEN_2 0x10          // 0001 0000
#define PKN_LEN_1 0x00          // 0000 0000
```

_instead of (the same)_:

```
#define SEQ_LEN_6 0x30          // 0011 0000
#define SEQ_LEN_4 0x20          // 0010 0000
#define SEQ_LEN_2 0x10          // 0001 0000
#define SEQ_LEN_1 0x00          // 0000 0000
```

> **alagoutte** commented:
> It is easy to kept the support of old release, there is never 1 and 4 CID length (See https://groups.google.com/a/chromium.org/d/msg/proto-quic/pR6FA7oyeV8/4Ec42-GrMQAJ )

I understand your point of view and i read your previous changes, but we will have a byte of public flags in which the **CID** `0000 1000` will become a bit of _presence_, instead of two:  `0000 1100`.

So we know that in **Q032** version and earlier we were jumping 8 bytes CID with this mask: 
`#define CID_LEN_8 0x0C` > `0000 1100` (the fact that lengths of 1 and 4 have never been used, doesn't matter for what concerns the backward compatibility).

Now, for **Q033** and later we will have:
`#define QUIC_FLG_CIDLT 0x08` > `0000 1000`, and the second bit has become the **diversification nonce**: 
`#define QUIC_FLG_NONCE 0x04` > `0000 0100`.  
Seems that we can't use old flags for QUIC latest versions because we have to know exactly how many bytes to skip (and for example you can not use a `memmem()` function to search version substring in the whole udp payload, i hope you were not considering this option).

A **workaround** might be `0000 1*00`, considering for example a 4 bytes **CID** as always 8 (let us remember that now `*` represents the diversification nonce). _It would work in case of non-chrome/google implementations where a cid of 1 and 4 bytes were allowed by specifications?_

I said that this patch was temporary and there will definitely need a worthy fix, but in your previous pull request i did not see the full compatibility. Anyway if you have completed a fix which takes into consideration all these things, it would be great.

alagoutte:
@backer-and  _It would work in case of non-chrome/google implementations where a cid of 1 and 4 bytes were allowed by specifications?_ There is no other public implementation for the moment...

for diversification nonce, if we have the version (like client Hello), it is possible to get the number of bytes to skip... (but yes, need to known the version...)

alagoutte:
i have upload a pcap with "Client HELLO" Q033 (and Public Reset Flags) on Wireshark bugtracker https://bugs.wireshark.org/bugzilla/show_bug.cgi?id=11494

PR #203, opened 2016-05-24
fix warning found by Clang Analyzer (scan-build)

alagoutte:


alagoutte:
There is always 5 warnings with Clang Analyzer : 

@kYroL01 can you look fox rx ?
protocols/rx.c:129:2: warning: Value stored to 'found' is never read
protocols/rx.c:134:2: warning: Value stored to 'found' is never read
protocols/rx.c:139:2: warning: Value stored to 'found' is never read
protocols/rx.c:144:2: warning: Value stored to 'found' is never read
protocols/rx.c:150:7: warning: Value stored to 'found' is never read

and also 
ndpi_main.c:2827:6: warning: Access to field 'version' results in a dereference of a null pointer (loaded from variable 'decaps_iph')
  if(decaps_iph->version == 4 && decaps_iph->ihl >= 5) {
     ^~~~~~~~~~~~~~~~~~~

kYroL01:
@alagoutte Sure, i'm going to watch when I arrive at home. I can't do now 'cause i'm on train.

kYroL01:
@alagoutte are u sure that [this](https://github.com/ntop/nDPI/pull/203/commits/e0e1ed6936081875c28d37120b9fec9960dfa0bd) is correct ?
It seems that we read _sport_ and _dport_ here
https://github.com/alagoutte/nDPI/blob/e0e1ed6936081875c28d37120b9fec9960dfa0bd/src/lib/protocols/rtcp.c#L29

alagoutte:
@kYroL01 Yes, the remove is line 51 (after)

kYroL01:
@alagoutte My fault. I have read wrong line ;)

kYroL01:
@alagoutte check for rx warnings now.
For the **ndpi_main.c:2827:6** warning I think I understand but I have to check better.

PR #207, opened 2016-05-26
ndpi_main: fix Value stored to 'saddr'/'daddr' is never read

alagoutte:
ndpi_main.c:3286:7: warning: Value stored to 'saddr' is never read
ndpi_main.c:3290:7: warning: Value stored to 'saddr' is never read
ndpi_main.c:3291:7: warning: Value stored to 'daddr' is never read


alagoutte:
@kYroL01 can you check ?

kYroL01:
Yes, I noticed that bug last time. Going to check tomorrow if the fix is correct. Thank you @alagoutte

lucaderi:
Thank you to both of you

PR #215, opened 2016-06-15
1.7 stable

ne-vlezay80:
Please add tcp reassembly and flow reconstruction support from NDPI.


kYroL01:
Just to know: why are u using the 1.7 instead new 1.8 ?

lucaderi:
@vel21ripn Looking at the changes they are small and unless you want to port them to the current dev version as @kYroL01 said, we can do that. Thanks for the contribution

lucaderi:
@vel21ripn Actually I have seen at the code more in details and the fixes are not necessary for 1.8 anymore. Thanks again for your contribution.

PR #216, opened 2016-06-15
1.8 stable

ne-vlezay80:
Please add tcp reassembly and flow reconstruction support from NDPI.


kYroL01:
@ne-vlezay80 I think you confuse a pull request with open an issue. I think your request is more an issue than a pull request, isn't it ?
I see NO changes, but only a request of changes.

PR #239, opened 2016-08-04
OpenVPN detection by opcode, packet ID and session ID

emanuele-f:
This detection is port independent and based upon session ID acknowledgment (2 packets) during OpenVPN handshake


kYroL01:
Thank you! 
Can you also provide a pcap to insert in test folder ?

lucaderi:
Note that the code contained an non-portable code (https://github.com/ntop/nDPI/commit/9e207f7f24fd693c8d5a095db43dd7077e9fc903). Next time please make sure the code compiles on all platforms and not just on one Linux distro.

emanuele-f:
Ok sorry

PR #244, opened 2016-08-12
Add STARTTLS detection

emanuele-f:


kYroL01:
1) Just to understand better: is there any chance that an IMAP TLS/SSL traffic from port 993 could be different from IMAPS ?
2) Could you please provide a pcap with imaps traffic to verify better the detection ?
Thank you

emanuele-f:
A server could possibly be configured to use port 993 for IMAP and then use STARTTLS, so yes in theory. I assumed IMAPS and IMAP + STARTTLS should both be tagged as "IMAPS"

kYroL01:
Ok, I think is correct. Thanks

PR #245, opened 2016-08-12
added Slack service

denisvm:
Added Slack.com service according to https://get.slack.help/hc/en-us/articles/205138367-Troubleshooting-Slack-connection-issues and service capture.


kYroL01:
Thank you

PR #252, opened 2016-08-22
Add hotmail service and enable full master protocol detection for sub-protocols

emanuele-f:
When a subprotocol is matched, SSL detection completes but master protocol remains "SSL" even if it could be IMAPS,SMTPS,... Now master protocol refinement also applies to subprotocols, so "SSL.GMail" becomes "IMAPS.GMail" or "SMPTS.GMail"


kYroL01:
Yes! Thanks for the fix!

PR #257, opened 2016-09-04
Some bugs in H323 dissector

stillunfolding:
[h323.c.zip](https://github.com/ntop/nDPI/files/453889/h323.c.zip)

There are two small bugs in the H323.c dissector. 

1: In the lines 68-70 the value "5" must be replaced with "6", because we check the first sixth bytes:

```

if(packet->payload_packet_len >= 5 && packet->payload[0] == 0x80 && packet->payload[1] == 0x08 &&
        (packet->payload[2] == 0xe7 || packet->payload[2] == 0x26) &&
        packet->payload[4] == 0x00 && packet->payload[5] == 0x00)
```

2- In the line 80, the || must be replaced with &&, because otherwise the condition will be true always.

```
     else if(packet->payload_packet_len >= 20 || packet->payload_packet_len <= 117)
        {
```


kYroL01:
fixed by https://github.com/ntop/nDPI/commit/7300e88cdb9d22c33e276862e29b1990b394a0c5

Next time I suggest to open an issue instead a pull request, or send a pull request with properly modifications.

Thank you for contribution

PR #258, opened 2016-09-05
Fix facebook certificate recognition

emanuele-f:
Fixes some SSL flows with "facebook.com" not being recognized. See test pcap.


lucaderi:
Thank you.
However this fix is a bit dangerous as we might end up matching things are not supposed to. Example welcometofacebook.com. But in order to do this we need to change the way we match traffic introducing a different matching scheme (e.g. initial substring match)

PR #276, opened 2016-10-14
Added detection of OnSatMail service packets

linusoft:


kYroL01:
I'm going to test it

kYroL01:
I think the pull is correct, but before merging I have some question:
1) you check if the packet is the IPv4. So this proto avoid IPv6 ?
2) do u have any other information (payload lenght, if protocol is over TCP or UDP) to put in the dissector ?
3) is better create a variable `struct ndpi_packet_struct *packet` and initialize to `&flow->packet`. In this way all the checks are doing on this packet, avoiding any (possibile) crush on the flow original structure (I can change this part after the merge).
4) Do u have pcap to put inside the tests directory ?

Thank you

src/lib/Makefile.am
```diff
@@ -154,6 +154,7 @@ libndpi_la_SOURCES = ndpi_content_match.c.inc \
 		     protocols/yahoo.c \
 		     protocols/zattoo.c \
 		     protocols/zeromq.c \
+                     protocols/onsatmail.c \
```
alagoutte, line 157: it is sort by name

src/lib/ndpi_content_match.c.inc
```diff
@@ -71,6 +71,12 @@ static ndpi_network host_protocol_list[] = {
     176.34.184.0/21
   */
   { 0xB022B800 /* 176.34.184.0/21 */, 21, NDPI_SERVICE_AMAZON },
+ 
+ /*    
+    OnSatMail
+    64.4.141.247/32
+  */
+  { 0x40048DF7 /* 64.4.141.247/32 */, 32, NDPI_SERVICE_ONSATMAIL },
```
alagoutte, line 79: only 1 IP ?

src/lib/protocols/onsatmail.c
```diff
@@ -0,0 +1,52 @@
...
+#ifdef NDPI_SERVICE_ONSATMAIL
+
+void ndpi_search_onsatmail(struct ndpi_detection_module_struct *ndpi_struct, struct ndpi_flow_struct *flow)
+{
+	if (flow->packet.iph)
+	{
+		// IPv4
+		u_int32_t src = ntohl(flow->packet.iph->saddr);
+		u_int32_t dst = ntohl(flow->packet.iph->daddr);
+		u_int16_t dport;
```
alagoutte, line 20: variable never used

src/lib/protocols/onsatmail.c
```diff
@@ -0,0 +1,52 @@
...
+		u_int16_t sport;
```
alagoutte, line 21: same issue

alagoutte (changes requested):

kYroL01:
I'm agree with @alagoutte . This is why I cannot still merge this pull request

PR #278, opened 2016-10-20
Avoid c99 complains about function prototypes

mgigante:
Please, can you merge these changes? They can simplify libndpi usage by 3rd parties applications.
Compiling with -Wstrict-prototypes -Werror flags generates error because of missing void in function prototypes (functions with no arguments).


kYroL01:
Thank you

PR #279, opened 2016-10-22
Add protocol category ids

emanuele-f:
Work in progress. Please double check protocol to category mappings.


kYroL01:
Thanks @emanuele-f I'm going to check the pull.

kYroL01 (commented):
I think the work is good, but IMHO we have to reduce the number of categories to keep thing simpler.
I.e. Network instead of (Network and Network_Tool).
What do u think about it ?

lucaderi:
I see there are a few things to change. For instance I would like to see enums instead of #defines (this has to be fixed in many other nDPI code locations so I am glad that this contrib is clean from this point of view). I prefer to accept the contribution as is and make selected fixes on the patched code.

PR #180, opened 2016-04-19
High level API and ndpiReader simplification

emanuele-f:
Migrates potentially reusable code from ndpiReader to ndpi_util module to promote a clean and fast nDPI integration

PR #288, opened 2016-11-03
setup the debug function

vpiserchia:


kYroL01:
Thanks

vpiserchia:
Hello,

I question about the internal workflow of the code; in the past it was supporting a multi step approach:
- look for all the packets in a flow for ALL the not excluded protocols.

Now it just stop at the first protocol match; is this true or am I missing something?
Would be possible to restore the previous behaviour accepting the impacts on the 
performance side?

thank you

kYroL01:
If a packet is detected as a protocol, it makes no sense to continue the detection on all the dissector. The performance are important.
In fact if a protocol for a flow is not detected, we put in the excluded bitmask.
Why you want to restore the previous behaviour ? You have many false positive for any protos ?

vpiserchia:
Thank for the replay; I'll try to explain because I think I was not so clear in the previous message.
As you know there are many cases where the same flow (TCP o UDP) can be used to transport different kind of traffic (e.g. STUN packets in RTP stream).
In the ORIGINAL_CODE this behaviour was covered with a multistep approach; now this is gone.
I'd like to activate some switches to enable it again.
My question is: can you kindly point me to the commit that changed the workflow?
And yes specifically for the RTP/STUN case I have many false positive. Also the "CLASSIC STUN" (that is the one defined RFC 3489) is not decoded/reconized anymore, and even if it decoded (using the origin code) then the RTP stream in the same flow is not decoded at all because nDPI now stop at the first match.
I hope it's clearer now.

vpiserchia:
There is I can add to help with this?

kYroL01:
@vpiserchia For the question of "original code" I think you refer of this: https://github.com/ntop/nDPI/pull/180/files

For the second question this function https://github.com/ntop/nDPI/blob/58e97c966c3aef178580d8f362df95274fbe9c8a/example/ndpi_util.c#L520
is the one that stop flow.
I think that you are agree for the RTP/STUN because I see that is any case it has a strange behaviour and I also think that we need a fix, but I don't think that the previous approach is better. Maybe we just need a better detection in the dissector.
Please, write me a private mail so we can discuss better of this things. campus@ntop.org 
I also wanna see stun pcap that is not recognize anymore.
Thanks

PR #289, opened 2016-11-11
Add Github service

mkfsn:
Added Github service according to:

- https://help.github.com/articles/what-ip-addresses-does-github-use-that-i-should-whitelist/
- https://developer.github.com/changes/2014-04-25-user-content-security/


kYroL01:
I'm going to check it

kYroL01:
Thanks

PR #290, opened 2016-11-12
Scan build

alagoutte:
Fix last warning found by Scan build (Clang Analyzer)

There is also this warning but false positive i think  :
ndpiReader.c:993:5: warning: Function call argument is an uninitialized value


kYroL01:
Correct. Thanks

Issue #224, opened 2016-07-05
add canon bjnp protocol


Niko78:
on my network I have few packets generated by canon software (kind of scan)
it is named BJNP protocol

I added pcap
[bjnp.zip](https://github.com/ntop/nDPI/files/347812/bjnp.zip)

thanks

Nico

alagoutte:
It is easy to write a heuristic ! each packet start with magic ID: BJNP

mkfsn:
Hi, I make a solution for it, and here is my solution:

https://github.com/mkfsn/nDPI/commit/66e646bed58f5354c19727671e9194f4ceefc005

alagoutte:
Look good, it is possible to make a PR ?

mkfsn:
@alagoutte: I'm new to Github projects, should I make a new PR for this, or [attach a PR to this issue](https://www.topbug.net/blog/2012/03/25/attach-a-pull-request-to-an-existing-github-issue/) ?

Thank you.

kYroL01:
@mkfsn Thanks for your work.
Just send a pull request normally, then I check it and merge into the main project.
If it's good I close this issue after merge.
Thanks

kYroL01:
Implemented [here](https://github.com/ntop/nDPI/commit/0f94a46f868e6818790b0c278121c081a63c9eeb)

PR #294, opened 2016-11-15
Add BJNP Protocol

mkfsn:
Added Canon BJNP protocol according to: #224


kYroL01:
@mkfsn Cannot check and merge. As u see there are any conflicts that must be resolved with `src/lib/ndpi_main.c`. Please checkout and send again.

PR #298, opened 2016-11-17
Update and merge Twitter networks

denisvm:
Removed duplicated and smaller networks, also added Twitter APAC networks


kYroL01:
Thank you. Is correct

PR #302, opened 2016-11-22
not very sure about this change

tianzhishuizhu:


kYroL01:
Maybe you have an old version of nDPI.
`struct ndpi_flow_info *flow ` is moved in ndpi_utils.c.

Update your nDPI, thanks.

PR #306, opened 2016-11-23
Checking for libdl vs. libc during configure

mgigante:
Please take a look at the changes whether they are ok or not.
ndpiReader fails to link on BSD systems


mgigante:
As side question on this pull request. Does the libnuma required on linux when using PF_RING?
During compilation test for the changes on this pull request I got a linking error about missing libnuma (which I actually do not have on my linux system). Browsing the nDPI and PF_RING code I cannot find any reference to to libnuma functions. Not 100% sure on this though.

kYroL01:
Thanks for the pull.

Yes you have to install libnuma to use PF_RING correctly.
Look at this http://www.gossamer-threads.com/lists/ntop/misc/39305

mgigante:
Thank you.
About libnuma and PF_RING, I believe it's better to move all the related stuff under a linux os conditional in configure.ac. I'll propose a new pull request as soon as I can.

kYroL01:
@mgigante for any PF_RING issue, please open pull request in the PF_RING repository.
Thanks for the contribution

PR #312, opened 2016-12-06
Removed some magic numbers

denisvm:
Translated some magic numbers to defines and others to simple numbers to make it more understandable.


kYroL01:
Thank you

PR #314, opened 2016-12-07
Fix 'IPVERSION undeclared' error on WIN32

sologub:
Windows does not contain the `netinet/ip.h` header file so nDPI throws a `'IPVERSION': undeclared identifier` compilation error when compiled with Visual Studio.
This PR fixes that error.


kYroL01:
Thanks

PR #319, opened 2016-12-14
Add Valve AS routes and domain names for better Steam detection.

k0ste:
Continuing to improve the detection traffic of Steam application.


kYroL01:
Thank you!

PR #321, opened 2016-12-29
SMPP: fix comparison of unsigned expression >= 0 is always true [-Wer…

alagoutte:
…ror,-Wtautological-compare]


kYroL01:
Thank you!

PR #328, opened 2017-01-26
Using autoconfigured variable for libdl linkage

mgigante:
Hi,
-ldl has been explicitly reintroduced at some point after my previous pull-request.
Was there any case skipped by configure which does not properly substitute DL_LIB in makefiles?
Should not that be the case, please pull this change in.
Or can you give me some details where the autoconfigured DL_LIB variable fails so I can check?
Thanks


kYroL01:
@mgigante  I think is fixed by https://github.com/ntop/nDPI/commit/b161a6d7f797dcdd3f258c03570853d555085798

PR #341, opened 2017-02-11
Easing OpenBSD build

numeronove:
Addedd macro `AM_PROG_CC_C_O` in `configure.ac`, needed to have `autogen.sh`
manage the error:

    C objects in subdir but `AM_PROG_CC_C_O' not in `configure.ac'

Modified case instance in `example/ndpi_util.c`, since `DLT_LINUX_SLL`
isn't defined in OpenBSD `pcap.h`.

### DIFF
--- a/example/ndpi_util.c
+++ b/example/ndpi_util.c
@@ -650,10 +650,12 @@ void ndpi_workflow_process_packet (struct ndpi_workflow * workflow,
     break;
 
     /* Linux Cooked Capture - 113 */
+#ifdef __linux__
   case DLT_LINUX_SLL :
     type = (packet[eth_offset+14] << 8) + packet[eth_offset+15];
     ip_offset = 16 + eth_offset;
     break;
+#endif
 
     /* Radiotap link-layer - 127 */
   case DLT_IEEE802_11_RADIO :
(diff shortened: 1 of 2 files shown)


kYroL01:
I'm testing. Thanks for pull req

example/ndpi_util.c
```diff
@@ -650,10 +650,12 @@ void ndpi_workflow_process_packet (struct ndpi_workflow * workflow,
     break;
 
     /* Linux Cooked Capture - 113 */
+#ifdef __linux__
   case DLT_LINUX_SLL :
```
alagoutte, line 654: Maybe add define of value if it is not define ?

numeronove: It should be defined automatically by gcc/g++

```shell
$ gcc -dM -E - < /dev/null | grep linux
#define __linux 1
#define __linux__ 1
#define __gnu_linux__ 1
#define linux 1
```

Or am I missing something?

kYroL01: I think @numeronove fix is correct. OpenBSD does not support DLT_LINUX_SLL

http://openbsd-archive.7691.n7.nabble.com/redhat-lt-gt-openbsd-tcpdump-td272594.html

PR #343, opened 2017-02-17
ndpi_content_match: Add origin routes for many services.

k0ste:
Run tests - get improvements.


alagoutte:
Do you have create a script for generate this file ?

k0ste:
No, but the results can be repeated.
I.e. I have a algorithm for the extraction of advertised prefixes.
For clear: route summarization string (ip_src, ip, mask_src, mask, ndpi_proto) - yes, by script, I mean I not have script "generate this file".

kYroL01:
Interesting work... Have to test. Thanks

PR #350, opened 2017-03-15
missing macro NDPI_PROTOCOL_IQIYI:

vpiserchia:


kYroL01:
Actually I don't underatand this commit.
IQIYI is just detected as a protocol. Why we need to add it as a service ?

vpiserchia:
Because that's macro is used in the NDPI_LOG function, and it fails

check here: 
https://github.com/ntop/nDPI/search?utf8=%E2%9C%93&q=NDPI_SERVICE_IQIYI

Maybe I got it wrong, and as you said IQIYI is a protocol and not a service, but then line https://github.com/ntop/nDPI/blob/1e014aa95b76e5a73a7a2bef9b729a25c914c735/src/lib/protocols/http.c#L206 in the http.c needs to refer to NDPI_PROTOCOL_IQIYI directly ?

just let me know, and I'll send a fix for this.
vito

kYroL01:
I think I have to change SERVICE to PROTOCOL.

vpiserchia:
in this case feel free to close this pull request

PR #351, opened 2017-03-16
Added ndpi_flow_ free/malloc and set_ functions:

vpiserchia:
Ideally these two new functions will provide more flexibility
in regards of application that want to use specific allocation
approaches, for example storage queues for this structures
from where the application can pop and release the data it
needs.
Semantically they can be set up as the other ndpi_malloc and ndpi_free
functions, that is via the set_ndpi_flow_malloc and set_ndpi_flow_free.
In case a NULL parameter is passed to the above functions the old
approach will be used, that means that the ndpi_malloc will be used as
the ndpi_flow_malloc function and ndpi_free_flow as the ndpi_flow_free one.

Application that use the old functions will be not affected by this change,
and the ndpi_free_flow can be deprecated in the future and made private.

### DIFF
--- a/example/ndpi_util.c
+++ b/example/ndpi_util.c
@@ -62,16 +62,13 @@
 #define GTP_U_V1_PORT                   2152
 #define TZSP_PORT                      37008
 
-#define SIZEOF_ID_STRUCT (sizeof(struct ndpi_id_struct))
-#define SIZEOF_FLOW_STRUCT (sizeof(struct ndpi_flow_struct))
-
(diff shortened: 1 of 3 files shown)


vpiserchia:
Any news, no acceptance/agreement about this ?

kYroL01:
Sorry for answering in late.
I'll check asap and if there is no problem i merge.
Thank you!

kYroL01:
Thank you!

PR #360, opened 2017-04-07
unified HTTPDownload name with other HTTP protocols (as HTTP_Download)

jaspervandenberg:
Currently, the HTTP protocols available are: HTTP_Proxy, HTTP_Connect, HTTP and HTTPDownload. As you can see, there is some inconsistency in the naming of these protocols. Because of this, this request changes the name HTTPDownload to HTTP_Download.


kYroL01:
For me is ok.
@lucaderi this change some behaviour with ntopng ?
Thanks

kYroL01:
Thanks

PR #365, opened 2017-04-19
Update ndpi_util.c

Ravi-t:
Fix for crash while analyzing fragmented IP packets


Ravi-t:
Fix for the crash while analyzing fragmented IP packets. As ndpiReader doesn't have fragmented IP support so just returning unknown protocol for fragmented packets.

kYroL01:
The commit is ok, but I'm doing some modification to improve stats.

kYroL01:
check https://github.com/ntop/nDPI/commit/9a94c34b91250643c383904b14352e282cfb056c and let me know.

Thank you

PR #379, opened 2017-05-17
Fix buffer overflows in http.c

pavlosantoniou:
Unchecked accesses to packet contents lead to stack-buffer-overflows.


kYroL01:
Thank you!

PR #380, opened 2017-05-18
Use correct subscripts when checking IP array in ndpi_main.c

pavlosantoniou:
The same index is checked three times in logical OR statement


kYroL01:
Thank you

PR #381, opened 2017-05-18
Check pointer for NULL before dereferencing

pavlosantoniou:
Pointer variable 'flow' is checked for nullness in line 2970.
Either this check is redundant (the pointer is always not NULL), or the check for nullness needs to take place before dereferencing the pointer in the lines above 2970 too.


kYroL01:
Thanks

PR #382, opened 2017-05-18
Remove redundant check from btlib.c

pavlosantoniou:
If the checked value is equal to 'e' it can never evaluate to false.


kYroL01:
Thanks

PR #385, opened 2017-05-19
Added -b for saving port based diagnose statistics in json

beratx:
Two type of statistics are saved; statistics for packet burst and statistics for ip burst.


kYroL01:
Thank you!

PR #388, opened 2017-05-26
Added timestamp, flow count and app. level protocol to (-b) diagnose …

beratx:
…statistics (in json format)


kYroL01:
Thank you!

PR #389, opened 2017-05-29
Added tinc protocol detection

wgugliel:


kYroL01:
Hi @deselmo . I'm testing asap before merge but is a nice work. What about this meta-protocol, as described [here](https://www.tinc-vpn.org/documentation-1.1/The-meta_002dprotocol.html) ?
It seems interesting to know. 

Thanks

wgugliel:
The problem is that it is used after the authentication protocol ([here](https://tinc-vpn.org/documentation/Authentication-protocol.html#Authentication-protocol)). The only unencrypted messages are the first four.

kYroL01:
I tested the pull request and it works perfectly. Thanks.
Just two questions: 
1) why you implemented that libcache instead using an exist one (for example I think UTHash could be the same in another way) ?
2) What is a real use case of this protocol ?

Thank you

wgugliel:
1. Libcache implements a cache that gives a maximum size, only stores the most recently used data: it is not a real key-value map, but a set of values. I did not find appropriate implementations of this data structure, if you have another one to recommend we can integrate it instead of libcache.

2. Tinc is a VPN, and this protocol is used by two tinc daemons to communicate with each other.

   > Tinc is a Virtual Private Network (VPN) daemon that uses tunnelling and encryption to create a secure private network between hosts on the Internet.

kYroL01:
What about [UThash](https://troydhanson.github.io/uthash/userguide.html) ?
Is a macro library, you just import the `.h` and everything is made with macros. Your key could be various things, also a struct.
I don't know if it is better or usefull for your implementation, but take a look.
If not, thanks for your huge work and I think is mergeable without problem.
We have an example [here](https://github.com/ntop/nDPI/commit/4cc54cceb5514acce67bf577f2116ee97505d805) 

Thanks!

wgugliel:
UTHash does not expose a way to iterate the structure in an LRU order, which is what we need here. The cache has a maximum size, and when it's full, we need a fast and efficient way to purge the oldest entries.

I looked at UTHash, but I actually don't know if it will be beneficial to this particular implementation.

kYroL01:
Ok I understand. Good to know

kYroL01:
Thank you!

PR #395, opened 2017-06-07
Use flag %d to print signed integers in ndpiReader.

pavlosantoniou:
In printf, the flag %u is used to print unsigned integers.
The flag %d must be used for signed integers, so that the first bit is interpreted as the sign.


kYroL01:
Yes, you're right. Thanks

PR #413, opened 2017-06-25
fix lync over ICE detection

YumingMa:
Please see https://tools.ietf.org/html/draft-ietf-mmusic-ice-06#section-7.4 and the attachment pcap. You could search 80540004 in the pcap
[candicatepcap.zip](https://github.com/ntop/nDPI/files/1100116/candicatepcap.zip)
.


kYroL01:
Thanks

YumingMa:
You are welcome. You could close it.

Issue #421, opened 2017-07-13
SSL client vs. server certificate possible issue


MicahLyle:
Using the latest development version (also occurs on the 2.0 stable branch), with the https google request in the pcap below (there is a DNS lookup and then an HTTPS google request in the PCAP file), the SSL flow classifies the `client_certificate` as `www.google.com` but `server_certificate` is empty. Shouldn't it be `client_certificate` empty and `server_certificate` `www.google.com`?

[google_https_request_1.pcap.zip](https://github.com/ntop/nDPI/files/1146067/google_https_request_1.pcap.zip)

kYroL01:
Thanks for reporting. I'll check asap.

MicahLyle:
@kYroL01 you're welcome. When I looked into it and stepped through the debugger, I think I've found at least some of what's happening. I attached two screenshots below.
<img width="1113" alt="screen shot 2017-07-14 at 11 49 27 am" src="https://user-images.githubusercontent.com/10660805/28226351-a591d012-688a-11e7-93ea-fc5e900b6107.png">
![screen shot 2017-07-14 at 11 49 58 am](https://user-images.githubusercontent.com/10660805/28226352-a5962806-688a-11e7-90eb-778539228fd1.png)

Basically, if you look at the wireshark picture, along with the code, what ends up happening is that the client hello contains the server name extension with www.google.com as the Server Name. The code gets that far, and sets www.google.com as the client certificate, and says *We're happy now* in the code comment. Then in ssl.c it does this:

```
      rc = getSSLcertificate(ndpi_struct, flow, certificate, sizeof(certificate));
      packet->ssl_certificate_num_checks++;

      if(rc > 0) {
	packet->ssl_certificate_detected++;
#ifdef CERTIFICATE_DEBUG
	printf("***** [SSL] %s\n", certificate);
#endif
	u_int32_t subproto = ndpi_match_host_subprotocol(ndpi_struct, flow, certificate, 
							 strlen(certificate), NDPI_PROTOCOL_SSL);
  
	if(subproto != NDPI_PROTOCOL_UNKNOWN) {
	  ndpi_set_detected_protocol(ndpi_struct, flow, subproto,
				     ndpi_ssl_refine_master_protocol(ndpi_struct, flow, NDPI_PROTOCOL_SSL));
	  return(rc); /* Fix courtesy of Gianluca Costa <g.costa@xplico.org> */
```

[...]

PR #422, opened 2017-07-18
SSL Certificate Fix

MicahLyle:
Fixes/Addresses https://github.com/ntop/nDPI/issues/421.

This pull request for ssl certificates still detects protocols as quickly as the previous version that only used client certificates. However, it adds the option to process extra packets after detection by adding a few variables to the `ndpi_flow_struct.` Now, if people want (in the example library or the libraries they write), they can monitor if the `ndpi_flow_struct` says that extra packet processing is available, and call the respective function to do so if they would like. It's very little overhead, all that needs to happen is packet header initialization and some stats/state updates, and then the function specified by the protocol dissector that detected the protocol is run, and as soon as either a max extra packet count is hit or the extra packet function decides it's done, no extra processing is run.

Protocols can now define a specific function that they can set if they want to do extra packet processing, it's easiest to see the code for details but let me know if you want a more detailed writeup.

I've checked comparing against all the tests before the pull request. Before the pull request, only skype.pcap.out and tinc.pcap.out were "failing tests." After this pull request, those two tests are still failing with the exact some output as before the pull request.

Every other test is passing. All the byte counts and classifications etc. are the same, the only differences is that some of the tests needed to have server certificates appended to them, which I did, and they are all now passing (and again, it's the exact same output as before except some of the lines also have a server certificate now).


kYroL01:
HI @MicahLyle give me a couple of days and I'll read all about it and test this pull request and we can discuss of it.

Really appreciate that. Thanks

kYroL01:
Hi @MicahLyle I think your pull request is good and nice work. I just need to understand better a couple of things:

1. I run `gdb` to understand the lifecycle and I see that the default is to check the extra packets, that is ok, but how a user could choose to avoid the extra packet ?
2. Why `ndpi_process_extra_packet` is not inside ssl dissector but in main ? If you put inside ssl dissector you could avoid to re-check the packet headers and work directly on ssl payload, isn't it ?

Thanks for all.

MicahLyle:
@kYroL01 
1. Well actually, the default in ndpi_reader.c and ndpi_util.c (both in the example program) is to check the extra packets, but notice that you have to explicitly include this code to do so:

```
  /* Protocol already detected */
  if(flow->detection_completed) {
    if(flow->check_extra_packets && ndpi_flow != NULL && ndpi_flow->check_extra_packets) {
      if(ndpi_flow->num_extra_packets_checked == 0 && ndpi_flow->max_extra_packets_to_check == 0) {
        /* Protocols can set this, but we set it here in case they didn't */
        ndpi_flow->max_extra_packets_to_check = MAX_EXTRA_PACKETS_TO_CHECK;
      }
      if(ndpi_flow->num_extra_packets_checked < ndpi_flow->max_extra_packets_to_check) {
        ndpi_process_extra_packet(workflow->ndpi_struct, ndpi_flow,
							  iph ? (uint8_t *)iph : (uint8_t *)iph6,
							  ipsize, time, src, dst);
        if (ndpi_flow->check_extra_packets == 0) {
          flow->check_extra_packets = 0;
          process_ndpi_collected_info(workflow, flow);
        }
      }
    } else if (ndpi_flow != NULL) {
      /* If this wasn't NULL we should do the half free */
      /* TODO: When half_free is deprecated, get rid of this */
      ndpi_free_flow_info_half(flow);
    }
    return(flow->detected_protocol);
  }
```

This is not present in the core API, meaning for folks with their own versions of ndpi_reader (or whatever their program they use is to call the API), they have to *explicitly specify* that they want to check extra packets and take the appropriate action by checking `ndpi_flow->check_extra_packets` (and maybe, in the case of the example, although this isn't the cleanest solution, also checking one of their own defined variables in flow->check_extra_packets). They can also, within the example program or their own program, override ndpi_flow->max_extra_packets_to_check, I.E. they could set this to 0 to never check extra packets, or change the `\#define MAX_EXTRA_PACKETS_TO_CHECK` macro along with the code to always set their own maximum value, meaning no extra packets are ever checked. To summarize there are two ways to disable extra packet processing:
1. Don't check `ndpi_flow->check_extra_packets` in the first place in the example code (I.E. with your own custom version of the example code) and hence also don't call `ndpi_process_extra_packet`
2. (Not as good of a method but would still work) - Put something in the code like this:
```
(At the top in ndpi_util.h)
#define MAX_EXTRA_PACKETS_TO_CHECK  0
.
.
.
(In ndpi_util.c)
if (ndpi_flow->check_extra_packets) {
  ndpi_flow->max_extra_packets_to_check = 0;
}
```

2. `ndpi_process_extra_packet` is in main because I wanted to provide a uniform way for any protocol dissector to add the option to process extra packets for that protocol. The point of `ndpi_process_extra_packet` is just to set up all the headers, because protocols may want to still know the header fields (I.E. TCP header or even QUIC header, etc.) if they are processing extra packets to get extra information. Also, to figure out payload location and have value to even call, you would pretty much need to call `ndpi_init_packet_header` anyways, which is what `ndpi_process_extra_packet` does. The reason I used `ndpi_process_extra_packet`was to provide any function that is an extra packet processing function (like the one in the ssl dissector) with a consistent data structure that it can use (I.E. packet with all the headers parsed). This way, people can add extra packet processing capabilities to any protocol. All the protocol dissector would have to do, is if somewhere in its dissector it wanted to add the capability to process extra packets (even if it knows it figured out the app protocol and master protocol by then, etc.) it just needs to add code like this:
```
// Set the flag for checking extra packets for this ndpi flow
flow->check_extra_packets = 1;
// (Optionally) add a maximum number of extra packets to check
flow->max_extra_packets_to_check = 7;
// Add the callback function for extra packet checking that would be used (if the user checks for extra packets in something like ndpi_util.c
// and calls ndpi_process_extra_packet, then ndpi_process_extra_packet would call their registered callback function, passing it (when called)
// a detection module struct and a flow with a packet field that has all the headers set)
flow->extra_packets_func = sslTryAndRetrieveServerCertificate;
```

kYroL01:
Thanks for the huge work

PR #435, opened 2017-08-11
Windows 10 detection using UA and formating

pauloangelo:
This pull request includes just a few lines of actual code.


kYroL01:
I have to test. Thanks for now!

PR #442, opened 2017-09-04
The bug in the function ndpi_parse_packet_line_info().

vel21ripn:
The function does not see the last line of the packet.


kYroL01:
Thanks!

PR #445, opened 2017-09-08
Extended filter produced by -b flag

beratx:
It now produces two filter: one to eliminate possible paket peaks, other for host peaks.


kYroL01:
thanks

Issue #444, opened 2017-09-06
Telegram Application detection failed in FORWARD chain of iptables


TaranehKhalili:
I want to detect telegram protocol for web and it's application.
I could detect and drop web.telegram. But the telegram application can not detect by ndpi. HTTP_DOWNLOAD is the protocol which is detected by the ndpi.
How is the process of detecting a protocol? first protocol checked by files in /src/lib/protocols OR by entries in ndpi_content_match.c.inc (IP addresses and FQDNs) ?
@kYroL01  
why telegram is detect in INPUT chain but in FORWARD chain http_download is detect instead of telegram??

kYroL01:
First of all nDPI check IP address and Port range and THEN check eventually dissectors in protocols directory.
I tested with your pcap and everything is working correctly. So, did u tried to update and recompile all ?

kYroL01:
I close because IMHO now the bug is fixed (as I said above). Reopen if necessary

TaranehKhalili:
I think telegram dissector cannot detect telegram flows between the application and server because of the protocol which telegram used for transferring data (MTproto). 
IMHO the telegram dissector code is not complete.

kYroL01:
Ok I'll check asap

TaranehKhalili:
Thanx
I think the matter is when you detect a protocol based on ip address (for example 149.154.166.120 for telegram) you should not continue the detection process. because of this problem http_download is the protocol which is detected instead of telegram.

TaranehKhalili:
well I could solve the problem by this change in ndpi_detection_process_packet function.
` flow->guessed_protocol_id = (int16_t)ndpi_guess_protocol_id(ndpi_struct, protocol, sport, dport, &user_defined_proto);
    if(flow->packet.iph) {
      if((flow->guessed_host_protocol_id = ndpi_network_ptree_match(ndpi_struct, (struct in_addr *)&flow->packet.iph->saddr)) == NDPI_PROTOCOL_UNKNOWN)
        flow->guessed_host_protocol_id = ndpi_network_ptree_match(ndpi_struct, (struct in_addr *)&flow->packet.iph->daddr); 
      if (flow->guessed_host_protocol_id == NDPI_PROTOCOL_TELEGRAM){   
         ret.master_protocol = flow->guessed_host_protocol_id, ret.protocol =   flow->guessed_host_protocol_id;
         return(ret);
       }
   }
[...]

PR #453, opened 2017-09-25
Detect telegram protocol

TaranehKhalili:
related to issue #444


kYroL01:
THe pull request is not correct. There are conflicts that you can resolve before send the request. Look at that and send it again, thanks.

PR #466, opened 2017-10-01
Fix error: Use of uninitialized structures in tcp_udp.c

vel21ripn:
Now the test results do not depend on the compiler optimization level and architecture.


kYroL01:
Thanks!

PR #469, opened 2017-10-03
Fix undefined behavior (detected by gcc):

vel21ripn:
ndpi_define.h.in:218: Macros NDPI_SET(), NDPI_CLR(), NDPI_ISSET()
  runtime error: left shift of 1 by 31 places cannot be represented in type 'int'

http.c:364: strncpy((char*)flow->protos.http.nat_ip,(char*)packet->forwarded_line.ptr,len);
 runtime error: null pointer passed as argument 2, which is declared to never be null

ndpi_patricia.c:ndpi_comp_with_mask(): int m = ((-1) << (8 - (mask % 8)));
  runtime error: left shift of negative value -1

Original function ndpi_comp_with_mask() is unreadable and slow.
We have only 2 type of address: 32 bit and 128 bit. The optimized version works 6 times faster (IPv4).

### DIFF
--- a/src/include/ndpi_define.h.in
+++ b/src/include/ndpi_define.h.in
@@ -215,9 +215,9 @@
 #define howmanybits(x, y)   (((x)+((y)-1))/(y))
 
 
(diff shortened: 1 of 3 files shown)


kYroL01:
Huge work! Thanks so much for this optimization!

PR #475, opened 2017-10-16
netfilter added

TaranehKhalili:


TaranehKhalili:
well I use the nDPI from netfilter and I did'nt get the updated one becouse it couse errors for me. can U check it.

lucaderi:
@taranehk I do not think that putting nDPI in the kernel is a good practice. We will soon introduce a solution for this by combining netfilter, efficiency, and user-space. Stay tuned.

TaranehKhalili:
@lucaderi thnx a lot!
Can u explain me when ? I wait for it

PR #476, opened 2017-10-17
Refactoring the debugging output.

vel21ripn:
levels of debug output:
 0 - ERROR: Only for errors.
 1 - TRACE: Start of each packets and if found protocol.
 2 - DEBUG: Start of searching each protocol and excluding protocols.
 3 - DEBUG_EXTRA: For all other messages.

Added field ndpi_struct->debug_logging for enable debug output of each protocols.

Simple macros for debugging output are added:
NDPI_LOG__ERROR(), NDPI_LOG__TRACE(), NDPI_LOG__DEBUG(), NDPI_LOG__DEBUG2(),
NDPI_EXCLUDE_PROTO()

### DIFF
--- a/example/ndpiReader.c
+++ b/example/ndpiReader.c
@@ -77,7 +77,9 @@ static json_object *jArray_topStats;
 static u_int8_t live_capture = 0;
 static u_int8_t undetected_flows_deleted = 0;
 /** User preferences **/
-static u_int8_t enable_protocol_guess = 1, verbose = 0, nDPI_traceLevel = 0, json_flag = 0;
+static u_int8_t enable_protocol_guess = 1, verbose = 0, json_flag = 0;
+int nDPI_LogLevel = 0;
+char *_debug_protocols = NULL;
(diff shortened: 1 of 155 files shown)


lucaderi:
@vel21ripn Can you please remove all __KERNEL__ statements? We do not want to support nDPi in the kernel as we believe it is not a good idea as you might end up crashing the host when receiving a bad packet.

lucaderi:
@vel21ripn Also please change NDPI_LOG__XXX in NDPI_LOG_XXX (remove double __)

vel21ripn:
`__KERNEL__` must be removed. I forgot to rollback the my changes for ndpi-netfilter :(
We have ndpi_log_level_t  { NDPI_LOG_ERROR,NDPI_LOG_TRACE,NDPI_LOG_DEBUG }
I believe that the same names for macros and elements of enum are not a good idea.
There are several possible options: remove double underscore, rename macros, rename elements in enum. If you choose to rename, then offer your options.

PR #477, opened 2017-10-19
Add libcache to installed headers and removed relative import

hvanderheide:
My application (fastnetmon) refused to compile with the latest checkout of nDPI because of the relative import for libcache.h. This only works while building the source package but breaks after installing the header files to the system.

Also because the libcache header file wasn't added to the include headers, it didn't get copied to the system on install.


kYroL01:
@hvanderheide Thanks a lot. I'll check both the pull request and let u know. ;)

kYroL01:
Thank you :)

kYroL01:
I think the other pull could be closed right ?

PR #478, opened 2017-10-19
Add ndpi_includes.h to include headers

hvanderheide:
Not sure if you still accept patches to this branch but without this the header file doesn't get installed to the system.


hvanderheide:
@kYroL01 this pull request is for the 1.7-stable branch. I needed the fix since the software I tried to compile doesn't support newer versions (yet, its on my list to fix eventually). Other branches might need a similar fix. If you prefer to leave these branches untouched I'll close the request and keep using my fork.

PR #484, opened 2017-10-27
Added DHCP class identifier support.

dsokoloski:
For improved OS detection via a service like [https://fingerbank.org/](url), three metrics can be combined to significantly improve accuracy:

1. DHCP parameter request list (fingerprint).
2. DHCP client identifier (when available).
3. Full UA from HTTP (when available).

The client-identifier is stored in the DHCP structure under the protocols union and is of 48 bytes in length.  Combined with the existing fingerprint buffer, the DHCP structure now occupies 96 bytes.  In other words, there is no additional memory required to store this additional information in the flow structure.

### DIFF
--- a/src/include/ndpi_typedefs.h
+++ b/src/include/ndpi_typedefs.h
@@ -1031,6 +1031,7 @@ struct ndpi_flow_struct {
 
     struct {
       char fingerprint[48];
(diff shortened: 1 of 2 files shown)


kYroL01:
Many thanks for this!

PR #486, opened 2017-10-31
Bug fix. Protocol SIP: command 'CANCEL sip:' is not recognized.

vel21ripn:
Add sip.pcap for tests ( Source: https://wiki.wireshark.org/SIP ).


kYroL01:
Thanks!

PR #493, opened 2017-11-22
Fixed ndpi_win32.h so that it compiles on MinGW+GCC

smithjacobj:
The Windows build process seems relatively untested and I had to make a couple of changes to make it build on MinGW-w64 (autotools Perl include paths actually seems to be broken on base MinGW, so technically this is on MSYS2 in the MinGW-w64 toolchain, where autotools works without fiddling...)

I'm not sure I like the `__forceinline` redefine, but it works and is restricted to MinGW GCC builds. The only alternative I can think of is defining an `ndpi_forceinline` macro or similar, but that's a more sweeping change.

Known issue: the example still doesn't compile (while the example has a `WIN32` ifdef, it tries to include `sys/socket.h` which is not "and will never be" a part of MinGW - you're supposed to just use `winsock2.h` which is also included). It does on Cygwin but that's because it's a full compatibility layer with the associated baggage. I just ran `make` in the `src/libs` dir and copied the library files; I don't know if `make install` matters on Windows anyway.

### DIFF
--- a/src/include/ndpi_win32.h
+++ b/src/include/ndpi_win32.h
@@ -24,7 +24,18 @@
 #ifndef __NDPI_WIN32_H__
 #define __NDPI_WIN32_H__
 
+// fix a MinGW build issue "error: multiple storage classes in declaration specifiers" due to MinGW
+// defining extern for __forceinline types
+#if (defined(__MINGW32__) || defined(__MINGW64__)) && defined(__GNUC__)
+// MinGW winnt.h uses FORCEINLINE which is originally defined as __forceinline, but requires extern
+#undef FORCEINLINE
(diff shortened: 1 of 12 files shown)


alagoutte:
May be add a .appveyor(.yml) build ?

smithjacobj:
I had some time to think and decided on a strategy that fit the current setup better and had less side effects, as you can see in 950f5db

kYroL01:
@smithjacobj take your time. For now I let this pull request open in case you want to modify properly.
Thanks

smithjacobj:
@kYroL01 feel free to merge the PR if 950f5db is suitable :)

lucaderi:
The changes look good. Thank you

PR #496, opened 2017-11-28
Added dissector for protocol check_mk

cymaniac:


lucaderi:
Thank you. I see little issues I will fix.

PR #497, opened 2017-11-30
Return before calling ndpi_match_host_subprotocol when...

dsokoloski:
ndpi_flow.**dns_dissect_response** is enabled.

Otherwise responses will never be seen if a host sub-protocol matched (ndpi_flow->packet.detected_protocol_stack[0] != NDPI_PROTOCOL_UNKNOWN).

Dissecting DNS responses is disabled by default and is likely not used by many implementations.  It is possible this is a bug that has gone unnoticed.

### DIFF
--- a/src/lib/protocols/dns.c
+++ b/src/lib/protocols/dns.c
@@ -197,6 +197,9 @@ void ndpi_search_dns(struct ndpi_detection_module_struct *ndpi_struct, struct nd
 	off++;
       }
 
+	if(is_query && ndpi_struct->dns_dissect_response)
+	  return; /* The response will set the verdict */
+
       flow->host_server_name[j] = '\0';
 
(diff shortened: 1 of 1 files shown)


src/lib/protocols/dns.c
```diff
@@ -217,9 +220,6 @@ void ndpi_search_dns(struct ndpi_detection_module_struct *ndpi_struct, struct nd
 #endif
     
       if(flow->packet.detected_protocol_stack[0] == NDPI_PROTOCOL_UNKNOWN) {
-	if(is_query && ndpi_struct->dns_dissect_response)
-	  return; /* The response will set the verdict */
```
dsokoloski, line 221: > The response will set the verdict

Except the response will never be "seen" if there was a sub-protocol detected.

kYroL01:
You're right, thanks!

PR #498, opened 2017-12-05
Add ndpi_set_proto_category, which is actually used by ntopng.

madpilot78:
While testing the FreeBSD port update I got an error about missing symbols when compiling ntopng, so I think this should be added to the exported ones.


madpilot78:
This change should be merged to the 2.2-stable branch if accepted.

kYroL01:
Check. Thanks ;)

PR #501, opened 2017-12-19
Fix ndpiReader long run crash due to dupilicate nodes exist, cause double free failed.

qianguozheng:
Fix ndpiReader long run crash due to dupilicate nodes exist, cause double free failed.


kYroL01:
THanks @qianguozheng . I'll check asap!

kYroL01:
Thanks @qianguozheng

PR #506, opened 2018-01-09
Fix MPLS header parsing in ndpiReader.

cavedon:
Reported at https://bugs.debian.org/886133.
The current parsing for the MPLS header in examples/ndpi_util.c has
multiple issues:
- the bitfield order is incorrect for little endian architectures
- ntohl() is applied to a 20 bit label, which has unclear purpose
- if multiple labels are detected, the while loop parsing labels will
never exit due to a missing re-read of the mpls label
- the last label is identified by looking inside the label field, while
it should be done by looking at the S bit

This change fixes the above issues.
Notice that bitfield ordering is implementation-dependent, so C
bitfields should not be used in the first place to parse network
packets.


kYroL01:
Many thanks!

PR #519, opened 2018-02-04
Add an Xcode project for easier debugging on MacOS

zyingp:
There is a Visual Studio project for Windows and we add an Xcode project for MacOS now. With the Xcode project, it's very easy to set breakpoints, viewing/changing variables etc.

We don't modify any codes in the "src" folder so our commit is very safe. As we noted in the README.md under the added MacOS directory, we only modify the `ndpiReader.c` under "example" directory to change the name of  the `main` function to `original_main` when an `APP_HAS_OWN_MAIN` macro is defined (we defined the macro in Xcode project). Also, all the added files are put under the added MacOS directory.


lucaderi:
Thank you

PR #521, opened 2018-02-07
Making host_protocol_list[] from .yaml config files.

vel21ripn:
Moving definition of array host_protocol_list[] from file "ndpi_content_match.c.inc"
into separated file "ndpi_network_list.c.inc".
Added parser/generator for configuration files ndpi_network_list_std.yaml and ndpi_network_list_tor.yaml

When generating the list "host_protocol_list[]", the correspondence
between the address and the network mask is checked, as well as the
aggregation of ipv4 networks.


vel21ripn:
I divided the list into two parts. The ndpi_network_list_std.yaml file contains these data, which is rarely changed (all protocols except TOR). The data for the TOR protocol is in a separate file ndpi_network_list_tor.yaml, since this file is updated very often.
Perhaps it makes sense to store the data of each protocol in a separate file.
The parser can union the data from the list of files.

lucaderi:
Any chance to fix the conflicts?

lucaderi:
Closed for inactivity. Will reopen when @vel21ripn will resume this thread

vel21ripn:
I have a question: how do you build the ndpi_content_match.c.inc file?
For now, I see that for the Tor protocol, the data is being automated. Are there any other protocols for which address information is obtained by automated means?
Because of this, in the first version of PR, I separated the "Tor" protocol from all the others.
Question: maybe it makes sense to make a separate file for each protocol?
Question: maybe it is necessary to abandon the yaml format? Make a section at the beginning of the file with comments, and then a list of networks, one network per line?

PR #522, opened 2018-02-07
Added option --enable-debug-messages.

vel21ripn:
The configurator option "--enable-debug-messages" is added to enable
debug information output (define NDPI_ENABLE_DEBUG_MESSAGES=1).
Mandatory inclusion of the file ndpi_config.h in all the compiled files.

### DIFF
--- a/configure.seed
+++ b/configure.seed
@@ -109,6 +109,10 @@ AS_IF([test "x$enable_json_c" != "xno"], [
 	     ])
        ])
 
+AC_ARG_ENABLE([debug-messages],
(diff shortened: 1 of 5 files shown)


kYroL01:
Thank you

vel21ripn:
"--enable-debug-messages" is required only for debugging. IMHO ntopng does not need this.

Major problem: the size of the structures depends on the definition of NDPI_ENABLE_DEBUG_MESSAGES. The option for configure (ndpi_config.h) is the easiest way to get the same size of the main structures.

Suggest your solution to this problem.

PR #527, opened 2018-03-01
host server name bug fix

dillinger79:
this change seems to fix the problem , but i am not sure if it causes any other issues.


kYroL01:
Thanks!

PR #539, opened 2018-03-25
Fixed memory leak when flow is zeroed out.

felix-jia:


kYroL01:
Have to check, thanks

lucaderi:
Thank you

PR #543, opened 2018-04-03
Repaired wrong timestamp calculation.

vel21ripn:
Fixed incorrect type conversion.
  Hardcoded constant replaced by structure field.


kYroL01:
That's great! thanks!

PR #545, opened 2018-04-08
added ndpi_process_extra_packet to exported symbols

vpiserchia:
In order to use the function, it must be added to the exported symbols


kYroL01:
Thanks!

PR #547, opened 2018-04-16
Fix several bugs found by Address Sanitizer (ASan)

zyingp:
I tried to use Address Sanitizer (ASan, https://github.com/google/sanitizers) to detect bad memory access. I used the build-in Address Sanitizer setting in Xcode project to enable it (Product->Scheme->Edit Scheme->Diagnostics, check "Address Sanitizer").

I found several bugs after running against several pcap files. These bugs are all caused by reading memory that is out of the assigned bound, e.g., the packet payload length is 3 and we memcmp it to a 4-char C string.


lucaderi:
Thank you

PR #551, opened 2018-04-21
Add  Apache JServ Protocol Dissector

leonn:
Add AJP Protocol Dissector.

Detection based on header parse and protocol packet type.

https://tomcat.apache.org/connectors-doc/ajp/ajpv13a.html


kYroL01:
Great! Thank you!

PR #552, opened 2018-04-24
Moved ndpi_config.h from ndpi_main.h to ndpi_main.c...

dsokoloski:
... and example source files.

This fixes the changes made in commit 0624afd422b7fbdd2b481a299ac9bf62fa9ec706 where the autoconf-generated config was being included from the public API via ndpi_main.h.

Autoconf configs are not meant to be included into global scope.

### DIFF
--- a/example/ndpiReader.c
+++ b/example/ndpiReader.c
@@ -18,6 +18,10 @@
  *
  */
 
+#ifdef HAVE_CONFIG_H
+#include "ndpi_config.h"
+#endif
+
 #ifdef linux
 #define _GNU_SOURCE
 #include <sched.h>
--- a/example/ndpi_util.c
+++ b/example/ndpi_util.c
@@ -21,6 +21,10 @@
  *
  */
 
+#ifdef HAVE_CONFIG_H
+#include "ndpi_config.h"
+#endif
+
 #include <stdlib.h>
 
 #ifdef WIN32
--- a/src/include/ndpi_main.h
+++ b/src/include/ndpi_main.h
@@ -24,7 +24,6 @@
 #ifndef __NDPI_MAIN_H__
 #define __NDPI_MAIN_H__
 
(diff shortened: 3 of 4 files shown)


vel21ripn:
I would add such a correction:
```
--- src/include/ndpi_typedefs.h_     2018-04-15 19:45:21.324983884 +0300
+++ src/include/ndpi_typedefs.h    2018-04-24 20:20:32.742909465 +0300
@@ -896,7 +896,6 @@
 
   ndpi_log_level_t ndpi_log_level; /* default error */
 
-#ifdef NDPI_ENABLE_DEBUG_MESSAGES
   /* debug callback, only set when debug is used */
   ndpi_debug_function_ptr ndpi_debug_printf;
   const char *ndpi_debug_print_file;
@@ -907,8 +906,6 @@
   #define NDPI_IP_STRING_SIZE 48
   char ip_string[NDPI_IP_STRING_SIZE];
 
-#endif
-
   /* misc parameters */
   u_int32_t tcp_max_retransmission_window_size;
```
 If there is no dependence of the size of the ndpi_detection_module_struct structure from NDPI_ENABLE_DEBUG_MESSAGES, then mandatory inclusion of ndpi_config.h in all compiled files will cease to be necessary.

dsokoloski:
I agree with leaving the debug stuff in the ndpi_detection_module structure, it's only about 60 bytes of extra space...  however:

> then mandatory inclusion of ndpi_config.h in all compiled files will cease to be necessary.

...that's the way you're supposed to include the autoconf config.h header: in each and every source file that requires declarations from it.  It should never be included in another global/public header file (especially true for a library with a public API).  Personally, every project I've worked on does it this way.

Regardless, as long as I don't have to maintain additional hacks to get this to compile in my project, I'm on board either way.

kYroL01:
Tested succesfully and merged!
Thanks a lot

PR #554, opened 2018-04-24
Extract SSL/TLS version and Cipher Suite

dsokoloski:
For review:

This patch extracts the SSL/TLS version and Cipher Suite from the TLS Handshake/ServerHello packet.  We use this information to look for and warn about known, insecure versions/ciphers.

### DIFF
--- a/src/include/ndpi_typedefs.h
+++ b/src/include/ndpi_typedefs.h
@@ -1003,8 +1003,7 @@ struct ndpi_flow_struct {
     } ntp;
 
     struct {
-      u_int8_t ssl_version, tls_version;
-      u_int16_t cipher_suite;
+      u_int16_t version, cipher_suite;
       char client_certificate[48], server_certificate[48];
     } ssl;
 
--- a/src/include/ndpi_typedefs.h
+++ b/src/include/ndpi_typedefs.h
@@ -1003,6 +1003,8 @@ struct ndpi_flow_struct {
     } ntp;
 
     struct {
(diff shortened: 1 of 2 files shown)


kYroL01:
@dsokoloski .
I think this should be fixed. In this way, the detection avoid to check the application_protocol determined by  the host name. I.e. if you run test/./do.sh you can see it:

```
facebook.pcap                           ERROR                                                                                                                            
../example/ndpiReader -p ../example/protos.txt -q -i pcap/facebook.pcap -w /tmp/reader.out -v 1                                                                          
1c1,2                                                                                                                                                                    
< Facebook      60      30511   2
---
> SSL   41      24766   1
> Facebook      19      5745    1
3c4
<       1       TCP 192.168.43.18:44614 <-> 31.13.86.36:443 [proto: 91.119/SSL.Facebook][19 pkts/2664 bytes <-> 22 pkts/22102 bytes][client: www.facebook.com]
 ---
 >       1       TCP 192.168.43.18:44614 <-> 31.13.86.36:443 [proto: 91/SSL][19 pkts/2664 bytes <-> 22 pkts/22102 bytes][client: www.facebook.com]
```


I understand what you're doing, but IMHO if I have an host name `facebook.com` is better to identify all the TLS flow as SSL.facebook.
Tell me what you think about it.
THanksTh

dsokoloski:
Ok!  Yes I understand the problem.  I fixed it and submitted a new PR.  Let me know if the new PR works for you (it passes my tests now).

Good catch!

PR #555, opened 2018-04-26
Fixed SSL/TLS version/cipher extraction logic.

dsokoloski:
A certificate must be returned if one was found previously (client certificate) or else detection will be guessed.


dsokoloski:
Broken... Missing other changes.  Will re-submit.

PR #557, opened 2018-04-26
Clean-up/sort: .gitignore

dsokoloski:
Added missing generated files (ex: configure, configure.tmp, ndpi_config.h, etc), and sorted (cosmetic).


kYroL01:
Thank you

PR #558, opened 2018-05-04
SSH: increase client's maximum payload length from 100 to 500

gozzy:
According to RFC4253, a client may send additional data right after
its identification string (before receiving the server's identification
string). For instance, PuTTY sends supported ciphers. This exceeds
100 bytes and nDPI fails to detect such SSH sessions.


gozzy:
Attached a sample PCAP.

[trafficssh.zip](https://github.com/ntop/nDPI/files/1973530/trafficssh.zip)

gozzy:
I don't think that this PR is able to break anything because the payload size is increased after SSH stage 1 (server's identification string) is passed. Stage 1 remains unchanged.

kYroL01:
Thank you!

PR #565, opened 2018-05-21
SMTPS on port 587 is not detected and Updated AmazonVideo

dark-vex:
Added support for SMTPS on port 587


dark-vex:
I forgot to split the two pull request so there is the modification for the SMTPS on port 587 and the new CDN hostnames used by AmazonVideo
prime-video
prime-video1

lucaderi:
Thank you very much: looks good!

PR #575, opened 2018-07-17
fix after showmax addition

alm-A:
wrong NDPI_LAST_IMPLEMENTED_PROTOCOL


kYroL01:
@alm-A This branch have many conflits. If you see latest changes on nDPI, the numbers of protocols change from #define to enum.
Please, update code and in case resend the PR accordingly with that change. 
Thank you

alm-A:
@kYroL01 yeah i've seen the changes since pull request

no need to send it back even now:
- fixed in d9c963061bbe561af4ea1e223fc4589fe333570d
- then system changed to enum
- protocol converted to NDPI_PROTOCOL_GENERIC in https://github.com/ntop/nDPI/commit/667e77c609d842a9a41930b590de206fb2b917fe#diff-37f0ad94bbd32b9fbff7d75f1307c652R8387

& everything's OK :)

PR #576, opened 2018-07-20
Adjust the SSDP dissector's minimum packet length

dsokoloski:
The existing SSDP dissector misses most of the SSDP traffic on my network because it skips payloads less than 100 bytes.  The smallest SSDP packet payload appears to be a NOTIFY packet:

mlen = "NOTIFY * HTTP/1.1" + 0x0d + 0x0a
mlen = 17 + 1 + 1
mlen = 19 bytes

The 100-byte clamp also misses M-SEARCH packets (see the included pcap file).


lucaderi:
Thank you

PR #577, opened 2018-07-20
Fix end-of-line bounds handling.

dsokoloski:
The existing implementation misses ending lines and as a result, fails
to match certain protocols (some SMTP for example, which needs to see at
least 3 commain/response matches).  It appears from the commit history
that an attempt was made (end-1) to prevent reads past payload length.

This can be ensured by simply not processing payloads that are < 3 bytes.

The updated logic for this loop is:
- Payload length is >= 3 bytes, or return.
- Loop over payload, compare for EOL (CR + NL) sequence.
- If found, process string.
- If index 'a' plus two is less than payload length, increment 'a' by
  one and continue.
- Loop return always increments index 'a' by one.


lucaderi:
Thank you

PR #585, opened 2018-08-17
Remove autoconf include from typedefs header.

dsokoloski:
Including this file in any "public" API header breaks all projects that also use autotools because macros such as PACKAGE_VERSION will be redefined.

### DIFF
--- a/src/include/ndpi_typedefs.h
+++ b/src/include/ndpi_typedefs.h
@@ -26,9 +26,6 @@
 
 #include "ndpi_define.h"
 
-/* Needed to have access to HAVE_* defines */
-#include "ndpi_config.h"
-
 /* NDPI_LOG_LEVEL */
 typedef enum {
   NDPI_LOG_ERROR,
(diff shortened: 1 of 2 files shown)


kYroL01:
@dsokoloski I see a lot of errors if i check this PR. Can u check again ? Thanks.

`error: ‘struct <anonymous>’ has no member named [...]` in main_ndpi.c

kYroL01:
Thank you!

simonemainardi:
unfortunately this isn't a real fix because the use of macro `#ifdef HAVE_HYPERSCAN` hasn't been removed from src/include/ndpi_typedefs.h https://github.com/ntop/nDPI/blob/06befec642103bb4b20be3ac9227fe42d92fde1b/src/include/ndpi_typedefs.h#L914

This prevents nDPI from compiling with hyperscan support. @kYroL01 please check and make sure to allow hyperscan builds with nDPI by properly reposition that macro. Thank you.

PR #586, opened 2018-08-17
Include <sys/types.h> for u_int16_t.

dsokoloski:
On various embedded environments (ex: OpenWrt, EdgeOS), <sys/types.h> must be included for u_intX_t types.  The Hyperscan changes made recently introduced a u_int16_t member in actypes.h which is undeclared in certain environments without this explicit include.


lucaderi:
Thank you

PR #589, opened 2018-08-18
parse protocol's string_to_match into hex which could be match by hyperscan

gou4shi1:
if one protocol do not have it's regex (pattern_to_match)
then parse it's string (string_to_match) into hex
so we can match every protocol with hyperscan
and we do not need to [add "\\\\" manually](https://github.com/ntop/nDPI/commit/d2bc3ea20927ace331abcd8b939cb6e74fbc8ce1).

### DIFF
--- a/src/lib/ndpi_main.c
+++ b/src/lib/ndpi_main.c
@@ -817,7 +817,7 @@ static char* string2hex(const char *pat) {
 
   patlen = strlen(pat);
   hexbuf = (char*)calloc(sizeof(char), patlen * 4 + 1);
-  if(!hexbuf) return(-1);
+  if(!hexbuf) return(NULL);
 
   for (i = 0, buf = hexbuf; i < patlen; i++, buf += 4) {
     snprintf(buf, 5, "\\x%02x", (unsigned char)pat[i]);
@@ -828,7 +828,7 @@ static char* string2hex(const char *pat) {
 }
 
 static int init_hyperscan(struct ndpi_detection_module_struct *ndpi_mod) {
-  u_int num_patterns = 0, i;
+  u_int num_patterns = 0, i, j;
   char **expressions;
   unsigned int *ids;
   unsigned char *need_to_be_free;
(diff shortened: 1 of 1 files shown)


lucaderi:
@simonemainardi Can you please review this pull request

simonemainardi:
@gou4shi1 do you think your function `string2hex` can always be used? That is both when parsing the regex as well as when parsing a regular string. In this way we could avoid using the double backslash to escape points in file ndpi_content_match.c.inc

If affirmative, please, try and create a pull request for that.

Also, make `string2hex` static to reduce the scope of it and make sure it is always defined when hyperscan is enabled.


Thank you for your contribution.

gou4shi1:
@simonemainardi
 "." will be parsed to "\x2E" which will lost the semantic meaning of dot
I will make the function static later.

gou4shi1:
@simonemainardi so...besides "add static", are there any problem of this PR?
I just make another PR with "static"?
I don't think that "string" and "regex" and be handled in a same way.
each rule in `ndpi_content_match.c.inc` can have string_to_match or have regex_to_match, 
string_to_match -> hex -> hyperscan
regex_to_match -> hyperscan
I means...we don't need to add regex_to_match from string_to_match by manually add "\\"

simonemainardi:
don't do another pull request, just commit on the same branch so this pull request will be updated automatically.

Make sure to update https://github.com/ntop/nDPI/pull/589/files#diff-414f7c417eb4be3896935c36a0972eb7R819 and check it's status as calloc could have failed.

gou4shi1:
@simonemainardi 
Modified.
Thanks for your review.

simonemainardi:
uhm

if `string2hex` fails you end up putting a -1 in `expressions[num_patterns]` and eventually try to free it resulting in a crash...

gou4shi1:
oh...I just "copy and paste" from the below line…forgot to check the return value of `string2hex`...

simonemainardi:
thank you

Issue #593, opened 2018-08-24
Compilation and Instalation


dwairlk:
Hi all,
I am compiling and installing nDPI in Ubuntu 14.04 Desktop, kernel 4.15.0-32-generic, x86_64. I am following the steps at https://www.ntop.org/guides/pf_ring/ft.html and https://www.ntop.org/pf_ring/how-to-accelerate-suricata-bro-snort-with-pf_ring-ft/
The compilation works fine but it is generating an error during installation.
Please, find below the outputs
`./autogen.sh
autoreconf: Entering directory `.'
autoreconf: configure.ac: not using Gettext
autoreconf: running: aclocal --force -I m4
autoreconf: configure.ac: tracing
autoreconf: running: libtoolize --copy --force
libtoolize: putting auxiliary files in '.'.
libtoolize: copying file './ltmain.sh'
libtoolize: putting macros in AC_CONFIG_MACRO_DIRS, 'm4'.
libtoolize: copying file 'm4/libtool.m4'
libtoolize: copying file 'm4/ltoptions.m4'
libtoolize: copying file 'm4/ltsugar.m4'
libtoolize: copying file 'm4/ltversion.m4'
libtoolize: copying file 'm4/lt~obsolete.m4'
autoreconf: running: /usr/bin/autoconf --force
autoreconf: running: /usr/bin/autoheader --force
autoreconf: running: automake --add-missing --copy --force-missing
configure.ac:7: installing './compile'
configure.ac:5: installing './missing'
example/Makefile.am: installing './depcomp'
autoreconf: Leaving directory `.'
checking for a BSD-compatible install... /usr/bin/install -c
checking whether build environment is sane... yes
checking for a thread-safe mkdir -p... /bin/mkdir -p
checking for gawk... gawk
checking whether make sets $(MAKE)... yes
checking whether make supports nested variables... yes
checking build system type... x86_64-pc-linux-gnu
checking host system type... x86_64-pc-linux-gnu
checking how to print strings... printf
checking for style of include used by make... GNU
checking for gcc... gcc
checking whether the C compiler works... yes
checking for C compiler default output file name... a.out
checking for suffix of executables...
checking whether we are cross compiling... no
checking for suffix of object files... o
checking whether we are using the GNU C compiler... yes
checking whether gcc accepts -g... yes
checking for gcc option to accept ISO C89... none needed
checking whether gcc understands -c and -o together... yes
checking dependency style of gcc... gcc3
[...]

Issue #590, opened 2018-08-22
Error to create the Tar Ball on 2.4


lucasbaile:
I am trying to create a tarball for the 2.4 release, following exactly the steps on the README.md:

```
./autogen.sh
./configure
make
```

All the tests works as well:

```
cd tests; ./do.sh
```

But when I try to run the `make dist` at the end I get the following error:

```
make  dist-gzip am__post_remove_distdir='@:'
make[1]: Entering directory `/home/zerum/nDPI-2.4'
if test -d "libndpi-2.4.0"; then find "libndpi-2.4.0" -type d ! -perm -200 -exec chmod u+w {} ';' && rm -rf "libndpi-2.4.0" || { sleep 5 && rm -rf "libndpi-2.4.0"; }; else :; fi
test -d "libndpi-2.4.0" || mkdir "libndpi-2.4.0"
 (cd src/lib && make  top_distdir=../../libndpi-2.4.0 distdir=../../libndpi-2.4.0/src/lib \
     am__remove_distdir=: am__skip_length_check=: am__skip_mode_fix=: distdir)
make[2]: Entering directory `/home/zerum/nDPI-2.4/src/lib'
make[2]: *** No rule to make target `distdir'.  Stop.
make[2]: Leaving directory `/home/zerum/nDPI-2.4/src/lib'
make[1]: *** [distdir] Error 1
make[1]: Leaving directory `/home/zerum/nDPI-2.4'
make: *** [dist] Error 2
```

I already have all the prerequisites installed, and I can create the tarball for the 2.2 release just fine, the problem seems to be with the 2.4 release

dsokoloski:
@lucaderi removed the automake file from ./src/lib because:

> Auto* tools are a nightmare, that's why

So now most of the handy targets that automake provided are missing (*dist* for example).

You can restore the automake system by doing the following:

1. Move the existing Makefile out of the way: `mv ./src/lib/Makefile ./src/lib/Makefile.simple`
2. Edit configure.seed, and add `./src/lib/Makefile` to *AC_CONFIG_FILES*.
3. Download [this](https://gitlab.com/netify.ai/public/netify-agent-libs/ndpi/raw/master/src/lib/Makefile.am) Makefile.am, saving it into: `./src/lib/Makefile.am`
4. Edit the above file, remove `-static` from *libndpi_la_LDFLAGS* (you likely don't want that).
5. Regenerate everything: `./autogen.sh && ./configure`

Easy eh? :)

[...]

Issue #588, opened 2018-08-18
nDPI compilation failure


marcopp2018:
Hi,
 
  I tried to install nDPI using the following commands:
    cd ~
    cd nDPI
    git clone https://github.com/ntop/nDPI.git
    ./autogen.sh
    make && make install

all commands worked as expected except the make install:

Making install in src/lib
make[1]: Entering directory '/root/nDPI/src/lib'
make[1]: *** No rule to make target 'install'.  Stop.
make[1]: Leaving directory '/root/nDPI/src/lib'
Makefile:468: recipe for target 'install-recursive' failed
make: *** [install-recursive] Error 1

Using 2.2.2 branch works fine

genterminl:
I have the same problem, except I did git clone first, then cd nDPI, ./autogen.sh, make, make install.  
I don't know if it's relevant, but in the main directory, there is a symlink lib pointing to a non-existent src/lib/.libs.  I also note that README.md says to do ./autogen, ./configure, make.  Since autogen already runs configure, isn't that redundant?  Also, there is not any instruction to do "make install" so is it missing, or am I missing something about how this should get installed somewhere where it can be found when compiling ntop.
(I have the same problem with master, 2.2-stable, and 2.2.2-stable.)
Edit: I think the docs need improving.  I finally found in the ntopng/doc/README.compilation a note that implies compiling ntopng will find nDPI, so I suppose it doesn't actually need to be installed.

kYroL01:
@marcopp2018 There's no error on compilation. The compilation works well.
Remember that nDPI is a library, so you cannot "install" it, but you need a program that call nDPI, as [https://github.com/ntop/ntopng](url) .
So as @genterminl said in the Edit, nDPI is found by other programs that use it. Alternatively you can build your own program that call nDPI API.

p.s. @genterminl yes, you're right about ./configure.

marcopp2018:
@kYroL01  
I was trying to use ftflow from the pfring_ft example.

The following commands works without any issue:
    ftflow -i eth0
When using layer 7 commands it fails:

ftflow -i eth0 -7
#########################################################################
 ERROR: You do not seem to have a valid PF_RING FT 7.3.0.180817 license [FF7B9D9E760669E9]
 ERROR: Please get one at http://shop.ntop.org/.
#########################################################################
 We're now working in demo mode with packet capture and
 transmission limited to 5 minutes
#########################################################################
*** ndpi_load_hostname_category not found, please install latest ndpi library ***
[...]

PR #594, opened 2018-08-24
Restored/repaired automake configuration.

dsokoloski:
Considering the Issues (#593, #590, #588) that have been reported regarding the removal of `src/lib/Makefile.am`, here is a PR for consideration to restore the automake configuration.

The existing `src/lib/Makefile` has been moved back to `src/lib/Makefile.simple` which can be called manually for anyone wishing to use it (which is likely not many).


lucaderi:
We have decided to avoid automake in src/lib as this has issues on embedded platforms due to the large number of tools it needs. https://github.com/ntop/nDPI/commit/a98c0a81be86936e96be7d2312ae11ef85b785b2 should fix the issue you reported

dsokoloski:
@lucaderi I haven't reported any issues related to this.  I have no horse in this race as I maintain my own fork.  I submitted this PR to help the community get back a working build system because there is a fairly steady stream of related issues being opened.

Just curious which embedded systems you are building for?  I cross-compile nDPI for all kinds of embedded environments and I have yet to have a problem with autotools.  In fact, it would be a nightmare to do without.

It almost sounds like you're compiling **on** an embedded system with a limited toolset/build environment, which would be very odd way to go about it.

Autotools is pretty much ubiquitous.  IMHO it would be easier for nDPI and nicer for the community to maintain a custom Makefile for those few cases where you can't use autotools.  :shrug:

PR #595, opened 2018-08-25
- Respect CC variable, avoiding hardcoding gcc which for example is n…

madpilot78:
…ot present in FreeBSD by default.

- Provide a SONAME for the library. Required by FreeBSD pkg management tool to track libraries properly.


lucaderi:
This I have fixed it already: thank you

PR #596, opened 2018-08-25
Add back pkgconfig file. It is required by ntopng to find the install…

madpilot78:
…ed shared library.

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -1,3 +1,7 @@
 ACLOCAL_AMFLAGS = -I m4
 SUBDIRS = src/lib example tests
+
+pkgconfigdir = $(prefix)/libdata/pkgconfig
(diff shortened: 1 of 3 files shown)


madpilot78:
Actually an inconsistency remains with regard to what this software used to do before.

Old version installed include files in PREFIX/include/libndpi-%VERSION%/libndpi, but reported only PREFIX/include/libndpi-%VERSION% in pkgconfig file. ntopng added the libndpi path.

This version installs in PREFIX/include/ndpi directly. I added this path to the pkgconfig file (which IMHO is the correct choice), but ntopng now needs an hint using the --with-ndpi-includes configure option to find them.

I'm not patching for this since it requires a coordinated choice in the two ports to be fixed.

lucaderi:
Thank you

PR #598, opened 2018-08-27
Move the configure include file inclusion and code depending on it in…

madpilot78:
… code protected by the NDPI_LIB_COMPILATION define, this should avoid it polluting the environment when including this file from ntopng.

This should fix the problem with configure generated include file being leaked when building ntopng, while keeping the same logic when building nDPI.

I have only compile tested this on FreeBSD. I need to run test this, which I will be able to do tomorrow.

In the while please review the patch, if the approach I've chosen is acceptable.

Thanks!

### DIFF
--- a/src/include/ndpi_typedefs.h
+++ b/src/include/ndpi_typedefs.h
@@ -854,6 +854,22 @@ typedef struct ndpi_proto {
 
 #ifdef NDPI_LIB_COMPILATION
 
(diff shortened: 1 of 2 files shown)


madpilot78:
I just tested the stable version of nDPI and ntopng with this change included on FreeBSD and it works fine here.

lucaderi:
Thank you

PR #602, opened 2018-09-07
Adding Entertainment and Video Portals protocols to Configuration

kumarks1122:


lucaderi:
Sorry you have to implement this with protocol categories. This seems to be a regional protocol

PR #603, opened 2018-09-09
Avoid possible NULL pointer dereference in ndpi_detection_process_packet

pavlosantoniou:
The flow pointer is checked for NULL in lines 4234 and 4236, thus implying the possibility of a passing a NULL pointer into the function.
To avoid a possible null pointer dereference, the access of the num_processed_packets member is moved after the check.


lucaderi:
Thank you

PR #606, opened 2018-09-11
Prevent missing text lines and invalid reads past end-of-buffer.

dsokoloski:
The recent revert commit 6373c355cfd497d57609b8ecfc62717758ac7154 applied to ndpi_parse_packet_line_info resurrects an old bug where the lines that are aligned with the end of the packet buffer will not be parsed.  The commit is an attempt to prevent invalid reads past the end of the packet buffer.  This PR moves the end-of-bounds test to before the 16-bit read and returns if true.  This fixes the case where a text line ends aligned to the end of the buffer, and it fixes the invalid read issue.

Here is a link to source code that demonstrates the problem: [test.c](http://sokoloski.ca/parse-line/test.c)

Here is a link to the Valgrind output from the above test program: [valgrind.log](http://sokoloski.ca/parse-line/valgrind.log)


lucaderi:
Thank you

PR #610, opened 2018-09-20
New dissector: Nest Log Sink

dsokoloski:
This is a dissector for the Nest Log Sink protocol which is used by the [Nest Protect](https://nest.com/ca/smoke-co-alarm/overview/).


lucaderi:
Thank you

PR #613, opened 2018-09-30
add youku app.

wang-xuefei:
1.Add youku apps protocol.
2.fix remake install and remake clean failed

### DIFF
--- a/example/Makefile.in
+++ b/example/Makefile.in
@@ -13,8 +13,8 @@ ndpiReader: $(OBJS) $(LIBNDPI)
 	$(CC) $(CFLAGS) -c $< -o $@
 
 install:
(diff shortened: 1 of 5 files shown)


kYroL01:
Hi @wang-xuefei .
1. I think this is a custom protocol and we cannot consider it as a usefull protocol to have a dissector. You just check for a server name. There is no DPI.
To manage this kind of protocol, nDPI provides the [_categories_](https://github.com/ntop/nDPI/blob/7117f0532c6229a2057d5f7cf1f0aa23abeb0ec8/src/include/ndpi_typedefs.h#L758), and u can use it.
2. Send a new PR with this fixes ( _fix remake install and remake clean failed_)

Thank you

PR #614, opened 2018-09-30
Do not use the available clang as the default compiler.

vel21ripn:
You can use the CC environment variable when running autogen.sh and configure.

### DIFF
--- a/configure.seed
+++ b/configure.seed
@@ -51,12 +51,6 @@ if test $SHORT_MACHINE = "arm"; then
 else
      AC_CHECK_LIB([numa], [numa_available], [LIBNUMA="-lnuma"])
 fi
-
-if test -z `which clang`; then
-CC=gcc
-else
-CC=clang
-fi
  
 HS_LIB=
 HS_INC=


kYroL01:
Hi @vel21ripn .
Is there any particular issue on using clang as default compiler ?

vel21ripn:
1) The default compiler in the system is defined by the symbolic link /usr/bin/cc and configure knows about it. 
2) You forgot about those who use cross-compilers.
3) At the moment, the kernel module is built only by the gcc compiler.
4) The developer should be able to explicitly specify which compiler to use.

**IMHO** Ignoring the CC environment variable in autogen.sh is a bad solution (it requires an additional configure run).

hlein:
+1 on this change.

Letting someone specify they want clang with env var or some other means is all well and good.

Forcibly changing CC=clang just because clang is found - 82% of the way through the configure script, as it happens - is a recipe for mysterious failures and therefore misleading bug reports.  Like the one I was about to file about an erroneous configure test failure, before I found that configure switched from CC=gcc to CC=clang partway through the run, and then this PR...

lucaderi:
Do you want to restrict this only to MacOS?

hlein:
> Do you want to restrict this only to MacOS?

Restrict what - switching over to clang?  And keep the default compiler on anything that is not MacOS?

Selfishly: that would be fine with me because I don't use MacOS, so on my platforms, it would be fixed ;)

More generally: I just don't see why "implicitly switch from cc to clang if clang is found" is a good idea?  Users with nonstandard compiler choices will be setting CC= or remapping /usr/bin/cc or something else.  In my case (on Gentoo Linux) I have both gcc (multiple versions) and clang installed, but /usr/bin/cc and /usr/bin/gcc point to the gcc version I wish to use for most everything, only clang for some packages that want it - and they handle their preference carefully.

More specifically: if you want to keep preferring clang, please do the check+switch as early as possible during configure.  Currently, configure makes it through lots of tests (including ironically, testing if CC is set to clang) before switching.  As it happens, my default CFLAGS contain some gcc-ism arguments that clang doesn't recognize.  If configure blew up immediately, on its first test-compile, it'd be clear something fundamental was wrong.  Instead, the switch happened after many "checking for..." tests had passed, and configure died on some random library check (live_pcap_open in lpcap, if I recall correctly).  Unfortunately, I don't know autoconf well enough to know how to change configure.seed->configure.ac to make that clang switch happen sooner.

lucaderi:
Thank you

PR #622, opened 2018-10-09
Use a more standard path for pkg-config files.

### LINKED ISSUE #621: Hardcoded pkgconfigdir is inappropriate on (at least?) most Linuxes
Makefile.am specifies where pkg-config files should go:

`pkgconfigdir = $(prefix)/libdata/pkgconfig`

This will typically resolve to /usr/libdata/pkgconfig or /usr/local/libdata/pkgconfig

That seems to be a/the common path on FreeBSD, but those paths aren't typically used on Linux systems I've looked at (Debian derivatives, and Gentoo). Linuxes appear to use/prefer /usr/lib{,32,64}/pkgconfig or /usr/share/pkgconfig, and not use /usr/{,local/}libdata/pkgconfig at all.

I haven't found a good canonical reference, but most packages I've looked at appear to use ${LIBDIR}/pkgconfig as their target by default.  Some support a --with-pkgconfigdir configure argument.

hlein:
The Linuxes I've checked all seem to use /usr/{,local/}lib*/pkgconfig,
not /usr/libdata/.  I think FreeBSD does use /usr/libdata/ but also
supports /usr/lib*/pkgconfig, although I have not tested this change
on a FreeBSD system.

This should fix #621.

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -1,7 +1,7 @@
 ACLOCAL_AMFLAGS = -I m4
 SUBDIRS = src/lib example tests
 
-pkgconfigdir = $(prefix)/libdata/pkgconfig
+pkgconfigdir = $(libdir)/pkgconfig
 pkgconfig_DATA = libndpi.pc
 
 EXTRA_DIST = autogen.sh


lucaderi:
Thank you

hlein:
Thanks for looking into this & the other issues/PRs I raised so quickly!

PR #624, opened 2018-10-09
Correct the include file path in -I CFLAGS argument.

hlein:
src/lib/Makefile.in install headers into ${prefix}/include/ndpi,
but the shipped libndpi.pc.in specifies a different path.  Therefore
anything trying to build using $(pkg-config --cflags libndpi) will
fail to find the nDPI headers.

This patch simply makes libndpi.pc.in agree with src/lib/Makefile.in.

### DIFF
--- a/libndpi.pc.in
+++ b/libndpi.pc.in
@@ -7,4 +7,4 @@ Name: libndpi
 Description: deep packet inspection library
 Version: @VERSION@
 Libs: -L${libdir} -lndpi
-Cflags: -I${includedir}/libndpi-@VERSION@
+Cflags: -I${includedir}/ndpi


lucaderi:
In Makefile.in I see
includedir = ${prefix}/include/ndpi

Why is your patch required?

hlein:
Oh, I worded that poorly.  The problem is not that src/lib/Makefile.in is wrong about includedir, the problem is that libndpi.pc.in is wrong - does not agree with Makefile.in.  This patch fixes libndpi.pc.in.

madpilot78:
This is already present in the dev branch. See commit d31a29b.

Maybe it should be cherry picked to stable?

lucaderi:
Thank you

PR #632, opened 2018-10-29
Add SONAME property to shared library

madpilot78:
Added command in the Makefile to correctly populate the SONAME property of the shared object.

The command should work fine with both clang and GCC.


madpilot78:
I'm investigating how to fix the build on the apple travis box.

madpilot78:
I fixed the failure. I'm open to comments or change requests.

lucaderi:
Thank you

PR #636, opened 2018-11-08
Improve SSL detection, prevent false positive SSL detection

miksyn:
`flow->protos` is a union: `flow->protos.stun_ssl.ssl.client_certificate == flow->protos.http.detected_os` (for example). In this case need additional check.


lucaderi:
Thank you

PR #642, opened 2018-11-28
Pull buildsys improvements

vitalivanov:
This is at least to build with DPDK included.


vitalivanov:
@lucaderi @cardigliano please check
In general build system is completely broken. Seems like a long time ago it was based on autotools but after that was not properly handled.

I created bug #643 to track that.

lucaderi:
Thank you

PR #647, opened 2018-12-14
invoke realloc 

wang-xuefei:
Fix the error caused by the return of NULL by realloc function.


lucaderi:
@wang-xuefei Sorry for the delay merging this patch. Unfortunately it does not work with the current code. Can you update it or do you want me to do manual merge?

PR #649, opened 2018-12-20
Bank website detected as TOR from SSL Dissector (bigram check)

dark-vex:
vpbank.com is detected as TOR traffic by SSL Dissector


lucaderi:
Thank you

PR #654, opened 2019-01-12
Update radius.c to RFC2865

u-devel:
According to RFC2865 code can have value up to 13, also payload length ranges should be applied.


lucaderi:
Thank you

PR #655, opened 2019-01-12
FIX H.323 broken detection

u-devel:
TPKT header length field can have value more then 255, and in fact in all of the cases I've met it is more then 255.
Thus checking real H.323 packet like this:
(packet->payload[2] == 0x00)
stop detecting H.323 conversation at all.


lucaderi:
Thank you

PR #659, opened 2019-01-16
Compute packet_direction consistently independently from endianness

cavedon:
Ensure packet_direction is computed in the same way on little endian and
big endian architectures.
This change will convert IP addresses and port from little endian to host
endian (instead of converting from network endian to host endian) so
that it does not change the behavior on little endian architecture where
ndpi is usually developed.
A better (but more invasive) change would be to use the ntoh*()
functions and then adjust all affected tests.


lucaderi:
Thank you

PR #661, opened 2019-01-20
Propagate prefix settings to Makefiles

pcein:
A "make install" was failing because the --prefix flag setting was not being propagated to the Makefiles.


lucaderi:
Thank you

ahkok:
Can we get a release with this? It's painful enough it took me 20 minutes of my time before I found this.

PR #662, opened 2019-01-24
Use ChangeLog date instead of build date

bmwiedemann:
in order to make builds reproducible.
See https://reproducible-builds.org/ for why this is good.

This date call works with GNU date and BSD date.
Also use UTC/gmtime to be independent of timezone.

This PR was done while working on reproducible builds for openSUSE.


bmwiedemann:
ping. Is there anything I can do to help get this simple patch merged?
We could also do alternative patches using `SOURCE_DATE_EPOCH`

lucaderi:
Thank you

Issue #607, opened 2018-09-13
Duplicate protocol name


spinenkoia:
The Vidto protocol is named PPStream.

spinenkoia:
@lucaderi

lucaderi:
Thank you

PR #666, opened 2019-01-31
fixing typo in ndpi_init_protocol_defaults

AndrewKarelin:
It was founded by spinenkoia. see https://github.com/ntop/nDPI/issues/607.


spinenkoia:
@lucaderi

PR #673, opened 2019-02-27
More accurate hostname verification in a DNS query.

vel21ripn:


lucaderi:
Thank you for the fix. However I think that at some point we need to implement better DNS parsing instead of using heuristics.

PR #675, opened 2019-03-06
Respect CXX variable during configure

chiehmin:
commit 6ac83ce7eb99 use CXX to link ndpiReader however CXX is not
properly set in configure step

Fixes: 6ac83ce7eb99 ("Fixes broken hyperscan compilation")


lucaderi:
Thank you

PR #677, opened 2019-03-12
Add NDPI prefix for HTTP_METHOD enum to avoid name collisions

chiehmin:
ref: https://stackoverflow.com/questions/35380279/avoid-name-collisions-with-enum-in-c-c99


lucaderi:
Thank you, good fix!

PR #678, opened 2019-03-12
The le32toh() function used in some places on BSD OSes requires sys/en…

madpilot78:
…dian.h to be included.


lucaderi:
Thank you

PR #679, opened 2019-03-13
Use relative path for library link

chiehmin:
relative link path can preserve link relation after packaging and
doplying to new machine

ex:
$ make DESTDIR=/tmp/ndpi install
$ tar czf ndpi.tar.gz /tmp/ndpi
$ tar xf ndpi.tar.gz -C /lib

Orignal install script will caused the /lib/libndpi.so.2 links to
/tmp/ndpi/libndpi.so.2.9.0 which is not intended


lucaderi:
Thank you. This is a good point I wanted to fix for a while but had no time. Glad to see you did it.

PR #682, opened 2019-03-26
Parse HTTP PATCH method

chiehmin:
HTTP PATCH is introduced in RFC 5789 (https://tools.ietf.org/html/rfc5789)


lucaderi:
Thanks

PR #691, opened 2019-04-05
Fix double-strdup memory leak in ndpi_handle_rule

pallas:
proto is being strdup'd both in the call to ndpi_set_proto_defaults and
inside of that function as well, leaking the memory.

### DIFF
--- a/src/lib/ndpi_main.c
+++ b/src/lib/ndpi_main.c
@@ -2837,7 +2837,7 @@ int ndpi_handle_rule(struct ndpi_detection_module_struct *ndpi_mod,
 			      ndpi_mod->ndpi_num_supported_protocols,
 			      0 /* can_have_a_subprotocol */, no_master,
 			      no_master,
-			      ndpi_strdup(proto),
+			      proto,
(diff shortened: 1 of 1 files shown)


emanuele-f:
This can be reproduced with ndpiReader with the `-p` option:

```
==30598== 5 bytes in 1 blocks are definitely lost in loss record 1 of 1
==30598==    at 0x483777F: malloc (in /usr/lib/valgrind/vgpreload_memcheck-amd64-linux.so)
==30598==    by 0x121B4B: ndpi_strdup (ndpi_main.c:498)
==30598==    by 0x126FB4: ndpi_handle_rule (ndpi_main.c:2825)
==30598==    by 0x127455: ndpi_load_protocols_file (ndpi_main.c:2943)
==30598==    by 0x11A101: setupDetection (ndpiReader.c:1483)
==30598==    by 0x11F178: test_lib (ndpiReader.c:2629)
==30598==    by 0x1190E9: main (ndpiReader.c:3317)
```

lucaderi:
Thank you

PR #702, opened 2019-05-09
 MomoChat detection signature

mmanoj:
Please arrange to merge the subject change


mmanoj:
Please arrange to merge the latest files uploaded.

mmanoj:
Dear nDPI Team,

Any update on merge above change.

mmanoj:
Dear Team,
Please arrange to merge this fix.

lucaderi:
I prefer to keep off nDPI protocols that are not that popular, and leave in the engine only the popular protocols. You can define (as we do for many EU protocols not used elsewhere) a protocols.txt file and put it there. This way the engine is kept compact while leaving people the freedom to add new protocols.

mmanoj:
@lucaderi 

Thanks for the advice and way forward. I will use as instructed. If any other protocols/services in your list, please let me know to contribute.

PR #704, opened 2019-05-13
Adding 2 missing Whatsapp IPs

clebig:
I discovered 2 IPs that were not caught by ndpi so I added them into ndpi_content_match.c.inc.

I also asked Whatsapp to add them to their public list, see https://github.com/WhatsApp/WhatsApp-Business-API-Setup-Scripts/issues/2#issue-443425815


lucaderi:
Thank you

PR #730, opened 2019-06-28
Alpine linux: "ln -F" not supported in alpine

hoermto:
removed -F option from makefile for alpine support. Actually not sure the relevance of -F flag for other OS. Works for Ubuntu 18 and Centos 7


lucaderi:
Thank you

PR #731, opened 2019-06-28
Stability Fixes

themadprofessor:
Fix some potential NULL derefs and out-of-bounds reads.


lucaderi (changes requested):
IMHO if you put them at the beginning of the file, people accessing flow->something will still access a pointer that is invalid after the last packet has been processed. I don't like this patch. Please explain why it is good

themadprofessor:
The null checks are at the start of functions which already had null checks interspersed within them. Having the null checks early means minimal processing is done if the flow is null. If the flow is null, it doesn't matter which packet it is as there is no flow to work with.

lucaderi:
Thank you

PR #734, opened 2019-07-12
fixed the printing layout, even with IPv6

glpaparelli:
We fixed the print alignment (and now it works with both IPv4 and IPv6) and the "counter" on the left that now counts the different hosts (or ja3) and it is not anymore a line counter.

### DIFF
--- a/example/ndpiReader.c
+++ b/example/ndpiReader.c
@@ -1421,7 +1421,7 @@ static void node_idle_scan_walker(const void *node, ndpi_VISIT which, int depth,
   }
 }
 
-/* *********************************************** */  
+/* *********************************************** */
 
 /**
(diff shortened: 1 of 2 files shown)


glpaparelli:
We have pinted the information the information about an host and or about a ja3 fingerprint

- with verbose = 1 it is printed the list of host with the number of ja3 associated 
- with verbose = 2 it's printed, for each host, the associated JA3 with the contacted server name (if available) and the security status. It is also printed the same information using the ja3 fingerprint to aggragate the host (so for each ja3 fingerprint the host with that ja3, ...) 

Now the printing style is on point with the rest of the program

lucaderi:
Thank you

PR #737, opened 2019-07-12
Add compiler warnings and fix them

themadprofessor:
Add -Wall and -Werror to compilation and fix the new errors raised by them.


lucaderi:
Thank you

PR #742, opened 2019-07-15
Fix build with GCC 9

### LINKED ISSUE #741: Compilation fail with GCC 9
Compiling nDPI fails with GCC 9.1.1.

Errors are:
```
protocols/hangout.c: In function ‘is_google_flow’:
protocols/hangout.c:65:58: error: taking address of packed member of ‘struct ndpi_iphdr’ may result in an unaligned pointer value [-Werror=address-of-packed-member]
   65 |     if(google_ptree_match(ndpi_struct, (struct in_addr *)&packet->iph->saddr)
      |                                                          ^~~~~~~~~~~~~~~~~~~
protocols/hangout.c:66:61: error: taking address of packed member of ‘struct ndpi_iphdr’ may result in an unaligned pointer value [-Werror=address-of-packed-member]
   66 |        || google_ptree_match(ndpi_struct, (struct in_addr *)&packet->iph->daddr)) {
      |                                                             ^~~~~~~~~~~~~~~~~~~
```

Seems like GCC 9 added some new checks for `-Waddress-of-packed-member`.  My research shows me that lots of other projects got new warnings because of this. Examples: 
https://lkml.org/lkml/2019/1/23/491 
https://github.com/golang/go/issues/29962 
https://github.com/openucx/ucx/issues/3119

clang has such warning, too. There is a lot of discussion here: https://reviews.llvm.org/D20561

Similar discussion for gcc: https://gcc.gnu.org/bugzilla/show_bug.cgi?id=51628

tansly:
Closes #741.

### DIFF
--- a/src/lib/ndpi_main.c
+++ b/src/lib/ndpi_main.c
@@ -3402,7 +3402,7 @@ static u_int8_t ndpi_detection_get_l4_internal(struct ndpi_detection_module_stru
 #ifdef NDPI_DETECTION_SUPPORT_IPV6
   else if(iph->version == 6 && l3_len >= sizeof(struct ndpi_ipv6hdr)) {
     NDPI_LOG_DBG2(ndpi_struct, "ipv6 header\n");
-    iph_v6 = (const struct ndpi_ipv6hdr *) iph;
+    iph_v6 = (const struct ndpi_ipv6hdr *) l3;
     iph = NULL;
   }
 #endif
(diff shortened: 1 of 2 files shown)


tansly:
Packed struct alignments warnings were fixed as I suggested in https://github.com/ntop/nDPI/issues/741#issuecomment-511285501.

There is one warning remaining:
```
ndpi_main.c: In function ‘ndpi_get_custom_category_match’:
ndpi_main.c:2272:3: warning: ‘strncpy’ specified bound 64 equals destination size [-Wstringop-truncation]
 2272 |   strncpy(ipbuf, name_or_ip, sizeof(ipbuf));
      |   ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```
I do not know the constraints of the input etc. so I'd better not touch this part. Someone familiar with this part of the code can find a proper way of fixing this, I believe.

lucaderi:
Thanks for the fix. I have implemented avoiding memory copies. Please check that issues are gone

PR #745, opened 2019-07-16
Add -Werror back in and resolve errors raised due to it

themadprofessor:
Tested with gcc-8 and gcc-9.
It was removed due to #741, which was fixed with f570eccf8143a07475cbd2dd3f4fd9b7261d1298.


lucaderi:
Implemented with https://github.com/ntop/nDPI/commit/9ca87a620a39dbec76e28ea9ebe149c62de5b011

themadprofessor:
9ca87a6 doesn't re-add -Werror to the cflags

PR #747, opened 2019-07-17
Add more compilers to travis

themadprofessor:
Add gcc-{8,9} and clang-{7,8} to travis.
Remove unnecessary install commands.
Change dependency installation to use apt addon
Remove forced clang in Makefile.am when using a version of clang (allows clang-8 to be used).


lucaderi:
Thank you

PR #754, opened 2019-07-19
Extend build system to fully utilise autotools 

### LINKED ISSUE #738: make dist doesn't work
I tried to do `make dist` to create a tarball but the process fails.

I'm getting the following error:
```
/home/user/nDPI> make dist
make  dist-gzip am__post_remove_distdir='@:'
make[1]: Entering directory '/home/user/nDPI'
if test -d "libndpi-2.9.0"; then find "libndpi-2.9.0" -type d ! -perm -200 -exec chmod u+w {} ';' && rm -rf "libndpi-2.9.0" || { sleep 5 && rm -rf "libndpi-2.9.0"; }; else :; fi
test -d "libndpi-2.9.0" || mkdir "libndpi-2.9.0"
 (cd src/lib && make  top_distdir=../../libndpi-2.9.0 distdir=../../libndpi-2.9.0/src/lib \
     am__remove_distdir=: am__skip_length_check=: am__skip_mode_fix=: distdir)
make[2]: Entering directory '/home/user/nDPI/src/lib'
make[2]: *** No rule to make target 'distdir'.  Stop.
make[2]: Leaving directory '/home/user/nDPI/src/lib'
Makefile:585: recipe for target 'distdir' failed
make[1]: *** [distdir] Error 1
make[1]: Leaving directory '/home/user/nDPI'
Makefile:684: recipe for target 'dist' failed
make: *** [dist] Error 2
```
Prior running the `make dist` I did the following steps:
I cloned the repository (`git clone https://github.com/ntop/nDPI.git`) (obviously)
I did `./autogen.sh`
I did `./configure`
I did `make`

All these steps were successful.

Using the following environment:
```
/home/user/nDPI/tests> uname -a
Linux user-vm 4.15.0-54-generic #58-Ubuntu SMP Mon Jun 24 10:55:24 UTC 2019 x86_64 x86_64 x86_64 GNU/Linux
```

themadprofessor:
Replace expanded .in files with .am files
Modify configure.seed to no longer require seeding
Example code is now built with DPDK when configure is given `--with-dpdk` rather than separate makefile
Fixes #738


dsokoloski:
Good luck getting this merged 😄. I tried several times not too long ago and the PRs were rejected because I was told that "autotools was a pain in the ass".  So I have to maintain my own fork with full autotools integration.

nickbroon:
I assume you mean https://github.com/ntop/nDPI/pull/594#issuecomment-415967395

The current nDPI code already makes use of automake, just in a very adhoc way.
https://github.com/ntop/nDPI/blob/dev/src/lib/Makefile.in#L4
Despite the comment at the above link autogen.sh actually invokes automake.

I too have spend years developing for embedded platforms and have never had issue with automake. Once the dist tarball is generated by a maintainer autoconf/automake are no needed to actually compile the code. So a tarball would build on an embedded platform using just configure/make.

dsokoloski:
@nickbroon Yes that was my last rejected PR.

> The current nDPI code already makes use of automake, just in a very ~~adhoc~~ *broken* way.

Yes I'm aware of the half-implementation... the project used to contain a Makefile.am under ./src not too long ago before it was removed for "reasons" outlined in the comments of #594.

Fingers crossed this attempt to fully reintegrate autotools is successful :)

lucaderi:
I don't like autotools, in particular for embedded systems. I don't see a value for this work, and actually the real fix is to cleanup the current code, as @nickbroon pointed out.
@themadprofessor Are you willing to do this?

themadprofessor:
What's wrong with autotools? Not sure why you would be compiling on an embedded system and not doing cross-compilation.
If I was to replace it, any preference was to what to replace it with? Cmake, Meson?

lucaderi:
The problem with auto tools is that I don't see what problem they solve with nDPI. If we remove json-c dependency we can probably avoid configure at all. They are an un-necessary legacy complexity.

themadprofessor:
Currently, the nDPI library uses configure to scan for pthread and hyperscan. The example code uses configure to scan for json-c, pcap, pthread and libm. Not using a makefile build tool (like autotools or Cmake) would reduce configuration options (using a custom build of hyperscan rather than system default for example), while adding developmental complexity for creating the makefiles and require manual construction of ndpi_config.h. While replacing autotools with raw makefiles is possible, personally I would see it as a misstep. I'm fine with replacing autotools with another makefile build tool.

nickbroon:
The lack of a full configure build system also means distribution need to patch if they want to things like change install paths, cflags or specify library locations.
 https://salsa.debian.org/debian/ndpi/blob/master/debian/patches/fix-makefile-prefix

Using a configure/build system like autotools, cmake, meson seems like a reasonable thing to do.
My past experience of embedded systems is that almost all will have the tools support for these build systems so that even a local build can be done instead of a cross build.

themadprofessor:
@lucaderi Any progress on this?

themadprofessor:
@lucaderi Nudge

lucaderi:
I told you, I don't like autotools. Sorry. We can improve configure of course but let's keep things simple

themadprofessor:
That's fair enough. As I asked previously, if I was to replace it, which makefile build tool should I use instead? Cmake, Meson, ...?

lucaderi:
I have nothing against you using cmake, this in addition to the standard make

lucaderi:
I close this pull for the time being

PR #758, opened 2019-07-23
Add support for WireGuard VPN

tansly:
[WireGuard](https://www.wireguard.com/) is an emerging VPN protocol.
This pull requests adds detection support for WireGuard. It reliably detects the protocol using a simple byte pattern analysis on a single packet, and does not generate false positives in existing tests.


lucaderi:
To be honest this dissector looks a bit weak, Could you harden it adding extra checks and share a pcap file we can use for testing it?

tansly:
Of course. I'll be looking into it.

tansly:
I added some extra checks that exploit the fixed size handshakes and sender/receiver index fields in the packets.

tansly:
> [...] and share a pcap file we can use for testing it?

I also added a pcap to the tests @lucaderi.

lucaderi:
Thank you

PR #760, opened 2019-07-23
Ndpi splt bd

rkerur:
Hi Luca,

Please take a look at the changes which includes 
(1) Adding port based classification for perf
(2) Adding Amazon static classification
(3) Xbox and PS4 static port classification
(4) SPLT and BD

Thanks,
Ravi


lucaderi:
The code looks good. let's merge it and start planning for using the classification.

PR #767, opened 2019-09-11
Fix cross-build from source, because of direct pkg-config check calls

LocutusOfBorg:
ndpi fails to cross build from source, because configure.seed hard codes
the build architecture pkg-config. It is best to use PKG_CHECK_MODULES
instead. Please consider applying the attached patch.

See: https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=939286

### DIFF
--- a/configure.seed
+++ b/configure.seed
@@ -70,29 +70,25 @@ if test $SYSTEM = "Darwin"; then
 fi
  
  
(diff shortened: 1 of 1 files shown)


lucaderi:
Thank you

PR #775, opened 2019-09-18
Fixed Skype protocol detection

ice0:


lucaderi:
I have reworked completely the STUN dissector so this fix is no longer necessary. Thank you

PR #778, opened 2019-09-20
Added support for dissection of DNP3 and IEC104 protocols

choyos:
I have to add support of both protocols for a thesis project.
Finally I am not be able to work with nDPI, but I want to contribute if you consider that could be added.
Regards


lucaderi:
Can you please update this contribution with the current source tree?

PR #784, opened 2019-09-26
Check for both src and dest combination for quick classification.

rkerur:

Check for both src and dest for quick classification. Either one of the match should result in Zoom.

### DIFF
--- a/src/lib/protocols/zoom.c
+++ b/src/lib/protocols/zoom.c
@@ -27,7 +27,18 @@
 
 #include "ndpi_api.h"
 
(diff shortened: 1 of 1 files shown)


lucaderi:
Why do we need this plugin at all?

rkerur:
Plugin file is redundant right now because it checks IP and ports only. It was added as a place holder to add signatures when IP and ports are not enough. Same argument as ps_vue.c file.

lucaderi:
This patch is no longer necessary as the zoom dissector has been removed

PR #788, opened 2019-09-29
Add Hulu specific IP address range.

rkerur:

Should Hulu be streaming or video category? Currently it is set to streaming.


lucaderi:
Since https://github.com/ntop/nDPI/commit/f2a5bbef173ee7f5447871f26024b9639735c096, unless Hulu is a protocol that nDPI needs to detect, it is preferable to handle it as streaming category. What do you think?

rkerur:
Make sense to me. Let me work on it and send the updated diffs.

lucaderi:
I had to merge this by hand as it was unmergeable after the recent changes

PR #792, opened 2019-10-03
Move Hulu and Disney+ to category match.

rkerur:

Hi Luca,

Patch is for RFC only. Not yet fully tested. Is this what you had in mind for streaming category?

Thanks.


lucaderi:
Exactly. If your goal is to indetify specific protocol families, better to use categories than protocols as people cannot configure hundred of protocols but can handle categories.

PR #799, opened 2019-10-19
Fix memory leak in ndpiReader.

rkerur:
Changes in library for strict function prototype.

Memory leak was due to earlier changes to use entropy fields which gets reset every 10 packets and memory not freed for previous allocation. Stick to original functionality of payload analysis of first <n> packets of a flow.


lucaderi:
Thank you

PR #808, opened 2019-10-29
Change type

sysoleg:
Use double type instead of float when printing time for CSV file.
Float type is too small for 64 bit int time so cast corrupts value.


lucaderi:
Thank you

PR #810, opened 2019-10-31
Adds fuzz target

catenacyber:
And configure option enable-fuzztargets

Later, a second fuzz target should be added reading a whole pcap and handling flows


lucaderi:
Thank you

PR #811, opened 2019-11-15
Python bindings fix and improve

aouinizied:
This pull request implement the following features regarding nDPI python bindings:
* Add missing structures fields.
* Fix alignment and padding issues.
* Fix detection giveup calls.
* Ensures nDPI >= 3.0 compatibility.
* Switch to Python class camel case naming.
* Add flow_printer example based on nfstream package.
* Polish and other minor fixes.


lucaderi:
Thank you

PR #817, opened 2019-12-05
Integration of the libinjection library

MrTiz:
Integration of the libinjection library to detect SQL injections and XSS type attacks in HTTP requests. The check is done on the URL. The library used is libinjection, author Nick Galbreath @client9 [https://github.com/client9/libinjection](https://github.com/client9/libinjection)


lucaderi:
Thank you

Issue #718, opened 2019-06-14
False positive Skype detection


Gelob:
Running ntopng and ndpi-2.9.0-1596 i saw a surge in my network traffic of 4Gbps and 2-3Mpps. When looking at ntopng I see it shows this traffic spike as Skype traffic. Really this is just Azure blob storage endpoint traffic, and its not skype. I don't have a pcap because I do flow based capturing but it was all dst port 443 to a single IP in the 13.96.0.0/13, 13.104.0.0/14, 13.64.0.0/11 space. When looking at passive DNS that IP has a hostname like blob.*.store.core.windows.net where * is a wildcard and would be a unique ID from Azure.

haroflow:
I've found this page from MS: https://www.microsoft.com/en-us/download/details.aspx?id=41653
It says "deprecated" but has links for weekly (apparently) updated Azure endpoints:
- Public: https://www.microsoft.com/en-us/download/details.aspx?id=56519
- US Gov: http://www.microsoft.com/en-us/download/details.aspx?id=57063
- Germany: http://www.microsoft.com/en-us/download/details.aspx?id=57064
- China: http://www.microsoft.com/en-us/download/details.aspx?id=57062

Maybe we could add these as NDPI_PROTOCOL_MICROSOFT or NDPI_PROTOCOL_AZURE?

Gelob:
Closing this as #823 was merged

PR #823, opened 2019-12-16
Changed 13.64.0.0/11 range from Skype to Microsoft

haroflow:
Blocking this range as Skype also blocks some azure-hosted websites like redesullog.com.br .
Skype blocking still works after these changes.

These ranges are not listed as Skype anymore, I think they might be for azure. Relates to #718


haroflow:
We were having some problems blocking Facebook, it was blocking web.whatsapp.com as well...

I believe the IP/Range list on https://developers.facebook.com/docs/whatsapp/network-debugging is already outdated.

We've scanned the 31.13.64.0/18 and 157.240.0.0/16 ranges, doing reverse lookups for all IPs and checking if the domain matches *whatsapp*, and we've found many IPs like that. I added those as WhatsApp. There may be other ranges with the same problem.

I tried adding a url match for *whatsapp*.fbcdn.net. but still showed up as Facebook... I don't know what I did wrong...

I'm open to discuss how we could solve this problem, it also happens with Microsoft/Azure/Office/Skype. I know it's hard to get it right, but maybe these reverse lookups can help...

Thanks!

lucaderi:
Thank you for this work. If I encounter problems I will look at them.

PR #824, opened 2019-12-18
Fix read buffer overflow in stun

catenacyber:
Found by oss-fuzz :
https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=19332

The line
`u_int16_t len = ntohs(*((u_int16_t*)&payload[offset+2]));`
accesses 2 bytes starting from `offset+2` so we need at least 4 bytes


lucaderi:
Thank you

PR #830, opened 2020-01-14
Adds another fuzz target

catenacyber:
This new target takes pcap files as inputs, by compiling part of ndpiReader as a library


lucaderi:
Thank you

PR #831, opened 2020-01-22
Generates automatically corpus for fuzz target

catenacyber:
Corpus is good for giving fuzzer a starting point


lucaderi:
Thank you

PR #832, opened 2020-01-22
Adds fuzz targets build and run to travis

catenacyber:
Build comes with the different sanitizers

Inspired by https://google.github.io/oss-fuzz/advanced-topics/ideal-integration/#regression-testing


lucaderi:
Thank you

PR #833, opened 2020-01-24
Fix sanitizers build on Travis

catenacyber:
Completes #832


catenacyber:
The PR seems now complete (you can squash the commits)
Now, there is only to fix the bugs...

lucaderi:
Great!

PR #836, opened 2020-02-01
nDPI now detects RCE injections in HTTP requests

MrTiz:
Initially I tried to implement the regular expressions via Intel Hyperscan; however the detections was too slow and generated many false positives and many false negatives. For this reason, it was decided to use PCRE instead of Intel Hyperscan, and this meant that the performance improved significantly. Furthermore, thanks to the libpcre library, it was possible to remove all false positives and all false negatives.


lucaderi:
Thank you ( I will fix the compilation with libfuzz)

PR #838, opened 2020-02-03
Adds missing checks

catenacyber:
Found by fuzzing

Should fix up to https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=20270 (not included)


lucaderi:
Thank you

PR #839, opened 2020-02-06
Adds oss-fuzz badge to REAMDE

catenacyber:
cf https://google.github.io/oss-fuzz/getting-started/new-project-guide/#status-badge


lucaderi:
Thank you

PR #840, opened 2020-02-10
travis: bump to bionic, add arm64 & s390x.

xnox:


xnox:
Observe that the travis pipeline now fails on the following new matrix builds:

- Ubuntu 18.04 LTS (bionic) clang amd64
- Ubuntu 18.04 LTS (bionic) clang & gcc on ARM64
- Ubuntu 18.04 LTS (bionic) clang & gcc on S390X

Please merge these tests, and fix the testsuites / software to support these additional architectures.

This is also reported as a regression in debian https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=950404

xnox:
@emanuele-f can you please look into merging this travis config that enables testing on more targets, and exposes regressions in architecture support in v3.0 release when compared to v2.6? Also In Debian many other architectures are affected ie. sparc alpha mips etc.

emanuele-f:
Will fix the floating point problem in the upcoming commit

PR #846, opened 2020-02-18
Fix various buffer over reads

catenacyber:
Found by oss-fuzz
Should fix 
https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=20540
https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=20747
https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=20712

The leaks remain :
https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=20727
https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=20731


lucaderi:
Thank you

PR #849, opened 2020-02-26
ndpi_flow2json should check http.url before serializer

annp1987:
This patch will fix issue: https://github.com/ntop/PF_RING/issues/557


lucaderi:
Thank you

PR #850, opened 2020-02-26
Fuzz4fix

catenacyber:
Should fix 
https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=20766
https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=20727
https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=20830
https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=20777

A better fit (against root cause) would be to read all IPv6 option headers in ndpi reader as ndpi library does

### DIFF
--- a/example/reader_util.c
+++ b/example/reader_util.c
@@ -1709,6 +1709,8 @@ ether_type_check:
   }
 
   if(workflow->prefs.decode_tunnels && (proto == IPPROTO_UDP)) {
+    if (header->caplen < ip_offset + ip_len + sizeof(struct ndpi_udphdr))
+      return(nproto); /* Too short for UDP header*/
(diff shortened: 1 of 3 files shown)


lucaderi:
Thank you

catenacyber:
Luca, why did you revert https://github.com/ntop/nDPI/pull/850/commits/5fc9d41eb0b337958fb7c7446ca690f8ba361285 ? with https://github.com/ntop/nDPI/commit/34ad06fef53626bc718b3e349a365f8fd0ef3646

It seems that https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=20727 is therefore not fixed

PR #856, opened 2020-03-12
Fuzzfix5

catenacyber:
These should fix all the current bugs found by oss-fuzz
except for the two with `#if LUCA` from the previous PR
and for the memory leak in https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=20799


lucaderi:
Thank you

PR #859, opened 2020-03-19
Checks enough data for UDP header

catenacyber:
@lucaderi did you say this breaks the tests ?
How so ?
When I run `./do.sh` in tests directory, I get only OKs


catenacyber:
Plus latest commit fixes new bugs from oss-fuzz

lucaderi:
Thanks

PR #867, opened 2020-04-10
Memory leaks

IvanNardi:
Other then some minor stuff, in this PR there are two fixes for two memory leaks:
* the first one, in ahocorasick code, is completely harmless, since the leak is tiny and fixed in size, i.e. it doesn't depend on the input traffic. But it might be worthy of fixing nonetheless, to have a clean asan output
* the other one is more severe, since some TLS idle flows were not properly freed


lucaderi:
Thank you

PR #869, opened 2020-04-13
Fix incorrect "<iq from=\"' parsing

disconnect3d:
This PR fixes a wrong size arg passed to memcmp when checking for "<iq from=\"" and "<iq from \'" strings.


lucaderi:
Thank you

PR #872, opened 2020-04-15
Update python example.

aouinizied:
In the following PR, we provide minor fix:
* Fix the flow_printer example with new API version.
* Enable flow statistical features (95 flow features in total).


lucaderi:
Thank you

PR #873, opened 2020-04-15
Fix for missing line breaks in CSV file

a-czyrny:
If "-v" is used as an argument, the line breaks are missing in the csv file ("-C"), because the argument "-J" is assumed. Moving the condition for the CSV file handler out of the scope of the "enable_joy_stats" condition removes this dependency and the CSV file is formatted correctly.


lucaderi:
Thank you

PR #874, opened 2020-04-15
Fuzz6fix

catenacyber:
Should fix bugs reported by oss-fuzz and Github Security Lab (so far)

One functional change is to export `ndpi_handle_ipv6_extension_headers` and use it in the example

### DIFF
--- a/src/lib/protocols/netbios.c
+++ b/src/lib/protocols/netbios.c
@@ -80,7 +80,8 @@ static void ndpi_int_netbios_add_connection(struct ndpi_detection_module_struct 
   char name[64];
   u_int off = flow->packet.payload[12] == 0x20 ? 12 : 14;
   
-  if(ndpi_netbios_name_interpret((char*)&flow->packet.payload[off], flow->packet.payload_packet_len - off, name, sizeof(name)) > 0)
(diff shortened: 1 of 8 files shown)


lucaderi:
Thank you

src/lib/protocols/netbios.c
```diff
@@ -80,7 +80,8 @@ static void ndpi_int_netbios_add_connection(struct ndpi_detection_module_struct
   char name[64];
   u_int off = flow->packet.payload[12] == 0x20 ? 12 : 14;
   
-  if(ndpi_netbios_name_interpret((char*)&flow->packet.payload[off], flow->packet.payload_packet_len - off, name, sizeof(name)) > 0)
+  if(off > flow->packet.payload_packet_len &&
```
lucaderi, line 83: This is a bug it should be (off < flow->.....)

catenacyber: Indeed, just force pushed the fix

Would you like me to open a new PR on this ?
Or do you fix it yourself ?

catenacyber: Thanks for already fixing this

PR #876, opened 2020-04-16
additional csv semicolon fix

a-czyrny:
Created function correct_csv_data_field to prevent duplicated code. Additionally used for _flow->ndpi_flow->protos.stun_ssl.ssl.alpn_ and _flow->ndpi_flow->protos.stun_ssl.ssl.tls_supported_versions_ to guarantee a valid csv output (commas replaced by semicolon) .


lucaderi:
Thank you

PR #877, opened 2020-04-17
Adds bound check before calling ndpi_handle_ipv6_extension_headers

catenacyber:
Should fix https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=21719 (introduced by latest PR fixing the behavior with multiple IPv6 extensions)

### DIFF
--- a/example/reader_util.c
+++ b/example/reader_util.c
@@ -1700,6 +1700,9 @@ ether_type_check:
     iph6 = (struct ndpi_ipv6hdr *)&packet[ip_offset];
     proto = iph6->ip6_hdr.ip6_un1_nxt;
     ip_len = ntohs(iph6->ip6_hdr.ip6_un1_plen);
+    if (header->caplen < ip_offset + sizeof(struct ndpi_ipv6hdr + ntohs(iph_v6->ip6_hdr.ip6_un1_plen))
+      return(nproto); /* Too short for IPv6 payload*/
+
 
     const u_int8_t *l4ptr = (((const u_int8_t *) iph6) + sizeof(struct ndpi_ipv6hdr));
     if(ndpi_handle_ipv6_extension_headers(NULL, &l4ptr, &ip_len, &proto) != 0) {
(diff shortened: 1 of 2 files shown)


lucaderi:
Thank you

example/reader_util.c
```diff
@@ -1700,11 +1694,15 @@ struct ndpi_proto ndpi_workflow_process_packet(struct ndpi_workflow * workflow,
...
     ip_len = ntohs(iph6->ip6_hdr.ip6_un1_plen);
+    if (header->caplen < ip_offset + sizeof(struct ndpi_ipv6hdr + ntohs(iph_v6->ip6_hdr.ip6_un1_plen))
+      return(nproto); /* Too short for IPv6 payload*/
+
 
     const u_int8_t *l4ptr = (((const u_int8_t *) iph6) + sizeof(struct ndpi_ipv6hdr));
     if(ndpi_handle_ipv6_extension_headers(NULL, &l4ptr, &ip_len, &proto) != 0) {
       return(nproto);
     }
+    iph6->ip6_hdr.ip6_un1_nxt = proto;
```
lucaderi, line 1705: This is wrong as you are modifying a read-only packet

example/reader_util.c
```diff
@@ -1700,11 +1694,15 @@ struct ndpi_proto ndpi_workflow_process_packet(struct ndpi_workflow * workflow,
     iph6 = (struct ndpi_ipv6hdr *)&packet[ip_offset];
     proto = iph6->ip6_hdr.ip6_un1_nxt;
     ip_len = ntohs(iph6->ip6_hdr.ip6_un1_plen);
+    if (header->caplen < ip_offset + sizeof(struct ndpi_ipv6hdr + ntohs(iph_v6->ip6_hdr.ip6_un1_plen))
```
lucaderi, line 1697: I had to fix it as this line contains a syntax error. Please check your code before pulling it

catenacyber: Sorry, I thought I had checked but I must have checked before adding the syntax error

example/reader_util.c
catenacyber: So, we need to pass `proto` as a an argument to `get_ndpi_flow_info6`.
Do you see another way ?

PR #879, opened 2020-04-20
Fix some compilation warnings

IvanNardi:
This PR fixes all warnings found when compiling with "-W -Wno-unused-parameter -Wno-sign-compare" flags.
Fixing the warnings emitted by "-Wunused-parameter -Wsign-compare" options (implicitly enabled by "-W" flag) will require more work and I am unsure if it is worth the effort


lucaderi:
Thank you!

PR #880, opened 2020-04-21
Gets right protocol after IPv6 header

catenacyber:
Following https://github.com/ntop/nDPI/pull/877#discussion_r411362220 to fix https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=20727


lucaderi:
Thank you

PR #881, opened 2020-04-22
Bug protocol was guessed

### LINKED ISSUE #852: protocol_was_guessed Is never used
protocol_was_guessed Is never used at ndpi_detection_giveup() - ndpi_main.c
 
it should be set to 1 when the guess happens within ndpi_detection_giveup

leonn:
Fix #852 
set protocol_was_guessed to true when guessed in ndpi_detection_giveup
Improves code indentation for more readability,


lucaderi:
Thank you!

PR #884, opened 2020-04-22
Added missing ndpiReader dependency for the install target

utoni:
The install target should be aware that the ndpiReader has to be built.


lucaderi:
Thank you

PR #897, opened 2020-05-10
Adds bound check in oracle protocol

catenacyber:
Found by oss-fuzz
https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=21780


lucaderi:
Thank you

PR #900, opened 2020-05-12
"Add Dridex malware detection protocol and detection of dangerous HTTP transfers"

Loures:


lucaderi:
Unfortunately dridex is a malware that cannot be tracked with IP addresses and thus this contribution cannot be accepted as is

PR #907, opened 2020-05-20
fix run ARM architecture will cause sigbus error

yskcg:
error info:
Using nDPI (3.3.0-2408-3d9285f1) [1 thread(s)]
Capturing live traffic from device br0...
Running thread 0...
Bus error

use the syslog  see:
Unhandled fault: alignment fault (0x92000021) at 0x00000000f67004aa

gdb info:
(gdb) r -i br0
Starting program: /tmp/ndpiReader -i br0
    warning: Unable to find libthread_db matching inferior's thread library, thread debugging will 
     not be available.

    -----------------------------------------------------------
* NOTE: This is demo app to show *some* nDPI features.
* In this demo we have implemented only some basic features
* just to show you what you can do with the library. Feel
* free to extend it and send us the patches for inclusion
------------------------------------------------------------

Using nDPI (3.3.0-2408-3d9285f1) [1 thread(s)]
Capturing live traffic from device br0...
[New LWP 9883]
Running thread 0...

Thread 2 "ndpiReader" received signal SIGBUS, Bus error.
[Switching to LWP 9883]
ndpi_search_someip (ndpi_struct=0xf7218008, flow=0xf6f021b0) at protocols/someip.c:115
115	protocols/someip.c: No such file or directory.
this will be a undefined handle in some ARM device ;

so ,use the safe way to get the value;

tesk ok!

### DIFF
--- a/src/lib/protocols/someip.c
+++ b/src/lib/protocols/someip.c
@@ -87,7 +87,7 @@ static void ndpi_int_someip_add_connection (struct ndpi_detection_module_struct 
   NDPI_LOG_INFO(ndpi_struct, "found SOME/IP\n");
 }
 
-static u_int32_t someip_data_cover_32(u_int8_t *data)
+static u_int32_t someip_data_cover_32(const u_int8_t *data)
 {
   u_int32_t value;
 
--- a/src/lib/protocols/someip.c
+++ b/src/lib/protocols/someip.c
@@ -87,6 +87,14 @@ static void ndpi_int_someip_add_connection (struct ndpi_detection_module_struct 
   NDPI_LOG_INFO(ndpi_struct, "found SOME/IP\n");
 }
 
+static u_int32_t someip_data_cover_32(u_int8_t *data)
+{
+  u_int32_t value;
+
+  memcpy(&value,data,sizeof(u_int32_t));
+
+  return value;
+}
 /**
  * Dissector function that searches SOME/IP headers
  */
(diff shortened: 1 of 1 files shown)


src/lib/protocols/someip.c
```diff
@@ -111,8 +119,8 @@ void ndpi_search_someip (struct ndpi_detection_module_struct *ndpi_struct,
   }
  
   //we extract the Message ID and Request ID and check for special cases later
-  u_int32_t message_id = ntohl(*((u_int32_t *)&packet->payload[0]));
-  u_int32_t request_id = ntohl(*((u_int32_t *)&packet->payload[8]));
+  u_int32_t message_id = ntohl(someip_data_cover_32(&packet->payload[0]));
```
lucaderi, line 122: I have the feeling that these issues are not limited to some_ip as probably these are related to memory alignment. Can you please tell me
- how I can reproduce the issue? On a Raspberry I have never seen this error
- if this is (as I think) a general ARM problem, I think we need to check all protocols not all some_ip. Did you run the nDPI/tests/do.sh command? Is everything working?
- do you have a some_ip pcap I can use for testing?

lucaderi (changes requested):

src/lib/protocols/someip.c
yskcg: #more details:
*   I have just portable the nDpi run my router RT-AC86U version:3.0.0.4.384_81351_koolshare
 
* when I just run the target ,like this :`./ndpiReader -i br0` ,and at the same time,it's broken,notify SIGBUS error occour,so ,there is just no have some_ip.pcap packet
* when the problem occour,i run the cmd `sysinfo `,loged as `Unhandled fault: alignment fault (0x92000021) at 0x00000000f67004aa`

* i don't run do.sh script in my router,because ,it's need envirement;

PR #909, opened 2020-05-27
Added check for suspicious HTTP headers

tmacchioni:


src/lib/protocols/http.c
```diff
@@ -600,21 +600,42 @@ static void http_bitmask_exclude_other(struct ndpi_flow_struct *flow)
...
+  struct ndpi_packet_struct *packet = &flow->packet;
   
   for(i=0; (i<packet->parsed_lines) && (packet->line[i].ptr != NULL); i++) {
-    printf("-->> [len: %u] [%s]\n", packet->line[i].len, packet->line[i].ptr);
-
-    if(match_found)
-      NDPI_SET_BIT(flow->risk, NDPI_HTTP_SUSPICIOUS_HEADER);
+    if((header_limit = memchr(packet->line[i].ptr, ':', (packet->line[i].len)))){
+      header_len = header_limit - packet->line[i].ptr;
+      for(s=0; suspicious_http_header_keys[s] != NULL; s++){
```
lucaderi: This code is suboptimal as it adds quite some overhead. Better to reduce the load with something like
switch(packet->line[i].ptr[0]) {
case "A'
/* check only strings that start with A */
break;
case "M'
/* check only strings that start with < */
break;
...
}

Something like this. So we can minimise the overhead without having to use complex datastructures

lucaderi (changes requested):
Please fix the reported issue prior to merge the code

PR #911, opened 2020-05-29
Added check for suspicious HTTP headers

tmacchioni:


src/lib/protocols/http.c
```diff
@@ -597,26 +597,83 @@ static void http_bitmask_exclude_other(struct ndpi_flow_struct *flow)
...
-    printf("-->> [len: %u] [%s]\n", packet->line[i].len, packet->line[i].ptr);
 
-    if(match_found)
-      NDPI_SET_BIT(flow->risk, NDPI_HTTP_SUSPICIOUS_HEADER);
-  }
-#endif
+  struct ndpi_packet_struct *packet = &flow->packet;
+
+  for(i=0; (i<packet->parsed_lines) && (packet->line[i].ptr != NULL); i++) {
+    switch(packet->line[i].ptr[0]){
```
lucaderi, line 635: An extra check is necessary here (length)

PR #913, opened 2020-06-03
Fix segment fault dev

yskcg:
# more details:
*   I have just portable the nDpi run my router RT-AC86U version:3.0.0.4.384_81351_koolshare
 
* when I just run the target ,like this :`./ndpiReader -i br0` ,and at the same time,it's broken,notify SIGBUS error occour,so ,there is just no have some_ip.pcap packet
* when the problem occour,i run the cmd `sysinfo `,loged as `Unhandled fault: alignment fault (0x92000021) at 0x00000000f67004aa`

* i don't run do.sh script in my router,because ,it's need environment;

# background
* i portable the nDpiReader to run at my router RT-AC86U version:3.0.0.4.384_81351_koolshare;
* config detect protocol like this:
   * `NDPI_BITMASK_ADD(all,NDPI_PROTOCOL_DNS);
   NDPI_BITMASK_ADD(all,NDPI_PROTOCOL_HTTP);
   NDPI_BITMASK_ADD(all,NDPI_PROTOCOL_DHCP);
   NDPI_BITMASK_ADD(all,NDPI_PROTOCOL_XBOX);
   NDPI_BITMASK_ADD(all,NDPI_PROTOCOL_NINTENDO);
   NDPI_BITMASK_ADD(all,NDPI_PROTOCOL_PLAYSTATION);` in `setupDetection` function;
* when run on router,it may product coredump,detail :
	`segfault at 0 ip 00007f1af00d5641 sp 00007f1aead2f878 error 4 in libc-2.17.so`*

so ,trace the problem,i guess that `protos.stun_ssl.ssl.server_names` may be NULL,that can cause the problem;

trace the code, the `server_names_len` may not equal zero ,but `protos.stun_ssl.ssl.server_names ` use the function `ndpi_strdup` to copy ,inner `ndpi_strup`,the `strlen` calculate the `NULL` pointer ,it cause segment fault!!!

# test code 

`#include <stdio.h>`

`#include <string.h>`

`#include <stdint.h>`

`char *ndpi_strdup(const char *s) {
  int len = strlen(s);

  char *m = malloc(len + 1);

  if(m) {
    memcpy(m, s, len);
    m[len] = '\0';
  }

  return(m);
}

int main()
{
	uint8_t *test_NULL = NULL;
	uint8_t *test = ndpi_strdup(test_NULL);

}

my test code  also take the same segment fault !!!


lucaderi:
Thank you

PR #920, opened 2020-06-17
Fixed stack overflow caused by missing length check

utoni:


utoni:
(also increased rdnSeqBuf to 2048 to get more RDN sequences)

lucaderi:
Thank you

PR #922, opened 2020-06-21
Fixed use after free caused by dangling pointer

utoni:
* This fix also improved RCE Injection detection


lucaderi:
Thank you

PR #923, opened 2020-06-22
Fixed RCE Injection detection test results (was improved by #922)

utoni:
I think there was an issue during merging/rebasing for the fix #922, which also improved RCE Injection detection.
It was basically reverted by 1a62f4c799186244da0a2923d1db79f8e4e3c10b, so `./tests/do.sh` is failing for the moment. This MR will fix that. =)


utoni:
Improvement was caused by PCRE not by my fix ... lol

PR #924, opened 2020-06-22
Allow -fsanitize for LLVM as well as for GCC.

utoni:
* renamed --with-llvm-sanitizer to --with-sanitizer
     * disable all optimisations (-O0) if --with-sanitizer set,
       no functions/paramaters/variables will be optimised, improves debugging
     * enable undefined behaviour sanitizer (ubsan)
     * enable leak sanitizer (lsan)
     * check if -fsanitize=fuzzer is available and --enable-fuzztargets set,
       fail if not (only supported by clang)
     * include level 3 debugging information (-g3), improves macro debugging
     * disabled ubsan misaligned pointer access and lshift overflow

### DIFF
--- a/configure.seed
+++ b/configure.seed
@@ -4,13 +4,11 @@ AC_CONFIG_MACRO_DIR([m4])
 
 AM_INIT_AUTOMAKE([foreign subdir-objects])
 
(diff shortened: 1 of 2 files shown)


lucaderi:
Thank you. This patch is very helpful!

PR #928, opened 2020-06-24
PROPOSAL: ndpiSimpleIntegration: added another integration example

utoni:
The ndpiReader is good for debugging purposes, but is not easy to understand for newcomers.
So as I want to use this library in another project anyway, I've decided to provide another integration example which shows the very basic features of it.


lucaderi:
I am ready to merge this code and I agree with you. I am just wondering if we need another tool or we we should instead simplify the ndpiReader. For instance I would remove all the thread stuff for instance. What do you think?

utoni:
Refactoring ndpiReader will definitely take some time. I would not remove all the threading stuff completely. Instead using an appropriate per thread packet distribution seems fine to me. However, I'm more concerned about what are the actual requirements for ndpiReader. IMHO it is good for statistics/ testing/ debugging, bad for showing a developer how the API works.

lucaderi:
Ok o'well integrate this for now, thanks

utoni:
Ah, just saw that failing CI, but you fixed it already.

PR #931, opened 2020-06-25
fixed autoconf version check

utoni:


utoni:
Sad that nDPI needs to support ancient systems.
If you tell me the CentOS version used, I might be able to find a workaround.

lucaderi:
The issue happens with CentOS 6. The problem of the changes in your latest PR is that it has broken the travis script used to check memory with LLVM. Do you have a clue how to fix the bug?

PR #932, opened 2020-06-26
Log

IvanNardi:
It is quite surprising to me that a basic (and useful!) feature such as --enable-debug-messages doesn't seem to work out-of-the-box.
I don't know if I understood all the ramifications of this issue, but this PR should be an improvement to the current situation, at least

### DIFF
--- a/example/ndpiReader.c
+++ b/example/ndpiReader.c
@@ -384,7 +384,9 @@ static void help(u_int long_help) {
 	 "  -V <1-4>                  | nDPI logging level\n"
 	 "                            | 1 - trace, 2 - debug, 3 - full debug\n"
 	 "                            | >3 - full debug + log enabled for all protocols (i.e. '-u all')\n"
-	 "  -u proto|num[,...]        | Enable logging only for such protocol(s)\n"
+	 "  -u all|proto|num[,...]    | Enable logging only for such protocol(s)\n"
+	 "                            | If this flag is present multiple times (directly, or via '-V'),\n"
+	 "                            | only the last instance will be considered\n"
 	 "  -T <num>                  | Max number of TCP processed packets before giving up [default: %u]\n"
 	 "  -U <num>                  | Max number of UDP processed packets before giving up [default: %u]\n"
 	 ,
@@ -736,11 +738,13 @@ static void parseOptions(int argc, char **argv) {
       if(nDPI_LogLevel < 0) nDPI_LogLevel = 0;
       if(nDPI_LogLevel > 3) {
 	nDPI_LogLevel = 3;
+	free(_debug_protocols);
 	_debug_protocols = strdup("all");
       }
       break;
 
     case 'u':
+      free(_debug_protocols);
       _debug_protocols = strdup(optarg);
       break;
 
@@ -819,6 +823,7 @@ static void parseOptions(int argc, char **argv) {
 
     case '8':
       nDPI_LogLevel = NDPI_LOG_DEBUG_EXTRA;
+      free(_debug_protocols);
       _debug_protocols = strdup("all");
       break;
 
@@ -3525,6 +3530,7 @@ int orginal_main(int argc, char **argv) {
     if(extcap_fifo_h) pcap_close(extcap_fifo_h);
     if(ndpi_info_mod) ndpi_exit_detection_module(ndpi_info_mod);
     if(csv_fp)        fclose(csv_fp);
+    free(_debug_protocols);
 
     return 0;
   }
--- a/example/ndpiReader.c
+++ b/example/ndpiReader.c
@@ -383,7 +383,8 @@ static void help(u_int long_help) {
 	 "                            | 3 = port stats\n"
 	 "  -V <1-4>                  | nDPI logging level\n"
 	 "                            | 1 - trace, 2 - debug, 3 - full debug\n"
-	 "                            | >3 - full debug + dbg_proto = all\n"
+	 "                            | >3 - full debug + log enabled for all protocols (i.e. '-u all')\n"
+	 "  -u proto|num[,...]        | Enable logging only for such protocol(s)\n"
 	 "  -T <num>                  | Max number of TCP processed packets before giving up [default: %u]\n"
 	 "  -U <num>                  | Max number of UDP processed packets before giving up [default: %u]\n"
 	 ,
@@ -402,7 +403,6 @@ static void help(u_int long_help) {
 	 "  --extcap-capture-filter\n"
 	 "  --fifo <path to file or pipe>\n"
 	 "  --debug\n"
-	 "  --dbg-proto proto|num[,...]\n"
     );
 #endif
 
@@ -434,7 +434,6 @@ static struct option longopts[] = {
   { "extcap-capture-filter", required_argument, NULL, '6'},
   { "fifo", required_argument, NULL, '7'},
   { "debug", no_argument, NULL, '8'},
-  { "dbg-proto", required_argument, NULL, 257},
   { "ndpi-proto-filter", required_argument, NULL, '9'},
 
   /* ndpiReader options */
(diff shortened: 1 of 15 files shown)


src/lib/ndpi_main.c
```diff
@@ -1954,7 +1954,8 @@ struct ndpi_detection_module_struct *ndpi_init_detection_module(ndpi_init_prefs
 
   if(ndpi_str == NULL) {
 #ifdef NDPI_ENABLE_DEBUG_MESSAGES
-    NDPI_LOG_ERR(ndpi_str, "ndpi_init_detection_module initial malloc failed for ndpi_str\n");
+    /* We can't use NDPI_LOG* functions yet! */
+    fprintf(stderr, "ndpi_init_detection_module initial malloc failed for ndpi_str\n");
```
utoni: A shared object should never assume that file descriptor 2 is a stderr fd, it could be anything, a fd to /dev/null, a socket, pipe or whatever fd.

nDPI already has a quite nice function called `ndpi_debug_printf`, which callback can be altered with `set_ndpi_debug_function`

src/lib/ndpi_main.c
```diff
@@ -1963,6 +1964,7 @@ struct ndpi_detection_module_struct *ndpi_init_detection_module(ndpi_init_prefs
 
 #ifdef NDPI_ENABLE_DEBUG_MESSAGES
   set_ndpi_debug_function(ndpi_str, (ndpi_debug_function_ptr) ndpi_debug_printf);
+  NDPI_BITMASK_RESET(ndpi_str->debug_bitmask);
```
utoni, line 1967: You've created the fancy function `ndpi_set_debug_bitmask` but not using it here. Why?

src/lib/ndpi_main.c
```diff
@@ -6247,6 +6249,12 @@ u_int ndpi_get_ndpi_detection_module_size() {
   return(sizeof(struct ndpi_detection_module_struct));
 }
 
+void ndpi_set_debug_bitmask(struct ndpi_detection_module_struct *ndpi_str, NDPI_PROTOCOL_BITMASK debug_bitmask) {
+#ifdef NDPI_ENABLE_DEBUG_MESSAGES
```
utoni, line 6255: I would recommend to put the preprocessor conditional one line up as this function is only mandatory if `NDPI_ENABLE_DEBUG_MESSAGES` enabled.

example/reader_util.h
```diff
@@ -336,4 +336,15 @@ float ndpi_flow_get_byte_count_entropy(const uint32_t byte_count[256], unsigned
 
 extern int nDPI_LogLevel;
 
+/* TODO: conditional compilation? */
```
utoni: Would've make sense to me. Either use `NDPI_ENABLE_DEBUG_MESSAGES`, `DEBUG_TRACE` or `DEBUG_PAYLOAD`. It's up to you. I would recommend the first one as it should be set with `./configure --enable-debug-messages`.

example/reader_util.h
```diff
@@ -336,4 +336,15 @@ float ndpi_flow_get_byte_count_entropy(const uint32_t byte_count[256], unsigned
...
+#if 1
+ #define LOG(log_level, args...)	\
```
utoni, line 340: Why didn't you use `NDPI_LOG` ? It is already there.

src/lib/protocols/http.c
```diff
@@ -589,8 +589,10 @@ static u_int16_t http_request_url_offset(struct ndpi_detection_module_struct *nd
   int i;
 
   NDPI_LOG_DBG2(ndpi_struct, "====>>>> HTTP: %c%c%c%c [len: %u]\n",
-		non_ctrl(packet->payload[0]), non_ctrl(packet->payload[1]),
```
utoni, line 592: Why removing `non_ctrl` ?

utoni (changes requested):
I'm actually not quite sure why you reinvented the wheel with `LOG()` and moved all those `NDPI_CURRENT_PROTO` after the `ndpi_api.h` include.

src/lib/ndpi_main.c
IvanNardi: AFAIU, it is a little bit more complicated.
This log should be reported at the beginning of ndpi_init_detection_module(), which initializes the library and must be the first function called by the user.
At that point:
* you don't have a detection_module context, so you can't log anything via NDPI_LOG* macros
* the user can't have already called set_ndpi_debug_function(), so a custom log function isn't available yet. We could use the default implementation of ndpi_debug_printf(), but it prints on stdout. Your valid concern about stderr, apply to stdout, too.

Since this allocation error should be uncommon (to say the least), I think we have only two options:
1) don't log anything
2) directly print on stdout or stderr, having such printf protected via NDPI_ENABLE_DEBUG_MESSAGES define

I have no strong opinion one way or the other

src/lib/ndpi_main.c
IvanNardi: This is the initialization, i.e. the default value.
I think, it is better to have a proper value in ndpi_str->debug_bitmask even if the user doesn't call ndpi_set_debug_bitmask()

src/lib/ndpi_main.c
IvanNardi: It is matter of taste. I think that having an ABI that doesn't depend too much on defines is preferable from an user point-of-view.
In the same .c file, set_ndpi_debug_function() has been already implemented with the same logic (I originally copied-and-pasted from it) but I see that the library is not always consistent about that.
No strong opinion

example/reader_util.h
IvanNardi: Ok, thanks for the suggestion

example/reader_util.h
IvanNardi: I thought I explained it in the first commit message, but probably I didn't make myself clear. Let me rephrase it.
From 66b759ca6, the ndpi_detection_module_struct object is an opaque data type for the user so you can't dereference it in reader_utils.c code; if you try, you get a compiler error, something similar to

../src/include/ndpi_define.h:188:26: error: invalid use of undefined type ‘struct ndpi_detection_module_struct’
  188 |     if(mod != NULL && mod->ndpi_debug_printf != NULL)                           \
      |                          ^~
reader_util.c:1775:4: note: in expansion of macro ‘NDPI_LOG’
 1775 |    NDPI_LOG(0, workflow->ndpi_struct, NDPI_LOG_DEBUG, "\n\nWARNING: IPv4 fragments are not handled by this demo (nDPI supports them)\n");
      |    ^~~~~~~~

src/lib/protocols/http.c
IvanNardi: I didn't. The function non_ctrl() has been removed on 8b1c68ae, one year ago!
This is the perfect example to show how the debug message stuff has been broken for a long time. If --enable-debug-messages flag had worked as intended, anyone compiling ndpi with such flag in the last year would have received a compiler error.
I don't think that for such low-level message we need non_ctrl() again, but if anyone thinks otherwise, I could definitely reintroduce it

IvanNardi:
Thanks for the review and for your time
I'll upload a new version in a few days, after having waited for other comments

src/lib/ndpi_main.c
utoni: I see the issue. We could add an "early-logging-callback" as parameter for `ndpi_init_detection_module`.
However this would be an api break.

lucaderi:
@lnslbrty Are we ready to merge this PR?

example/ndpiReader.c
```diff
@@ -740,6 +740,10 @@ static void parseOptions(int argc, char **argv) {
       }
       break;
 
+    case 'u':
+      _debug_protocols = strdup(optarg);
```
utoni, line 748: possible leak

utoni (changes requested):
LGTM

example/ndpiReader.c
IvanNardi: Nice catch

lucaderi:
Thank you all

PR #935, opened 2020-06-27
fixed fuzzing tests in a way that ./tests/do.sh is now able to use co…

utoni:
…rpus *.pcap files from ./tests/pcap

(also unified intendation, using tabs+spaces together freaks me out)


lucaderi:
I agree with this patch some things are better now: thanks

utoni:
I'm probably not done here as removing the *.zip corpus file lets the oss-fuzzing CI job fail as it requires a corpus zipfile..
Will investigate and fix.

PR #937, opened 2020-06-27
Re-Added corpus zipfile generation required for google-oss fuzzer

utoni:
There will be another MR which goes to https://github.com/google/oss-fuzz and will generate the corpus zipfile before fuzzing starts.


lucaderi:
Thanks you

PR #939, opened 2020-06-29
Improved fuzz_ndpi_reader which supports now SMP/MT w/o race-coniditi…

utoni:
…ons.

./tests/do.sh can supports SMP/MT via environment variables.
Removed -fno-sanitize=shift as well, was fixed by 317d3ffd.

(fuzz_ndpi_reader_with_main is required as ndpiReader is sometimes too restrictive results in non reporoducible errors)


lucaderi:
Thank you

PR #940, opened 2020-06-29
Fixed CodeInspector issues.

utoni:
* Removed compiler search list for AC_PROG_CC, AM_PROG_CC_C_O, AC_PROG_CXX, AC_PROG_CC_STDC
   The list rendered the AX_CHECK_COMPILE_FLAG([-fsanitize=fuzzer]) useless as it did use
   clang for AX_CHECK_COMPILE_FLAG but gcc during the compile process. Seems broken somehow.

### DIFF
--- a/configure.seed
+++ b/configure.seed
@@ -13,10 +13,18 @@ AS_IF([test "${with_sanitizer+set}" = set],[
 
 LT_INIT
 
-AC_PROG_CC(clang gcc)
-AM_PROG_CC_C_O(clang gcc)
-AC_PROG_CXX(clang++ g++)
-AC_PROG_CC_STDC(clang gcc)
+SYSTEM=`uname -s`
+if test $SYSTEM = "Darwin"; then
+  AC_PROG_CC(clang gcc)
+  AM_PROG_CC_C_O(clang gcc)
+  AC_PROG_CXX(clang++ g++)
+  AC_PROG_CC_STDC(clang gcc)
+else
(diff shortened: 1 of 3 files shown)


configure.seed
```diff
@@ -13,10 +13,10 @@ AS_IF([test "${with_sanitizer+set}" = set],[
...
 LT_INIT
 
-AC_PROG_CC(clang gcc)
-AM_PROG_CC_C_O(clang gcc)
-AC_PROG_CXX(clang++ g++)
-AC_PROG_CC_STDC(clang gcc)
+AC_PROG_CC
+AM_PROG_CC_C_O
+AC_PROG_CXX
+AC_PROG_CC_STDC
```
lucaderi: Doing this change, on my Mac clang won't be used. I don't think it's a good idea to revert the change. What do you think?

utoni: Strange. Can you post/upload `config.log` w/ and w/o those changes?
Would help me a lot.

lucaderi:
[config.zip](https://github.com/ntop/nDPI/files/4866934/config.zip)

configure.seed
lucaderi: Attached

utoni: Thanks, will investigate.

utoni: keeping my fingers crossed that it will work now

lucaderi:
Thanks

PR #941, opened 2020-06-29
Copy&Paste ./tests/ossfuzz.sh from https://github.com/google/oss-fuzz…

utoni:
…/pull/4041

 * Changing the OSS-Fuzz script from our side is much more easier then
   opening a PR to google/oss-fuzz every time we have to change a single line.
 * https://github.com/google/oss-fuzz/pull/4041 will be updated once this PR is merged


lucaderi:
Thanks

PR #943, opened 2020-06-29
Fixed heap overflow in tls esni extraction triggered by manipulated p…

utoni:
…ackets.


lucaderi:
Good catch!

PR #947, opened 2020-07-02
added shell script to download wireshark traces, can be used in combi…

utoni:
…nation with ./tests/do.sh


lucaderi:
Thanks

Issue #915, opened 2020-06-03
How to extract the browser info using ndpi


subritm:
I wrote a sample function along with ndpiReader to print the packet info and browser info from the flow structure, flow->http.user_agent , but this seems only to be available for the HTTP type that is port 80.

[TCP] SRC IP -- DST IP - HTTP - Mozilla/5.0 (Windows NT 6.1; Win64; x64; rv:77.0) Gecko/20100101 Firefox/77.0
[TCP] SRC IP -- DST IP - Google - Mozilla/5.0 (Windows NT 6.1; Win64; x64; rv:77.0) Gecko/20100101 Firefox/77.0

I also want to determine the browser info for others, like Facebook, Youtube, etc..
Does ndpi already have that info into it, How to extract or add ?

utoni:
HTTP user agents are indeed not parsed for non-80 ports even if HTTP was detected.
Will investigate.

lnslbrty:
Ok, I was wrong. HTTP user agents are indeed parsed for non standard ports as well, if the HTTP protocol was detected.
HTTP UserAgents can only parsed if HTTP is used. IMHO Facebook and Youtube are encrypting basically all traffic. So no UserAgent extraction possible.

utoni:
Could you please check if ~~#948~~ #950 fixes your issue? Thanks.

IvanNardi:
With latest versions, user-agent should always be extracted from HTTP flows.
If that is not the case, please open a new bug.

PR #948, opened 2020-07-02
Improved HTTP line parsing if request splitted into multiple packets.

utoni:
Ignore TCP packets with empty payload as they modify the internal flow state.


utoni:
May also fix #915.

utoni:
(Closed this PR as there is a detection diff which needs some investigation)

PR #952, opened 2020-07-06
Fixed heap overflow caused by missing lengthcheck in reader uutil.

utoni:


lucaderi:
Thanks

PR #954, opened 2020-07-07
fixed memory leak in ndpi_cluster_bins / binUnitTest

utoni:

I'm not quite sure what "cluster bins" is used for, but that fixed at least both leaks.


lucaderi:
Thank you: my fault!

Ideally bins will be used soon for traffic behaviour analysis where DPI will fall short

utoni:
thx for the info

PR #955, opened 2020-07-07
Fixed thunder protocol detection heap overflow caused by missing leng…

utoni:
…thcheck.


utoni:
travis-ci fails due to a memory leak in `printFlowsStats` @ `example/ndpiReader.c:2461:44`

PR #956, opened 2020-07-08
Fix memory leak reported in #955

IvanNardi:
This leak is clearly happening because of missing/disabled code in "bins"
handling.
This is a trivial workaround useful until such code is completed.
In this way tavis-ci will stop complaining


lucaderi:
thanks

PR #958, opened 2020-07-09
More travis, autogen and autoconf improvments.

utoni:
* removed travis-ci CFLAGS/LDFLAGS as same is done via autoconf
 * removed double quotes in autogen, causes errors if multiple configure args used
 * let configure script fail with an exit code != 0 if libpcap-dev not found
 * travis-ci should use ubuntu-bionic for all builds referencing apt

(re-added CXX as required for -ffuzzer)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -67,6 +67,7 @@ matrix:
 
     - os: linux
       compiler: gcc-8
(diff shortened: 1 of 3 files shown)


utoni:
duplicated PR...

PR #959, opened 2020-07-09
More/stuff

utoni:
* fixed memory leak due to missing free of cluster_ids / bins
* tests/do.sh: exit with return value != 0 if ndpiReader returns != 0 e.g. due to a sanitizer error

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -124,7 +124,7 @@ matrix:
         - MATRIX_EVAL="CC=clang-7"
 
     - name: fuzza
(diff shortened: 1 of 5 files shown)


utoni:
There is a leak in the libpcap version shipped with ubuntu/xential.
But switching the fuzza/fuzzl build back to ubuntu/bionic results in install errors for clang-7.
Further investigation needed, not merge ready yet.

utoni:
May work now, hope that I did not forget anything.

So the issue was indeed because libpcap for xential was leaking a constant amount of memory.
Upgrading ASAN/UBSAN job to bionic solved that issue.

PR #960, opened 2020-07-09
Fixed heap overflow caused by missing lengthcheck in 802.11 LLC heade…

utoni:
…r parsing.


lucaderi:
Thanks

PR #961, opened 2020-07-10
Fixed race condition in ndpi_ssl_version2str() caused by static quali…

utoni:
…fier in the version string buffer.

 * added also GREASE supported tls versions as specified in
   https://tools.ietf.org/html/draft-davidben-tls-grease-01#page-4


lucaderi:
Good catch!

PR #963, opened 2020-07-13
add improved boundary check and check malloc return is NULL

yskcg:
### Commit Summary

* add the malloc return and free pointer check
* add the boundary check ,when parse the HTTP protocol headers

### File Changes

* M `example/ndpiReader.c`
* M `src/lib/ndpi_main.c`

### Why changes

* valgrind memcheck give the warnings

### DIFF
--- a/example/ndpiReader.c
+++ b/example/ndpiReader.c
@@ -2943,6 +2943,9 @@ static void ndpi_process_packet(u_char *args,
   /* allocate an exact size buffer to check overflows */
   uint8_t *packet_checked = malloc(header->caplen);
 
+  if(packet_checked == NULL){
+    return ;
+  }
   memcpy(packet_checked, packet, header->caplen);
   p = ndpi_workflow_process_packet(ndpi_thread_info[thread_id].workflow, header, packet_checked, csv_fp);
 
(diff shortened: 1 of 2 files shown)


example/ndpiReader.c
```diff
@@ -3047,7 +3050,10 @@ static void ndpi_process_packet(u_char *args,
      Leave the free as last statement to avoid crashes when ndpi_detection_giveup()
      is called above by printResults()
   */
-  free(packet_checked);
+  if(packet_checked){
```
utoni, line 3053: NUL check already done

src/lib/ndpi_main.c
```diff
@@ -4928,7 +4928,7 @@ void ndpi_parse_packet_line_info(struct ndpi_detection_module_struct *ndpi_str,
   packet->line[packet->parsed_lines].len = 0;
 
   for (a = 0; ((a+1) < packet->payload_packet_len) && (packet->parsed_lines < NDPI_MAX_PARSE_LINES_PER_PACKET); a++) {
-    if((packet->payload[a] == 0x0d) && (packet->payload[a+1] == 0x0a)) {
+    if(((a + 1) < packet->payload_packet_len) &&(packet->payload[a] == 0x0d) && (packet->payload[a+1] == 0x0a)) {
```
utoni, line 4931: Isn't it already checked in the loop condition?

utoni (changes requested):

lucaderi:
Thanks, nice checks but don't forget that this is just a demo app

src/lib/ndpi_main.c
yskcg: Yeah，U are right,I don't attention the loop case;

PR #964, opened 2020-07-16
1:add free pointer NULL check;2:fix xbox and teredo protocol detected…

yskcg:
### Commit Summary

* 	when free pointer ,add the `NULL` check;
*  fix protocol `xbox` and `teredo` will be detected error when together use udp port `3544`;

### File Changes

* src/lib/ndpi_main.c
* src/lib/protocols/xbox.c 

### Why 
As there is no perfect method to distinct the protocol `xbox ` and `teredo` when use `ndpi_get_guessed_protocol_id` to get the protocol,if `xbox and teredo` all set the defaults udp port `3544`,I think there are other protocols will be mixed when use the same port at present code structural;

note**) 

I just add 
   `NDPI_BITMASK_RESET(all);`
   
   `NDPI_BITMASK_ADD(all,NDPI_PROTOCOL_DNS);`
   
   `NDPI_BITMASK_ADD(all,NDPI_PROTOCOL_HTTP);`
   
   `NDPI_BITMASK_ADD(all,NDPI_PROTOCOL_DHCP);`
  
  ` NDPI_BITMASK_ADD(all,NDPI_PROTOCOL_XBOX);` 
 
**not add `NDPI_PROTOCOL_TEREDO `** ,if i add `NDPI_PROTOCOL_TEREDO ` it detecte correct

### DIFF
--- a/src/lib/ndpi_main.c
+++ b/src/lib/ndpi_main.c
@@ -100,10 +100,16 @@ void *ndpi_calloc(unsigned long count, size_t size) {
 /* ****************************************** */
 
 void ndpi_free(void *ptr) {
-  if(_ndpi_free)
-    _ndpi_free(ptr);
-  else
-    free(ptr);
+  if(_ndpi_free){
+    if(ptr){
+      _ndpi_free(ptr);
+    }
+  }else{
+    if(ptr){
+      free(ptr);
+    }
+  }
+    
 }
 
 /* ****************************************** */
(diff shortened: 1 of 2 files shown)


src/lib/protocols/xbox.c
```diff
@@ -85,6 +85,12 @@ void ndpi_search_xbox(struct ndpi_detection_module_struct *ndpi_struct, struct n
 	ndpi_int_xbox_add_connection(ndpi_struct, flow);
 	NDPI_LOG_INFO(ndpi_struct, "found xbox udp port connection detected\n");
 	return;
+    }else if(dport == 3544 || sport == 3544)
```
lucaderi: This should not be necessary as ndpi via guess should handle that

src/lib/ndpi_main.c
```diff
@@ -4928,7 +4934,7 @@ void ndpi_parse_packet_line_info(struct ndpi_detection_module_struct *ndpi_str,
   packet->line[packet->parsed_lines].len = 0;
 
   for (a = 0; ((a+1) < packet->payload_packet_len) && (packet->parsed_lines < NDPI_MAX_PARSE_LINES_PER_PACKET); a++) {
-    if(((a + 1) < packet->payload_packet_len) &&(packet->payload[a] == 0x0d) && (packet->payload[a+1] == 0x0a)) {
+    if((packet->payload[a] == 0x0d) && (packet->payload[a+1] == 0x0a)) {
```
lucaderi, line 4937: Why did you remove this check?

src/lib/ndpi_main.c
```diff
@@ -100,10 +100,16 @@ void *ndpi_calloc(unsigned long count, size_t size) {
 /* ****************************************** */
 
 void ndpi_free(void *ptr) {
```
lucaderi, line 102: If your patch is necessary then it means that we're freeing NULL pointers. Can you explain here this happens in the code? Better to fix the root cause.

lucaderi (changes requested):

src/lib/ndpi_main.c
yskcg: In the for loop case ,is exist

src/lib/protocols/xbox.c
yskcg: ok

src/lib/ndpi_main.c
yskcg: if the ptr is NULL,it's not will be free;In the file ndpi_classify.c ,function ndpi_classify, use the ndpi_free  will have the risk free NULL pointer
`if(!merged_lens || !merged_times) {
    ndpi_free(merged_lens);
    ndpi_free(merged_times);
    return(score);
  }`

PR #953, opened 2020-07-06
fix broken sanitizer CFLAGS, --enable-fuzztargets should also set -fs…

utoni:
…anitize

Signed-off-by: Toni Uhlig <matzeton@googlemail.com>

Can you verify if that works for Mac OS X as well. Thanks!

PR #965, opened 2020-07-19
fixes autoconf issue on ubuntu 18 (#953)

utoni:


IvanNardi:
It fixes the issue reported in #953. Thank you very much!

PR #966, opened 2020-07-20
Enables fuzz targets even without sanitizer

catenacyber:
Should fix ossfuzz build

Continues https://github.com/google/oss-fuzz/pull/4077

cc @lnslbrty 
This kinds of reverts 9bbcc2548e7084a38a76e292e52b38158fd0a667
But this should fix ossfuzz build which relies on setting itself `CFLAGS` with what it wants as sanitizers
What do you think ?


utoni:
It still may be a good idea to use -g3 and -O0 for fuzz builds as it helps a lot finding the cause for *SAN reported errors.

catenacyber:
> It still may be a good idea to use -g3 and -O0 for fuzz builds as it helps a lot finding the cause for *SAN reported errors.

Indeed it is a good idea, but in my humble opinion, it should be independent.
It seems oss-fuzz sets `CFLAGS=-O1 ` for fuzzing

lucaderi:
@catenacyber Note that you patch raised a problem (not sure it introduced it but for sure now it shows up)

configure: error: --enable-fuzztargets requires -fsanitize=fuzzer which is only supported by LLVM
586The command "if [ -n "$QA_FUZZ" -a "$QA_FUZZ" != "msan" ]; then ./configure --enable-fuzztargets ; else ./configure ; fi" exited with 1.

Any clue? See https://travis-ci.org/github/ntop/nDPI/builds/710681334?utm_medium=notification&utm_source=email

PR #967, opened 2020-07-21
Ssh signature checking

edoermini:


src/include/ndpi_typedefs.h
```diff
@@ -76,6 +76,8 @@ typedef enum {
   NDPI_TLS_NOT_CARRYING_HTTPS,
   NDPI_SUSPICIOUS_DGA_DOMAIN,
   NDPI_MALFORMED_PACKET,
+  NDPI_SSH_OBSOLETE_CLIENT_SIGNATURE,
+  NDPI_SSH_OBSOLETE_SERVER_SIGNATURE,
```
IvanNardi: Please read the comment at the beginning of this enumeration: you should update some other stuff, too

src/lib/protocols/ssh.c
```diff
@@ -60,17 +62,102 @@ static void ndpi_search_ssh_tcp(struct ndpi_detection_module_struct *ndpi_struct
...
+/* ************************************************************************ */
+
 static void ssh_analyse_signature_version(struct ndpi_detection_module_struct *ndpi_struct,
 					  struct ndpi_flow_struct *flow,
 					  char *str_to_check,
 					  u_int8_t is_client_signature) {
 
+  if (str_to_check == NULL) return;
+  
+  char *copy = (char*)malloc(sizeof(char)*strlen(str_to_check));
```
IvanNardi: Missing free?

src/lib/protocols/ssh.c
```diff
@@ -60,17 +62,102 @@ static void ndpi_search_ssh_tcp(struct ndpi_detection_module_struct *ndpi_struct
   
 /* ************************************************************************ */
 
+static int ssh_has_old_signature(char *signature) {
+  int is_old = 0;
+  int i = 0;
+
+  char *old_versions[46] = {
```
lucaderi: This is a linear scan that could be avoided if you could parse the version as numeric field instead of treating the version as a string. In essence anything other than OpenSSH 7.x looks obsolete.

src/lib/protocols/ssh.c
```diff
@@ -60,17 +62,102 @@ static void ndpi_search_ssh_tcp(struct ndpi_detection_module_struct *ndpi_struct
...
+  signature = strtok_r(NULL, "-", &rest);
+
+  if (signature == NULL) return;
+
+  obsolete_ssh_version = ssh_has_old_signature(signature);
+
+  #ifdef SSH_DEBUG
+    if(obsolete_ssh_version)
+      printf("[SSH] %s: obsolete signature\n", signature);
+  #endif
```
lucaderi: In addition to the string check, a more effective check is to verify the cipher list ad explained in https://www.linuxminion.com/deprecated-ssh-cryptographic-settings/

lucaderi (changes requested):
Please perform the changes requested in order to merge this PR

src/lib/protocols/ssh.c
```diff
@@ -57,20 +60,62 @@
...
-    NDPI_SET_BIT(flow->risk, is_client_signature ? NDPI_SSH_OBSOLETE_CLIENT_SIGNATURE : NDPI_SSH_OBSOLETE_SERVER_SIGNATURE);
+    string example: SSH-2.0-OpenSSH_5.3
   */
+  strcpy(copy, str_to_check);
+
+  /* SSH */
+  strtok_r(copy, "-", &rem);
+
+  /* 2.0 */
+  strtok_r(NULL, "-", &rem);
```
lucaderi: Too much code. Use sscanf instead

src/lib/protocols/ssh.c
```diff
@@ -54,23 +59,46 @@
   that usually is packet 14
 */
 
-/* #define SSH_DEBUG 1 */
+#define SSH_DEBUG 1
```
utoni: Please disable that `#define`.
My nDPI integration does not like writing stuff to `stdout`.

src/lib/protocols/ssh.c
```diff
@@ -54,23 +59,46 @@
...
 					  u_int8_t is_client_signature) {
 
+  if (str_to_check == NULL) return;
+
+  int major = 0;
+  int minor = 0;
+  int patch = 0;
+  int obsolete_ssh_version = 0;
+
+  if (sscanf(str_to_check, "SSH-%*f-OpenSSH_%d.%d.%d", &major, &minor, &patch) < 0)
```
utoni: IMHO that should be more generalized as there are **lots** of other SSH implementation available besides `OpenSSH`.

utoni (changes requested):

utoni (changes requested):
Please squash some of your commits. `git log` should stay as clean as possible.

src/lib/protocols/ssh.c
lucaderi: if(sscanf(....)  != 3)...
Do no use generic checks <0, >0 etc

src/lib/protocols/ssh.c
```diff
@@ -57,43 +62,95 @@
...
 
-  for(i=0; i<cipher_len; i++)
-    printf("%c", cipher[i]);
+  char *rem;
+  char *cipher;
+  int found_obsolete_cipher = 0;
 
-  printf("\n");
-#endif
+  char *obsolete_ciphers[6] = {
```
lucaderi: Do not put a size (6) but add a NULL at the end of this data structure.Also please use "const char*..."

src/lib/protocols/ssh.c
```diff
@@ -57,43 +62,95 @@
...
+  char *copy = (char*)ndpi_calloc(cipher_len, sizeof(char));
+
+  if (strncpy(copy, ciphers, cipher_len) == NULL) 
+    return;
+
+  cipher = strtok_r(copy, ",", &rem);
+
+  while (cipher && !found_obsolete_cipher) {
+
+    for (int i = 0; i < 6; i++) {
```
lucaderi: Here scan until NULL and not up to 6

src/lib/protocols/ssh.c
```diff
@@ -57,43 +62,95 @@
...
+  char *obsolete_ciphers[6] = {
+    "arcfour256",
+    "arcfour128",
+    "3des-cbc",
+    "blowfish-cbc",
+    "cast128-cbc",
+    "arcfour",
+  };
+
+  char *copy = (char*)ndpi_calloc(cipher_len, sizeof(char));
```
lucaderi, line 154: You don't check if memory has been allocated really. Crash can happen

src/lib/protocols/ssh.c
```diff
@@ -57,43 +62,95 @@
...
-#ifdef SSH_DEBUG
-  u_int i;
-  
-  printf("[%s] ", is_client_signature ? "CLIENT" : "SERVER");
 
-  for(i=0; i<cipher_len; i++)
-    printf("%c", cipher[i]);
+  char *rem;
+  char *cipher;
+  int found_obsolete_cipher = 0;
```
lucaderi: Change it to u_int8_t

src/lib/protocols/ssh.c
```diff
@@ -57,43 +62,95 @@
...
 					  u_int8_t is_client_signature) {
 
+  if (str_to_check == NULL) return;
+
+  int major = 0;
+  int minor = 0;
+  int patch = 0;
+  int obsolete_ssh_version = 0;
+
+  if (sscanf(str_to_check, "SSH-%*f-OpenSSH_%d.%d.%d", &major, &minor, &patch) != 3)
```
lucaderi: Did you search fo other SSH implementations other than OpenSSH? Please do

lucaderi (changes requested):

src/lib/protocols/ssh.c
utoni: FYI: https://en.m.wikipedia.org/wiki/Comparison_of_SSH_servers

It would be very nice to have at least additional support for dropbear SSH as that implementation is used in embedded environments.

lucaderi:
Thank you

PR #968, opened 2020-07-22
Reenables CI with sanitizers

catenacyber:
Follows #966 
cc @lnslbrty

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -120,7 +120,7 @@ matrix:
         - MATRIX_EVAL="CC=clang-7"
 
     - name: fuzza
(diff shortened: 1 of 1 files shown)


utoni:
Thanks @catenacyber.

I'm somehow confused why we need `CXXFLAGS` to build pure C executables.

catenacyber:
The trick is that libFuzzer (from LLVM) is written in C++ and when we use `clang -fsanitize=fuzzer`, we build an executable with its C++ main function

utoni:
Got it now, thanks for clarification!

PR #970, opened 2020-07-26
Minor fixes

IvanNardi:
Fix a memory leak and an issue (re)-introduced lately in configure script

### DIFF
--- a/configure.seed
+++ b/configure.seed
@@ -9,8 +9,8 @@ AC_ARG_ENABLE(fuzztargets, AS_HELP_STRING([--enable-fuzztargets], [Enable fuzz t
 AM_CONDITIONAL([BUILD_FUZZTARGETS], [test "x$enable_fuzztargets" = "xyes"])
 
 AS_IF([test "${with_sanitizer+set}" = set],[
-  CFLAGS+="${CFLAGS} -g3 -O0 -Wno-unused-value -fsanitize=address -fsanitize=undefined -fno-sanitize=alignment -fsanitize=leak -fno-omit-frame-pointer"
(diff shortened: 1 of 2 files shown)


configure.seed
```diff
@@ -9,8 +9,8 @@ AC_ARG_ENABLE(fuzztargets, AS_HELP_STRING([--enable-fuzztargets], [Enable fuzz t
 AM_CONDITIONAL([BUILD_FUZZTARGETS], [test "x$enable_fuzztargets" = "xyes"])
 
 AS_IF([test "${with_sanitizer+set}" = set],[
-  CFLAGS+="${CFLAGS} -g3 -O0 -Wno-unused-value -fsanitize=address -fsanitize=undefined -fno-sanitize=alignment -fsanitize=leak -fno-omit-frame-pointer"
-  LDFLAGS+="${LDFLAGS} -fsanitize=address -fsanitize=undefined -fno-sanitize=alignment -fsanitize=leak"
+  CFLAGS="${CFLAGS} -g3 -O0 -Wno-unused-value -fsanitize=address -fsanitize=undefined -fno-sanitize=alignment -fsanitize=leak -fno-omit-frame-pointer"
```
lucaderi, line 12: I am not a fuzzy expert, so you need to check with @catenacyber before changing this line again

IvanNardi: Seems fair.
Just to help the review of @catenacyber, this is the context of the issue I am facing
https://github.com/ntop/nDPI/pull/953#issuecomment-654868804

catenacyber: Looks good to me
Should be `CFLAGS="${CFLAGS} someflafs` or `CFLAGS+=someflags` but not both additions

Change comes from commit 9bbcc2548 from @lnslbrty

PR #973, opened 2020-08-05
Add risk flag about suspicious ESNI usage

IvanNardi:
In a Client Hello, the presence of both SNI and ESNI may obfuscate the real domain of an HTTPS connection, fooling DPI engines and firewalls, similarly to Domain Fronting.

Such technique is reported in a presentation at DEF CON 28:
"Domain Fronting is Dead, Long Live Domain Fronting: Using TLS 1.3 to evade censors, bypass network defenses, and blend in with the noise"
Full credit for the idea must go the original author

At the moment, the only way to get the pdf presention and related video is via
https://forum.defcon.org/node/234492
Hopefully a direct link (and an example pcap) will be available soon

I am not sure how useful in practice all this stuff is (I mean, ESNI itself is far, far away from a standard thing), but the code added in this PR is really minimal and trivial


src/lib/protocols/tls.c
```diff
@@ -1432,6 +1432,11 @@ int processClientServerHello(struct ndpi_detection_module_struct *ndpi_struct,
 	      NDPI_SET_BIT(flow->risk, NDPI_TLS_NOT_CARRYING_HTTPS);
 	    }
 
+	    if(flow->protos.stun_ssl.ssl.encrypted_sni.esni &&
+	       flow->protos.stun_ssl.ssl.client_requested_server_name[0] != '\0') {
```
utoni, line 1436: indent - 1

utoni (approved):
LGTM. Thanks.

src/lib/protocols/tls.c
IvanNardi: > indent - 1

Sorry but I don't see any indentation issue. For the sake of coherence, I tried to indent this if-block like the previous one: one tab + spaces.
What am I missing?

utoni: Sorry, my fault. I wasn't seeing the obvious.

utoni (approved):

lucaderi:
This is an interesting addition. ESNI will arrive, its' just a matter of this so I think that what you did it's very useful

PR #974, opened 2020-08-06
Suspicious ESNI usage: add a comment and a pcap example

IvanNardi:
See: 79b89d286605635f15edfe3c21297aaa3b5f3acf

Sorry for the double PR, but they just made the documentation easily available


lucaderi:
Interesting topic

PR #977, opened 2020-08-09
Enable building on OpenBSD 6.7

adek05:
The major blocker is incompatibility of interfaces and types returned for packet timestamps. BSD retunrs bpf_timeval, while non-BSD systems return timeval. These two types are essentially the same, so I am introducing a third type which is aliased to one or another depending on the operating system used.

### DIFF
--- a/src/include/ndpi_includes.h
+++ b/src/include/ndpi_includes.h
@@ -57,7 +57,7 @@
 #if defined __NetBSD__ || defined __OpenBSD__
 #include <netinet/in_systm.h>
 
-#ifdef __OpenBSD__
+#if defined __OpenBSD__
 #include <pthread.h>
 
 #endif
@@ -66,5 +66,9 @@
 #endif
 
 #endif	/* Win32 */
+
+#if defined __OpenBSD__
+#include "ndpi_includes_OpenBSD.h"
+#endif /* __OpenBSD__ */
 
 #endif /* __NDPI_INCLUDES_H__ */
--- a/src/include/ndpi_includes_OpenBSD.h
+++ b/src/include/ndpi_includes_OpenBSD.h
@@ -1,0 +1,43 @@
+/*
+ * ndpi_includes_OpenBSD.h
+ *
+ * Copyright (C) 2011-16 - ntop.org
+ *
+ * This file is part of nDPI, an open source deep packet inspection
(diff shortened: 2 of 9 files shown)


src/include/ndpi_includes_OpenBSD.h
```diff
@@ -0,0 +1,43 @@
...
+#endif /* IPPROTO_SCTP */
+
+#endif /* __OpenBSD__ */
+
+
+#ifdef __OpenBSD__
+#include <net/bpf.h>
+typedef struct bpf_timeval pkt_timeval;
+#else
+typedef struct timeval pkt_timeval;
```
utoni, line 40: This line is not reached on non OpenBSD platforms, so the typedef is not known.

The issue is related to a pre processor conditional in ndpi_includes.h

src/include/ndpi_includes.h
```diff
@@ -57,7 +57,7 @@
 #if defined __NetBSD__ || defined __OpenBSD__
 #include <netinet/in_systm.h>
 
-#ifdef __OpenBSD__
```
utoni, line 60: Is `#ifdef` not supported by OpenBSD's CPP?

utoni (changes requested):
I would actually recommend to merge `ndpi_includes_OpenBSD.h` with `ndpi_includes.h` as the first header does not contain OpenBSD exclusive code.

It is somehow confusing reading a header file tied to a certain platform but in fact it contains code used for all platforms.

src/include/ndpi_includes.h
adek05: Oh, this is supported. I think I made it by mistake.

lucaderi: @adek05 Can you please update this PR so I can merge it?

utoni: @lucaderi I will take over this PR ASAP.

lucaderi:
Thank you

adek05:
Apologies. I was out for a week. Glad that this got merged!

lucaderi:
Please check if everything works on OpenBSD after my changes. Thank you

adek05:
Builds just fine on 59ac73b37f6d058b9058bc9c6e7d652f8f0eeb90.

PR #983, opened 2020-08-15
Replaced obsolete libpcap pcap_lookupdev with pcap_findalldevs.

utoni:


lucaderi:
Thank you

PR #986, opened 2020-08-18
Fix/openbsd build

utoni:


lucaderi:
@lnslbrty Do we still need to merge this PR or https://github.com/ntop/nDPI/pull/977 is enough?

utoni:
PR #977 should do the job. But I did not verify that.

lucaderi:
Ok so let's close this PR for the time being. Thanks

utoni:
Oh wait. #977 was broken, wasn't it?

PR #987, opened 2020-08-19
Updated MySQL protocol detection to support server version 8.

utoni:


lucaderi:
Thank you

Issue #934, opened 2020-06-27
add support of (IETF) QUIC


alagoutte:
There is now more on more IETF QUIC ( https://quicwg.org/ ) with client (Firefox Nighty, ntctp2) and server (nginx, picoquic...) (See [Implementation ](https://github.com/quicwg/base-drafts/wiki/Implementations)

it will be nice also to support IETF QUIC

You can found pcap (of last draft) on Wireshark [bug 13881](https://bugs.wireshark.org/bugzilla/show_bug.cgi?id=13881)

lucaderi:
Good idea, thank you. Sample pcap https://bugs.wireshark.org/bugzilla/attachment.cgi?id=17806

alagoutte:
There is multiple draft release actually used, it is better to support multiple version...

You can use the A Guide to Parsing QUIC Client Hellos for Network Middlebox Vendors from #933, it is also explain for IETF QUIC

Issue #933, opened 2020-06-27
Add support of gQUIC Q050


alagoutte:
With last release of gQUIC (Q050), the packet use Initial Handshake from IETF QUIC but always the gquic crypto but with a specifc salt

more information about this is available here [A Guide to Parsing QUIC Client Hellos for Network Middlebox Vendors](https://docs.google.com/document/d/1GV2j-PGl7YGFqmWbYvzu7-UNVIpFdbprtmN9tt6USG8/preview)

A pcap is available https://bugs.wireshark.org/bugzilla/show_bug.cgi?id=16633

PR #989, opened 2020-08-22
Improve QUIC detection

IvanNardi:
This PR provides a complete rework of QUIC dissector to handle both GQUIC and IETF-QUIC.

First things first: credits. Most of the crypto code used to decrypt QUIC payload and header have been shamelessly taken from Wireshark repository. All credits for that code should go to the original authors.

There are two major critical points to take into account before merging this PR:
1) A dependency has been added on the external library libgcrypt to handle the low-level crypto primitives. Besides the usual fuss to manage an external dependency, the usage of libgcrypt itself is not so trivial: details are provided in the commit message
2) This PR has been tested only on x86_64, Ubuntu 18.04 (gcc and clang, both)

From a technical point of view:
* the first commit improves GQUIC detection and adds support for IETF-QUIC;
* the second commit adds the autoconf magic to use libgcrypt and adds its initialization code;
* the third commit updates TLS dissector to handle QUIC flows, too;
* the last commit finally enables sub-classification of QUIC flows.

Lots of details are provided in the commit messages.

The code has been written with readability in mind: handling all these QUIC versions and all these packet formats is quite complex. Hopefully in a few months, with a final QUIC RFC, we could drastically simplify the code getting rid of GQUIC and all the temporary versions of QUIC: at that point we can focus on performance, IMO.

There are lots of codes in this PR, so there will be a lot of bugs, too. Please share any pcaps triggering any issues


lucaderi:
This looks like a great contribution. I have fixed a compilation issue but so far looks good to me. Many thanks.

This PR fixes  #934 and #933

IvanNardi:
Thanks to you for taking care of all the fallout in
Travis/Makefiles/Readmes...

Issue #990, opened 2020-08-24
FTBFS: nDPI/example/../src/lib/libndpi.a(quic.o): undefined reference to symbol 'gcry_cipher_setiv@@GCRYPT_1.6'


jean-christophe-manciot:
Ubuntu groovy
binutils: 2.35-2ubuntu1 (for /usr/bin/ld)
gcc 4:10.1.0-1ubuntu1
make: 4.3-4ubuntu1
dpdk: v20.08
ndpi: 1.8-2268-gb8307eb

Compiling DPDK with:
```
export CFLAGS+=-Wno-error
make config T=x86_64-native-linuxapp-gcc
make
```
then nDPI with:
```
./autogen.sh
./configure  --build=x86_64-pc-linux-gnu \
             --prefix=/usr --sysconfdir=/etc --localstatedir=/var
make
```
leads to:
```
...
[...]

PR #991, opened 2020-08-24
QUIC: minor fixes

IvanNardi:
LGTM found a real issue on a boundary check
Fix unit tests: a pcap ha been uploaded twice (with different names)
Fix compilation when using DPDK (see #990)


lucaderi:
Thanks

PR #992, opened 2020-08-24
Fixed broken pkg-config file which did not care about gcrypt/pcre.

utoni:


lucaderi:
Thanks

PR #998, opened 2020-08-28
Fixed invalid memory access leading to a SIGSEGV in ndpiReader's option parser.

utoni:

### DIFF
--- a/example/ndpiReader.c
+++ b/example/ndpiReader.c
@@ -436,7 +436,6 @@ static struct option longopts[] = {
   { "capture", no_argument, NULL, '5'},
   { "extcap-capture-filter", required_argument, NULL, '6'},
   { "fifo", required_argument, NULL, '7'},
-  { "debug", no_argument, NULL, '8'},
   { "ndpi-proto-filter", required_argument, NULL, '9'},
 
   /* ndpiReader options */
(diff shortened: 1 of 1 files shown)


example/ndpiReader.c
```diff
@@ -436,7 +436,6 @@ static struct option longopts[] = {
   { "capture", no_argument, NULL, '5'},
   { "extcap-capture-filter", required_argument, NULL, '6'},
   { "fifo", required_argument, NULL, '7'},
-  { "debug", no_argument, NULL, '8'},
```
IvanNardi, line 439: AFAIK "debug" is an extcap option. It is probably better to keep it to be (fully) compatible with Wireshark interface

lucaderi: @IvanNardi Why Wireshark should be involved in this?

IvanNardi: "--debug" option has been added in d8b2189c, when the extcap interface has been implemented.
Wireshark provides a common interface/code to enable debug in extcap programs via GUI or via tshark  and such option is part of it.
But I just checked and ndpiReader unfortunately never fully implemented such functionality so you can safely remove this option.
Very sorry for the noise

utoni: AFAIS https://www.wireshark.org/docs/man-pages/extcap.html does not mention anything about `--debug` but maybe I'm missing something here.

@IvanNardi Nah, there is no noise. Any hint, any question that keeps the project sane&stable should be appreciated. So, thanks for the !noise.

IvanNardi: Wireshark provides a lot of code to drastically ease development of extcap programs:
https://gitlab.com/wireshark/wireshark/-/blob/master/extcap/extcap-base.h
Some of it implicitly defines a mandatory interface (described in the man page), and the rest is convenient, but optional: the debug stuff is obviously optional
I thought that ndpiReader used that common code but I was wrong: it implements the mandatory stuff from scratch

utoni: Ok, thanks for clarification. What do you think? Re-add `--debug` and redirect it to `-V` (or re-add `-8` as well) ?

IvanNardi: Delete it. To have debug functionality from Wireshark/tshark you need more code anyway; this option is not sufficient.
You are not removing any functionality....

PR #999, opened 2020-08-30
QUIC: add support for GQUIC T050 and T051

IvanNardi:
QUIC versioning wasn't complex enough without T05X family...
These versions are very similar to Q050, but use TLS as their handshake
protocol.


lucaderi:
Good stuff!

PR #1003, opened 2020-09-01
Fixed false positive detection for Skype.SkypeCall (affects at least Cisco HSRP and RADIUS).

utoni:


src/lib/protocols/kerberos.c
```diff
@@ -309,7 +309,7 @@ void ndpi_search_kerberos(struct ndpi_detection_module_struct *ndpi_struct,
 		printf("name_offset=%u [%02X %02X] [byte 0 must be 0x1b]\n", name_offset, packet->payload[name_offset], packet->payload[name_offset+1]);
 #endif
 
-		if(name_offset < (packet->payload_packet_len+1)) {
+		if(name_offset < (packet->payload_packet_len - 1)) {
```
lucaderi, line 312: Ahhh you're right!

lucaderi:
Thanks

PR #1008, opened 2020-09-06
Improved dnscrypt v1/v2 protocol detection (and DoH provider).

utoni:


lucaderi:
Thank you

PR #1012, opened 2020-09-08
QUIC: extract User Agent information

IvanNardi:
Some *QUIC implementations (even some recent ones) leave UA information easily available.
I don't know how long we will be so lucky, but try using them for the time being.

Note that UA is saved in flow->http.user_agent which might not be expected, since these flows are not classified as HTTP


src/lib/protocols/tls.c
```diff
@@ -1365,6 +1372,44 @@ int processClientServerHello(struct ndpi_detection_module_struct *ndpi_struct,
 		    }
 		  }
 		}
+	      } else if(extension_id == 65445 /* QUIC transport parameters */) {
+		u_int16_t s_offset = offset+extension_offset;
+		uint32_t final_offset;
+		int using_var_int = is_version_with_var_int_transport_params(quic_version);
+
+		if(!using_var_int) {
+		  u_int16_t seq_len = ntohs(*((u_int16_t*)&packet->payload[s_offset]));
```
utoni: I would recommend an additional check for `seq_len` against `packet->payload_packet_len` as it seems dangerous to me using that length without any length check.

src/lib/protocols/tls.c
```diff
@@ -1365,6 +1372,44 @@ int processClientServerHello(struct ndpi_detection_module_struct *ndpi_struct,
...
+		} else {
+	          final_offset = MIN(total_len, s_offset + extension_len);
+		}
+
+		while(s_offset < final_offset) {
+		  u_int64_t param_type, param_len;
+
+                  if(!using_var_int) {
+		    param_type = ntohs(*((u_int16_t*)&packet->payload[s_offset]));
+		    param_len = ntohs(*((u_int16_t*)&packet->payload[s_offset + 2]));
```
utoni, line 1400: No length check for `s_offset + 2`.

src/lib/protocols/tls.c
```diff
@@ -1365,6 +1372,44 @@ int processClientServerHello(struct ndpi_detection_module_struct *ndpi_struct,
...
+		    s_offset += 4;
+		  } else {
+		    s_offset += quic_len(&packet->payload[s_offset], &param_type);
```
utoni, line 1406: Probably same here: What happens if the var int stored at s_offset is an 8 byte var int?

src/lib/protocols/tls.c
```diff
@@ -1365,6 +1372,44 @@ int processClientServerHello(struct ndpi_detection_module_struct *ndpi_struct,
...
+		    s_offset += quic_len(&packet->payload[s_offset], &param_len);
```
utoni, line 1411: And here; no length check done.

src/lib/protocols/tls.c
```diff
@@ -1365,6 +1372,44 @@ int processClientServerHello(struct ndpi_detection_module_struct *ndpi_struct,
...
+
+#ifdef DEBUG_TLS
+		  printf("Client SSL [QUIC TP: Param 0x%x Len %d]\n", (int)param_type, (int)param_len);
+#endif
+		  if(param_type==0x3129) {
+#ifdef DEBUG_TLS
+		      printf("UA [%.*s]\n", (int)param_len, &packet->payload[s_offset]);
+#endif
+		      http_process_user_agent(ndpi_struct, flow,
+					      &packet->payload[s_offset], param_len);
```
utoni, line 1425: Due to missing length checks: In the worst case `s_offset` has now a value of `final_offset + 15`. `final_offset` can have the same value as `total_len` which has the same value as `packet->payload_packet_len`. Looks like a possible overflow to me.

utoni (changes requested):
Nice work by the way!

IvanNardi:
> Nice work by the way!

Thanks for the review and for catching these errors!. I'll push another version

IvanNardi:
I don't know if I got everything right, but it should be a step into the right direction. What do you think?

lucaderi:
> This pull request **introduces 1 alert** when merging [8db084a](https://github.com/ntop/nDPI/commit/8db084ab06ae73e26c2373e92b35cbc5407d41bd) into [ea42400](https://github.com/ntop/nDPI/commit/ea4240055d5301b52d9fa934b8d14c126375eb17) - [view on LGTM.com](https://lgtm.com/projects/g/ntop/nDPI/rev/pr-66dcb83f78cd3f4739219ede25a07e73cc6e08f5)
> 
> **new alerts:**
> 
> * 1 for Suspicious packet->payload based integer arithmetic

@IvanNardi Can you please have a look at this before merging?

IvanNardi:
> > This pull request **introduces 1 alert** when merging [8db084a](https://github.com/ntop/nDPI/commit/8db084ab06ae73e26c2373e92b35cbc5407d41bd) into [ea42400](https://github.com/ntop/nDPI/commit/ea4240055d5301b52d9fa934b8d14c126375eb17) - [view on LGTM.com](https://lgtm.com/projects/g/ntop/nDPI/rev/pr-66dcb83f78cd3f4739219ede25a07e73cc6e08f5)
> > **new alerts:**
> > 
> > * 1 for Suspicious packet->payload based integer arithmetic
> 
> @IvanNardi Can you please have a look at this before merging?

I saw this warning. According to the help, its definition is: "An arithmetic operation influenced array access is suspicious if it uses an integer value that is likely to be network-controlled, and may require a closer manual audit."
Not sure what to do about it: in this specific case, the code seems fine to me (boundaries are checked)
Note that LGTM reports other 15 similar warnings in the project, all with the same pattern: cast from packet buffer to some integer data.
https://lgtm.com/projects/g/ntop/nDPI/alerts/?mode=list&id=cpp%2Fpacket-payload-integer-arithmetic

src/lib/protocols/tls.c
```diff
@@ -1365,6 +1373,60 @@ int processClientServerHello(struct ndpi_detection_module_struct *ndpi_struct,
...
+		  }
+		} else {
+	          final_offset = MIN(total_len, s_offset + extension_len);
+		}
+
+		while(s_offset < final_offset) {
+		  u_int64_t param_type, param_len;
+
+                  if(!using_var_int) {
+		    if(s_offset+3 >= final_offset)
```
utoni, line 1397: Not sure here, but shouldn't it be `s_offset+4 >= final_offset` or `s_offset+3 > final_offset` ?

utoni (approved):
@IvanNardi, @lucaderi 
LGTM! :+1:

src/lib/protocols/tls.c
IvanNardi: Neither am I sure... I read 4 bytes from s_offset to s_offset + 3 included. The last byte read should be valid, i.e. < final_offset. So the test seems correct... Am I missing something?

utoni: You are right. Maybe it is just a LGTM false positive.

lucaderi: Can I merge this code?

utoni: There is nothing to argue against merging this PR from my side.

IvanNardi: > There is nothing to argue against merging this PR from my side.

I agree

aouinizied: @IvanNardi Do you think other TAG/Values may be extracted in the future and stored within a quic field in ndpi_flow_struct? For example SCID or ConnID. This can be useful to track flows belonging to the same QUIC session and also QUIC user ID on the server-side.
In this case, maybe we should move the UAI under a ndpi_flow_struc.quic.uai and populate this quic struct incrementally.
Reference: https://petsymposium.org/2019/files/papers/issue3/popets-2019-0046.pdf

aouinizied: I'm thinking of it in scenarios where an app using nDPI will export QUIC tags as a Json or whatever. What do you think about it?

PR #1014, opened 2020-09-09
Improved Teamspeak(3) protocol detection.

utoni:
Removed all those port based "weak" detection.
Added a new pattern which works for all current Teamspeak protocol 3 versions.


lucaderi:
Thank you

PR #1017, opened 2020-09-18
Added support for mingw xcompile.

utoni:
What does work:
- mingw dll build

What does not work:
- CI build (needs some matrix adjustment)
- all binaries which require a libpcap (or similar) library

### DIFF
--- a/src/include/ndpi_win32.h
+++ b/src/include/ndpi_win32.h
@@ -31,6 +31,8 @@
 #define __mingw_forceinline __inline__ __attribute__((__always_inline__,__gnu_inline__))
 #endif
 
+#undef _WIN32_WINNT
+#define _WIN32_WINNT _WIN32_WINNT_WIN8
 #include <winsock2.h>
 #include <windows.h>
 #include <ws2tcpip.h>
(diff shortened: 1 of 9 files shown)


src/include/ndpi_win32.h
```diff
@@ -48,6 +51,17 @@
...
+#ifndef MIN
+#define MIN(X, Y) (((X) < (Y)) ? (X) : (Y))
+#endif
+
+#ifndef IPPROTO_SCTP
+#define IPPROTO_SCTP 132
+#endif
+
+#undef gettimeofday
+#define gettimeofday mingw_gettimeofday
```
gvanem, line 63: Not good for clang-cl:
```
ElasticSearch.cpp(140,7): error: use of undeclared identifier 'mingw_gettimeofday'; did you mean 'gettimeofday'?
      gettimeofday(&tv, NULL);
      ^~~~~~~~~~~~
      gettimeofday
../nDPI/src/include/ndpi_win32.h(63,22): note: expanded from macro 'gettimeofday'
#define gettimeofday mingw_gettimeofday
                     ^
../include/ntop_win32.h(93,12): note: 'gettimeofday' declared here
extern int gettimeofday(struct timeval * tp, struct timezone * tzp);
           ^
```

utoni: Whoops. Totally forgot to test this PR with clang-cl

PR #1024, opened 2020-09-22
Added missing files to `make dist' target which are not required to b…

utoni:
…uild nDPI but still somehow essential.

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -8,4 +8,4 @@ endif
 pkgconfigdir = $(libdir)/pkgconfig
 pkgconfig_DATA = libndpi.pc
 
-EXTRA_DIST = autogen.sh configure.seed
+EXTRA_DIST = README.md CHANGELOG.md CONTRIBUTING.md README.protocols autogen.sh configure.seed wireshark python windows utils packages doc/nDPI_QuickStartGuide.pages doc/nDPI_QuickStartGuide.pdf example/MacOS example/Win32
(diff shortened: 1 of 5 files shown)


.travis.yml
```diff
@@ -138,7 +138,7 @@ matrix:
         - MATRIX_EVAL="CC=clang-7"
 
     - name: fuzza
-      env: CXXFLAGS="-g3 -O0 -DFUZZING_BUILD_MODE_UNSAFE_FOR_PRODUCTION -fsanitize=address -fsanitize=fuzzer-no-link" CFLAGS="-g3 -O0 -DFUZZING_BUILD_MODE_UNSAFE_FOR_PRODUCTION -fsanitize=address -fsanitize=fuzzer-no-link" LDFLAGS="-g3 -O0 -fsanitize=address" QA_FUZZ=asan CC=clang-7 && CXX=clang++-7 ASAN_SYMBOLIZER_PATH=/usr/local/clang-7.0.0/bin/llvm-symbolizer
```
IvanNardi: Just out of curiosity: what is the practical difference between -Oo and -Og? I mean, did you see any real issue with -Oo?
(I saw a similar commit 8e7a707be104f04210b12687c94fdeb7ca53f37b)

utoni: AFAIK, -O0 means no optimizations and -Og enables optimization feature that do not decrease debugging capabilities. I've just used it here to test if it fixes aligned memory access. But unfortunately, I didn't.

IvanNardi:
Thanks for the reply

PR #1027, opened 2020-09-29
don't pass arguments to ./configure as 1 string

borisVanhoof:
this causes the following error if there are arguments after --host :
'checking host system type... config.sub: too many arguments'


utoni (approved):
You're right, `$@` is the right one. I've used `$*` w/o `"` in PR #1024 but this is still not correct.
So you are basically fixing this issue. Thanks.

PR #1039, opened 2020-10-21
Improve skype detection

igorribeiroduarte:


src/lib/protocols/skype.c
```diff
@@ -73,6 +103,21 @@ static void ndpi_check_skype(struct ndpi_detection_module_struct *ndpi_struct, s
 	    ndpi_set_detected_protocol(ndpi_struct, flow, NDPI_PROTOCOL_SKYPE_CALL, NDPI_PROTOCOL_SKYPE);
 	  }
 	}
+
+        if (flow->detected_protocol_stack[0] == NDPI_PROTOCOL_UNKNOWN) {
+          const uint8_t id_flags_iv_crc_len = 11;
+          const uint8_t crc_len = 4;
```
utoni: I would prefer `sizeof(flow->l4.udp.skype_crc)` to document where this magic `4` comes from.

src/include/ndpi_typedefs.h
```diff
@@ -775,6 +775,7 @@ struct ndpi_flow_udp_struct {
 
   /* NDPI_PROTOCOL_SKYPE */
   u_int8_t skype_packet_id;
+  char skype_crc[4];
```
utoni: An `u_int8_t` aka `unsigned char` would fit better as it indicates a non-printable byte buffer.

src/lib/protocols/skype.c
igorribeiroduarte: Fixed.

src/include/ndpi_typedefs.h
igorribeiroduarte: Fixed.

utoni (approved):

lucaderi:
Thank you

PR #1040, opened 2020-10-21
Added new IPs for TeamViewer

dougword:
We searched for new IPs to help identify TeamViewer.

When opening the program it tries do connect to router1.teamviewer.com, router2.teamviewer.com... router16.teamviewer.com. 

So we decided to scan all involved IPs, and checked with their respective reverse DNS. We then grouped them into subnetworks.

We also added their registered ranges related to their ASNs. For this, we searched using whois (whois -h whois.arin.net 'z . teamviewer*') and on these sites: https://bgpview.io/search/teamviewer#results-v4, https://bgp.he.net/AS43304#_prefixes, https://bgp.he.net/AS208187#_prefixes, https://bgp.he.net/AS208175#_prefixes).

After adding theses subnetworks we were able to block TeamViewer, without them we weren't able to block it using nDPI.

More details can be seen on https://pastebin.com/hZ1NEHsY


lucaderi:
Thank you

PR #1047, opened 2020-10-29
Fix/packaging

utoni:
Fixes packaging for Debian, Ubuntu and OpenWrt and adds a CI job for Ubuntu packaging.

I've just copy'n'pasted the nDPI OpenWrt build file from openwrt-packages.
The previous file was broken and I'm keeping OpenWrt's nDPI up2date anyway.


lucaderi:
Thank you

PR #1048, opened 2020-11-03
:bulb: Add mongodb protocol dissector

leonn:


IvanNardi:
Shouldn't  src/include/ndpi_api.h be auto generated by autogen.sh script?

leonn:
@IvanNardi  I was not aware of this, you're right @lucaderi can I submit a fix to this? the only thing affected was the `NDPI_API_VERSION` definition

Issue #1050, opened 2020-11-04
v3.4, ndpi_free() causes SIGSEGV trying to free tcp.tls.message.buffer


RudeDude:
Syntax error probably caused buffer pointer to equal 0x1

A call to ndpi_free() in ndpi_main.c tries to free flow->l4.tcp.tls.message.buffer but pointer equals 0x1 which is inaccessible
Example of my own additional debug print says [TLS Mem] DE-Allocating (0 bytes, 0 used) message buffer at 0x1.

Possible copy-paste from protocols/tls.c lines 141-142 onto line 145? It looks like a comma where a semicolon and newline would be desired. newbuf is checked but assignment goes awry.

P/R diff https://github.com/RudeDude/nDPI/pull/1/files

Relevant stack dump:
Program terminated with signal SIGSEGV, Segmentation fault.
#0 0x00007f1806f42af1 in free () from /lib64/libc.so.6
#1 0x00007f180777328d in ndpi_free (ptr=0x1) at ndpi_main.c:113
ntop#2 0x00007f180777f9a5 in ndpi_init_packet_header (ndpi_str=0x7f17d01c4010, flow=0x7f178488b370, packetlen=76) at ndpi_main.c:3716
ntop#3 0x00007f1807782b9f in ndpi_detection_process_packet (ndpi_str=0x7f17d01c4010, flow=0x7f178488b370,
packet=0x7f177eb83c6a <error: Cannot access memory at address 0x7f177eb83c6a>, packetlen=76, current_time_ms=1604495709604, src=0x7f17845e2dd0, dst=0x7f17845e2eb8)
at ndpi_main.c:4645

RudeDude:
submitted pull request #1051 https://github.com/ntop/nDPI/pull/1051

RudeDude:
ok, this P/R may not actually fix my issue entirely. I put a guard around the offending line in ndpi_main
`if( flow->l4.tcp.tls.message.buffer_len == 0 ) { ... }`

And it just hit for me printing out the info I added:
[CFAI] l4_packet_len: 24 payload_packet_len: 0 actual_payload_len: 0 iph.saddr: 59F8A783 tcp.source: 27689 iph.daddr: 407C4674 tcp.dest: 9600
[CFAI] Skip freeing of (0 bytes, 0 used) message buffer at 0x1.

utoni:
Do you have any PCAP file to share which triggers that SIGSEGV?
[...]

PR #1051, opened 2020-11-04
fixes issue #1050 Syntax error caused buffer pointer to equal 0x1

RudeDude:
Possible copy-paste from lines 141-142


utoni (commented):
The compiler should be able to evaluate expressions from left to right.
In that context the meaning of `,` and `;` is equal.

But I agree with you, it is just easier to read.

utoni:
By the way; there is a lot work to do if you want to replace all comma operators:
`grep -Hrn '^.*=.*,\s.*=.*$' src/lib/`
;)

lucaderi:
Thanks

PR #1052, opened 2020-11-08
Improved SSH protocol detection.

utoni:
Removed the max packet length check as it is unreliable (e.g. for dropbear ssh services).


lucaderi:
Thanks

PR #1056, opened 2020-11-13
nDPI dns dissector extended

Roby73:
the dns dissector has been extended to parse many query/response messages (zone transfer on tcp included)
added host and server variables, for mantaining trace in a client/server exchange;
add tcp fragmentation managing and realignments (prevalently for dns packets)
added a dns-zone-transfer pcap file to test


utoni:
@Roby73: Please take a look into https://travis-ci.org/github/ntop/nDPI/jobs/743478347
It seems like there is something broken.

utoni (commented):
Wouldn't it make more sense to move your changes to the `DNS` protocol detection/dissection into `src/lib/protocols/dns.c`.

AFAIS what you did is to extend the examples with a dissection feature that should probably part of the core library. It make sense to me to have that information available for all nDPI users/developers and not only for the ones using the `./examples`.

Roby73:
> @Roby73: Please take a look into https://travis-ci.org/github/ntop/nDPI/jobs/743478347
> It seems like there is something broken.

Perhaps, it depends from here:
 first:
  _ndpi_connection_tracking(ndpi_str, flow);	  
   /* call the extra packet function (which may add more data/info to flow) */	  
   if(flow->extra_packets_func) {_
after:
  _**if (** ndpi_connection_tracking(ndpi_str, flow) **) {**
    /* call the extra packet function (which may add more data/info to flow) */	 
    if(flow->extra_packets_func) {_	  
I hope to have solved with a further condition on protocol!

Roby73:
> `DNS`
If I understand your comment, I have already extended the dns protocol dissector into the file dns.c/h (and ndpi_typedefs.h), I touched the ndpi_main.c files (and others) to make some of the dns information available in the upper layers and, in the best possible way, manage the reordering of segments received out of order; I also differentiated host and server name. Finally I made these implementations visible to users in ndpiSimpleIntegration, but not in ndpiReader. The library core already contains my changes.
What does "AFAIS" word mean? Thanks R.

utoni:
@Roby73 : AFAIS = As Far As I Can See

What I meant was that `dns_utils` should maybe part of the core library so that additional DNS information can be parsed in the library itself and not only for the examples. And it would also be nice to have this information available via nDPI's serialization interface so my nDPI daemon can make use of that. ;)

Roby73:
> @Roby73 : AFAIS = As Far As I Can See
> 
> What I meant was that `dns_utils` should maybe part of the core library so that additional DNS information can be parsed in the library itself and not only for the examples. And it would also be nice to have this information available via nDPI's serialization interface so my nDPI daemon can make use of that. ;)

Ok, I have done it. Now dns_utils.c has been removed, but dns_utils.h it is preferable to maintain, for its content.
I have not touched files about serialization.

Roby73:
Closed to renew

PR #1057, opened 2020-11-13
Add Virtual Asssitant (Alexa, Siri) support.

aouinizied:
This PR implements the support of a subset of Virtual Assistants (Siri, Alexa). It also introduces a new category VIRTUAL_ASSISTANT.
Future plan is to regroup Alexa, Bixby, GoogleAssitant, Siri within a specific category.

Zied


tests/do.sh
```diff
@@ -5,7 +5,7 @@ cd "$(dirname "${0}")"
 READER="../example/ndpiReader -p ../example/protos.txt -c ../example/categories.txt"
 
 RC=0
-PCAPS=`cd pcap; /bin/ls *.pcap`
+PCAPS=`cd pcap; /bin/ls *.pcap*`
```
utoni: I'm not a fan of using wildcards for filename suffixes as this matches literally anything with a `*.pcap* ` infix e.g. `not.a.pcap.file`.

My recommendation would be `/bin/ls *.pcap *.pcapng`.

utoni (changes requested):

tests/do.sh
aouinizied: @lnslbrty Thanks! I fixed it.

PR #1064, opened 2020-11-16
Implement DGA detection performances tracking workflow.

aouinizied:
This PR implements a workflow to track DGA detection performances using Travis CI.

* a Testset is used for DGA versus NON-DGA
* Current implementation evaluation is used as a baseline.
* On commit: Evaluation is automatically performed.
* If a modification harms the detection performances, the test fails.
* a guideline is provided on how data is obtained and split.

This is mandatory in order to track improvements of current heuristic or future ML-based approach.

Zied


lucaderi:
Interesting work

PR #1067, opened 2020-11-20
Quic fixes

IvanNardi:
Some QUIC improvements:
* fix return value on error path on quic_cipher_init()
* allow dissection of sessions forcing version negotiation


lucaderi:
Thank you

Issue #1072, opened 2020-11-26
Persistent Application Protocol IDs


lucaderi:
Some people complain that numeric nDPi protocol IDs are not persistent. Protocols are removed from nDPI when they are obsolete (i.e. not present in modern traffic) as they keep space that will be unused. This is due to the fact that keeping short the list of protcools makes nDPI more efficient and bitmaps are shorter. Some people complain about this, as when protocol X is removed, X can be used by a totally new different protocol.

As some people do not like this, it is a good time to introduce persistent IDs that won't change and won't be recycled. This way every protocol will have a persistent Id that can be used if necessary instead of the current application protocol that in essence will become just an internal id.

lucaderi:
@lnslbrty @IvanNardi @aouinizied What do you think?

pjaitken:
Internally to nDPI, it might not matter much that application IDs are reused. However nDPI nolonger stands alone: in the SDN world it is now used as part of larger systems. It's crucial that external systems are able to disambiguate applications - so application IDs must be permanent, long-lived, and cannot be re-used.

Two examples:

1. IPFIX / netflow export
Application information can be exported by netflow or IPFIX as described in RFC 6759. IANA has allocated an application engine ID for nDPI¹. Re-using application IDs would result in flows being attributed to the wrong application.
¹ https://www.iana.org/assignments/ipfix/ipfix.xhtml#classification-engine-ids

2. DANOS
DANOS² uses nDPI to identify applications. Firewall rules can be configured to allow or block the identified applications. Communication between the control plane and the dataplane requires persistent application IDs; re-using IDs would result in the wrong applications being configured.
² https://www.danosproject.org

Two proposals:

1. Parity of internal and external IDs.
Each new application is assigned a new ID. Old IDs are never re-used. ndpi_protocol_id_t grows monotonically. Each application only has a single ID which is used both internally and externally.
\+ simple to understand
\- less efficient for nDPI?

2. Separate internal and external IDs.
Each application is allocated an external ID from a monotonically growing list. Mapping functions convert between the internal and external IDs. nDPI's internal IDs may be reused, but external IDs are never reused.
\+ more efficient for nDPI?
\- extra code just for the external IDs

aouinizied:
@lucaderi My first thought was to extend the bitmaps but I wasn't aware that this may harm nDPI performances.
We can start from currently implemented protocols as feed for external_ids and we continue to change internal ones as always and external_ids grow without reuse or change. That's fair enough.

[...]

PR #1090, opened 2020-12-15
DO NOT MERGE: Example implementation of permanent IDs

pjaitken:
Add APIs to map between internal protocol IDs
and permanent external IDs.


pjaitken:
This is an example of how [issue 1072](https://github.com/ntop/nDPI/issues/1072) might be addressed.

src/lib/ndpi_main.c
```diff
@@ -6922,3 +6940,19 @@ int ndpi_check_dga_name(struct ndpi_detection_module_struct *ndpi_str,
...
+
+ndpi_protocol_extid_t ndpi_intid_to_extid(u_int16_t protoId) {
+  if(protoId >= NDPI_LAST_IMPLEMENTED_PROTOCOL)
+    return NDPI_PROTOCOL_EXT_UNKNOWN;
+
+  return(ndpi_ext_protoid[protoId]);
+}
+
+u_int16_t ndpi_extid_to_intid(ndpi_protocol_extid_t extId) {
+  if(extId >= NDPI_PROTOCOL_EXT_MAX)
```
utoni, line 6954: As `exitId` is most likely an `int` (depending on the highest value in `ndpi_protocol_extid_t` enum) a check for `extId < 0` seems reasonable to me.

utoni (commented):
LGTM but one question;

How do we want to handle custom protocol ids? As @IvanNardi mentioned in #1072 it is necessary that public ids and custom ids stay distinct.

lucaderi:
The problem exists and the idea is interesting however I would work differently. As @pjaitken did 
- I would add and "external" or "permanent" id to each protocol
- add a new preference where you can specify whether you want to have non-permanent ids (default, today's policy) or permanent. This way nothing changes for all users
- do not introduce ndpi_intid_to_extid / ndpi_extid_to_intid as they are probably not necessary
- Patch ndpi_detection_process_packet and ndpi_detection_giveup to return a permament id if the preference is set

What do you think?

src/lib/ndpi_main.c
pjaitken: This check is to avoid reading beyond the bounds of ndpi_int_protoid at line 6957.

pjaitken:
@lnslbrty:

> How do we want to handle custom protocol ids? As @IvanNardi mentioned in #1072 it is necessary that public ids and custom ids stay distinct.

First, I think we need to understand why a user needs a permanent ID for a custom protocol. Why don't they submit a PR for their protocol and receive an ID from ntop?

To ensure global uniqueness of the permanent IDs requires either:
* a central registry (eg, the ndpi_protocol_extid_t enum). Users must submit a PR against the registry.
* or, every user incorporates a unique ID into their protocol IDs - eg, we defined that protocol IDs are 32-bit numbers where the upper 16 bits are the defining user's IANA Enterprise Number¹ and the lower 16 bits are an ID assigned by that Enterprise. Permanent nDPI IDs have the topmost bits set to zero.

If global uniqueness isn't required (so, different users could be using the same IDs) then we could use a numbering scheme where a block of ID numbers is allocated from which custom IDs can be taken, and we ensure that such a range is outwith the public IDs that would ever be allocated by ntop. eg IDs 10,000 - 10,9999 are set aside for this purpose. Or IDs with the top bit set are understood to be "permanent custom protocol IDs".

¹ https://www.iana.org/assignments/enterprise-numbers/enterprise-numbers

pjaitken:
@lucaderi 

>     * Patch ndpi_detection_process_packet and ndpi_detection_giveup to return a permament id if the preference is set
> 
> What do you think?

I like this idea.

Update: thinking about this some more: with this scheme, it would not be possible for an external device to know whether the nDPI instance was using permanent, non-permanent, or user-defined ids. Some values might refer to different applications depending which scheme was in use. Or, protocol names might need mapped to different IDs depending which scheme was in use.

One solution would be to provide an API to indicate which scheme is being used. However, that's not useful if the IDs are being examined later in time or in a different location, when/where the API is not available.

A better solution would be to divide the number space so there's no ambiguity. So it seems the number space needs to be divided into three:
* non-permanent IDs
* global permanent IDs (in the nDPI code)
* user defined IDs

If we allow that the IDs are 32-bits, then the first 4 or 8 bits could be used to distinguish the "type", with the remaining bits indicating the ID.

If the "type" is allocated like so:
* 0 = non-permanent IDs
* 1 = global permanent IDs
* 2 = user-defined IDs

Then:
* 0x00000000 - 0x00000000 = non-permanent IDs
* 0x1000000 - 0x1fffffff = global permanent IDs
* 0x20000000 - 0x2fffffff = user-defined permanent IDs
* 0x3xxxxxxx = reserved

Thoughts?

src/lib/ndpi_main.c
utoni: I know. I meant that `extId` is an int and if you check for `NDPI_PROTOCOL_EXT_MAX`, you should also check for `< 0`. That way you can make sure that all values for `extId` are valid indices.

lucaderi:
Hi @pjaitken it is time to address your needs and merge this PR. My suggestion is

- reserve id's range as you suggested
- internal IDs are internal, nobody should be aware of it. nDPI will try to compact protocol IDs to save memory but again this is something the nDPI apps should not be aware of
- all IDs nDPI reports are external, meaning that before the API returns an ID this has to be persistent
- we need to find a solution for bitmaps that are currently limited to 512 and that need to address larger values. 


Suggestions/comments?

lucaderi:
This PR is no longer necessary as nDPI has now the ability to map internal Ids (low numbers all compact) with high user-returned protocolIds (16 bit)

pjaitken:
Thanks @lucaderi.

Issue #1096, opened 2020-12-25
[Bug]replace free to ndpi_free


pengtianabc:
In `ndpi_main.c:ndpi_load_hostname_category`, use `ndpi_strdup`, so it should use `ndpi_free` instead of `free`

lucaderi:
Fixed by https://github.com/ntop/nDPI/pull/1097

PR #1097, opened 2020-12-25
Bugfix for host check

pengtianabc:
this bug is from commit `427002d14` `2020-05-06 00:31:40`
issue: https://github.com/ntop/nDPI/issues/1096


utoni (approved):
makes sense

PR #1105, opened 2021-01-04
QUIC: improve handling of SNI

### LINKED ISSUE #1077: SNI: flow->protos.stun_ssl.ssl.client_requested_server_name vs flow->host_server_name
Where is the right place to save SNI (for QUIC connections)?

At the moment, the situation is not coherent:
* TLS, IETF-QUIC and GQUIC T05x save the SNI in `flow->protos.stun_ssl.ssl.client_requested_server_name` (QUIC code calls `tls::processClientServerHello()`)
* GQUIC != T05x and FB_ZERO save the SNI in `flow->host_server_name` (note that the original QUIC code already did it; FB_ZERO is a dead protocol, but I report it nonetheless for the sake of completeness)

Having the same protocol saving the same information in two different places (depending on its version) is clearly a bug. 
Where should I save the SNI? I think it should be `flow->protos.stun_ssl.ssl.client_requested_server_name`, but I would like to hear some comments on that.
Note that QUIC dissector already sets a lot of `flow->protos.stun_ssl.ssl` fields other than `client_requested_server_name`

Thanks in advance

IvanNardi:
The goal of these patches is to make QUIC  as similar as possible to TLS regarding SNI handling and all these features that users are used to have for TLS traffic (custom categories, risk enumeration, DGA check...)

Please note that I am not been able to test the serialization patch (it has been copied from TLS part)

Close #1077


lucaderi:
Good work

PR #1107, opened 2021-01-04
QUIC: add suppport for DNS-over-QUIC

IvanNardi:
Even if it is only an early internet draft, DoQ has already (at least) one deployed implementation.
See: https://www.zdnet.com/article/ad-blocker-adguard-deploys-worlds-first-dns-over-quic-resolver/
Draft: https://tools.ietf.org/html/draft-huitema-dprive-dnsoquic-00

In the future, if this protocol will be really used, it might be worth to
rename NDPI_PROTOCOL_DOH_DOT in NDPI_PROTOCOL_DOH_DOT_DOQ


lucaderi:
This is a great PR as encrypted DNS will change many things in Internet monitoring

PR #1108, opened 2021-01-05
debug message bugfix

ragostino:
you can not look for memory enlargement, if you print debug message after updating the variables.

### DIFF
--- a/src/lib/protocols/tls.c
+++ b/src/lib/protocols/tls.c
@@ -139,18 +139,18 @@ void ndpi_search_tls_tcp_memory(struct ndpi_detection_module_struct *ndpi_struct
   avail_bytes = flow->l4.tcp.tls.message.buffer_len - flow->l4.tcp.tls.message.buffer_used;
   
   if(avail_bytes < packet->payload_packet_len) {
-    u_int new_len = flow->l4.tcp.tls.message.buffer_len + packet->payload_packet_len;
+    u_int new_len = flow->l4.tcp.tls.message.buffer_len + packet->payload_packet_len - avail_bytes + 1;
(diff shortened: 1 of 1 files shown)


src/lib/protocols/tls.c
```diff
@@ -139,18 +139,18 @@ void ndpi_search_tls_tcp_memory(struct ndpi_detection_module_struct *ndpi_struct
   avail_bytes = flow->l4.tcp.tls.message.buffer_len - flow->l4.tcp.tls.message.buffer_used;
   
   if(avail_bytes < packet->payload_packet_len) {
-    u_int new_len = flow->l4.tcp.tls.message.buffer_len + packet->payload_packet_len;
+    u_int new_len = flow->l4.tcp.tls.message.buffer_len + packet->payload_packet_len - avail_bytes + 1;
```
lucaderi, line 142: Why is this change necessary?

ragostino: When calculating **avail_bytes**, it subtracts from the current buffer size that already used, obtaining the exact number of bytes available (in the buffer).
So if not enough to contain the bytes of the packet, it proceeds with the increase of the bytes: not the exact bytes number, but using the size of the new packet, resulting in a total size equal to the sum of the packet size and previously available bytes.
With this change, it shall increase the buffer size, exactly of amount strictly needed, and _could be useful in saving a few bytes_ during job.

ragostino: PS. About the computational complexity, this change introduces two (2) sums more, for each memory enlargement. So if before we have O(s), now we will have O(3s). Finally, if we care more about memory space, the change is recommended; otherwise if the computational efficiency of the algorithm is more interested, it could be better to discard this change.

PR #1121, opened 2021-01-20
Disable tests that require libgcrypt if --disable-gcrypt set.

### LINKED ISSUE #1117: configure --disable-gcrypt still runs UT that depend on it.
```
./autogen.sh
./configure --disable-gcrypt
make
cd test; do.sh
```

`./configure --disable-gcrypt` should disable those UT that depend on gcrypt, so that running the UT still succeeds.

utoni:
This PR may fix #1117.

There was another semantic issue configure.seed which still needs some testing (especially for mips target).


pjaitken:
Thanks @lnslbrty !

PR #1122, opened 2021-01-21
added TCP segments management 

Roby73:
This changes introduce a tcp segments management: the buffering, the sorting and the reassembling of segments. When segments arrive they are buffered and for each next packet, it checks for order of next and if previous ones are all arrived; then sort them and reassembly. Then it continue with analisis.
To make less invasive changes, there is a define FRAG_MAN, that if defined, enables this changes; otherwise, undefining it, the old code is used.
The flag is introduced into configure.seed and compares into the ndpi_config.h file.
The code has been commented, but if there are any question, I will be pleased to answer you.
Best regards


lucaderi:
This looks like a good addition to nDPI: thank you for your contribution

utoni:
This changeset introduces some memory leaks. It seems to me that fragments aren't correctly free'd on flow end / library shutdown. Investigation required.
But the code is not easy to read and parts (e.g. changes made to configure.seed) are not correct.
I would suggest to revert those two commits.

utoni:
(You can see those memory leaks in the CiFuzz page or by building it manually with `CC=clang CXX=clang++ ./autogen.sh --with-sanitizer --enable-fuzztargets && make` and running `./tests/do.sh` afterwards.)

Roby73:
Hi all.
@lucaderi Please, I am pleased to have been able to give my contribution.
@lnslbrty Thank you for reporting. I will check as soon as possible.

PR #1127, opened 2021-02-03
Fix some memory leaks in reassembler code

IvanNardi:


IvanNardi:
@lucaderi , @Roby73 could you take a look at this change, please? I am not familiar with this code

PR #1132, opened 2021-02-09
Fixed memory leaks caused by conditional free'ing for some TLS connec…

utoni:
…tions.

 * Members of tls_quic struct should also free'd if the detected master protocol is IMAPS / POPS / SMTPS / etc.

### DIFF
--- a/example/reader_util.c
+++ b/example/reader_util.c
@@ -812,7 +812,7 @@ static struct ndpi_flow_info *get_ndpi_flow_info(struct ndpi_workflow * workflow
       LOG(NDPI_LOG_ERROR,
 	       "maximum flow count (%u) has been exceeded\n",
 	       workflow->prefs.max_ndpi_flows);
-      exit(-1);
+      return NULL;
     } else {
       struct ndpi_flow_info *newflow = (struct ndpi_flow_info*)ndpi_malloc(sizeof(struct ndpi_flow_info));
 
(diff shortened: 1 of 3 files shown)


IvanNardi:
I was working  on a similar patch, since I saw the leak too.
The description is correct but I don't know if the fix is correct: tls_quic_stun struct is a member of an union. I think we need to be careful to free only valid pointers.

utoni:
Dangling pointer is indeed a critical issue and there are some possible double/invalid free's in nDPI e.g. `kerberos_buf.pktbuf`, 

There are multiple ways to prevent this. Use` memset()` in `ndpi_flow_malloc`, Set all dynamically allocated pointers in the flow struct to `NULL` after flow allocation (similar to `ndpi_reset_packet_line_info`).

We could also set the (not) correct subprotocol (TLS) which should also work and we wouldn't have to add protocol conditionals for SMTPS / POPS / IMAPS / etc. But for me it is irrational that TLS can be a subprotocol of SMTPS. It just makes more sense vice versa as SMTPS is the payload ("subprotocol") of TLS.

IvanNardi:
Another option might be to move "struct tls_quic_stun" out of the "union protos" (like we already did for most of http fields). Drawback: memory usage because "struct ndpi_flow_struct" will be bigger.
No idea which solution is the best one...

utoni:
I would love to see that `struct ndpi_flow_struct` doesn't get more bloated (if there there is another solution available).
For now a `sizeof(struct ndpi_flow_struct)` returns **2512 bytes**.
As that value might not be the effective memory used for all struct members due to memory alignment and resulting "holes". But it gives an insight about memory usage per flow. And especially for embedded devices, which usually suffer from random access memory shortage, that might be a killer for small embedded routers at some point.

utoni:
So I've decided to go with the protocol conditional spam and added another fix for `fuzz_ndpi_reader`.

PR #1142, opened 2021-02-26
Add support for known and malicious SSL SHA-1 certificate hashes (maintained by SSLBL).

utoni:
As I saw recently, @lucaderi implemented a JA3 blacklist. I had the same idea for `nDPId` but was too slow.
I also wanted a SSL SHA-1 cert blacklist. So that is my solution (pretty similar to the JA3 blacklisting besides the binary<->hex-string conversion stuff).


lucaderi:
Great! I have now a question for @lnslbrty: what if I add (via configure) libcurl support to download (where appropriate) blacklists instead external-download+nDPI load ?

utoni:
You mean downloading the blacklists during nDPI startup/init phase?
Should be possible. But I would still give the user the choice to maintain her or his own blacklists "on disk".
So additional API calls would be fine e.g. with ndpi_download_malicious_sha1_file().

PR #1151, opened 2021-03-09
HTTP: fix memory access in ndpi_http_parse_subprotocol()

IvanNardi:


lucaderi:
Interesting fix: I didn't think it was necessary, thanks

IvanNardi:
> Interesting fix: I didn't think it was necessary, thanks

I bump into this error running ndpiReader against some pcaps; I might try to provide you a short trace triggering the crash

IvanNardi:
> > Interesting fix: I didn't think it was necessary, thanks
> 
> I bump into this error running ndpiReader against some pcaps; I might try to provide you a short trace triggering the crash

It seems to me that most of the pcaps triggering the crash have one thing in common: the http headers spawn multiple packets. Example: the hostname is in one packet but the GET/POST header is in another one. 
No sure If I am getting that right or if this is useful...
[http-9.zip](https://github.com/ntop/nDPI/files/6111063/http-9.zip)

Issue #1148, opened 2021-03-05
protocols supported by nDPI (3.4 stable) but not identified by nDPI sample reader app 


subhajit-cdot:
Hi,

Below are the protocols which are supported by nDPI (3.4 stable) but not identified by nDPI sample app using the attached pcap files corresponding to each protocol listed:

    1. Kontiki
    2. ZATTOO
    3. SHOUTCAST
    4. TVUPLAYER
    5. QQLIVE
    6. AIMINI
    7. WARCRAFT3
    8. CITRIX
    9. PPStream
    10. Hotmail
    11. FASTTRACK
    12. Soulseek
    13. Bittorent (TCP sessions)
    14. UPNP
    15. TOR
    16. Skype Chat/Videocall (some versions)
    17. Git
    18. FIX
    19. QUIC
    20. TFTP
    21. DOFUS
    22. Whatsapp (chat)
    23. CORBA
    24. Telegram
    25. kakaotalk
    26. Wechat
    27. Hangout Duo
    28. Hotspot shield
    29. TRUPHONE
    30. Lotus Notes
    31. RSYNC 

Please let me know if there is any issue in the pcap files or are these application signatures different from the existing dissector? 

Download PCAP files: [https://gofile.io/d/LkB0uw](url)
Thanks & Regards,
Subhajit

IvanNardi:
A quick comment on the QUIC traces.
Packet 4 of client.pcap can't be the first QUIC packet sent from the client. Therefore, or that is not a QUIC session, or the initial handshake is missing, IMO; nDPI needs the initial packets to properly detect QUIC traffic.
Could you double check, please?

utoni:
I've just looked at the aimini traces and I see two issues here.

The first issue is that the aimini protocol classification needs an update here.
The second issue is the nDPI core which stops after detecting the first protocol and does not care about possible subprotocols of the detected protocol (mentioned in #1141). If my understanding of subprotocols is correct; nDPI protocol detection routines should not stop after detecting a protocol which can have subprotocols e.g. HTTP and TLS. At the moment, the subprotocol flag is only used if a protocol classification decides to change the detected protocol which makes IMHO only sense if the protocol classification (in that case: aimini) callback gets called (which is not the case for aimini).

I'm working on a PR which is not done yet (needs some testing).

subhajit-cdot:
@lnslbrty Hi Toni,
[...]

PR #1156, opened 2021-03-20
Refactored nDPI subprotocol handling and aimini protocol detection.

utoni:
This is the outcome of the subprotocol refactoring.

There are still no clear and easy to understand definitions of a protocol, master protocol, subprotocol and app protocol.
IMHO there are only protocols and applications that make use of those.
I would stay away from vague defined terms e.g. a master protocol and a subprotocol.

Here is what I did:

1. Refactored callback buffer handling. That significantly increased nDPI's performance.
2. Refactored subprotocol handling.
3. Refactored aimini protocol detection. The HTTP part could also be moved to the application / content-based matching part in `src/lib/ndpi_content_match.c.inc`. Unsure here.

I've also took care of Python code and updated it according to the new C structures.
Also cleaned up some unused structure members in `ndpi_proto_defaults` and `ndpi_packet_struct`.

The next steps are to write some documentation for the different types of protocols and to investigate and improve the rest of the protocol detection routines mentioned in #1148.


lucaderi:
First of all thanks: this is an interesting code cleanup that paves the way to simpler code and ideally even faster performance.

PR #1159, opened 2021-03-30
Mining: lru cache is ipv4 only (for the time being)

IvanNardi:
Fix memory error with ipv6 traffic.

It is a simple workaround until the cache supports ipv6 keys too


lucaderi:
Good catch

PR #1161, opened 2021-04-05
GTP: fix parsing of GTP headers

IvanNardi:
Message length checks and basic headers are not uniform across GTP-U,
GTP-C and GTP-PRIME.

Note that, even if the length checks were wrong, the GTP sessions were almost
always correctly classified because of the "guessing" algorithm.

This patch has been tested with GTP-U, GTP-C-V1, GTP-C-V2 and GPT-PRIME-V2
traffic using ndpiReader with "-d" flag (to avoid "guessing" algorithm) and
without "-t" flag (to avoid GTP-U de-tunneling).

See #1148

### DIFF
--- a/src/lib/protocols/gtp.c
+++ b/src/lib/protocols/gtp.c
@@ -24,38 +24,87 @@
 
 #include "ndpi_api.h"
 
+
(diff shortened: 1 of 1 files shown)


lucaderi:
Thank you

PR #1175, opened 2021-04-26
Check for common ALPNs and set a flow risk if not known.

utoni:
* Increased risk bitmask to 64bit (instead of 32bit).
 * Removed annoying "Unknown datalink" error message for fuzzers.

### DIFF
--- a/example/reader_util.c
+++ b/example/reader_util.c
@@ -1773,9 +1773,15 @@ struct ndpi_proto ndpi_workflow_process_packet(struct ndpi_workflow * workflow,
     break;
 
   default:
-    /* We shoudn't be here, because we already checked that this datalink is supported.
-       Should ndpi_is_datalink_supported() be updated? */
+#ifndef BUILD_FUZZTARGETS
+    /*
+     * We shoudn't be here, because we already checked that this datalink is supported.
+     * Should ndpi_is_datalink_supported() be updated?
+     *
+     * NOTE (toni): We get here quite often during fuzzing and it spams consoles and logfiles. ;)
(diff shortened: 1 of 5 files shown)


example/reader_util.c
```diff
@@ -1773,9 +1773,15 @@ struct ndpi_proto ndpi_workflow_process_packet(struct ndpi_workflow * workflow,
...
-       Should ndpi_is_datalink_supported() be updated? */
+#ifndef BUILD_FUZZTARGETS
+    /*
+     * We shoudn't be here, because we already checked that this datalink is supported.
+     * Should ndpi_is_datalink_supported() be updated?
+     *
+     * NOTE (toni): We get here quite often during fuzzing and it spams consoles and logfiles. ;)
+     */
     printf("Unknown datalink %d\n", datalink_type);
+#endif
```
IvanNardi, line 1784: Nice catch :)
This check is clearly fine, but I want to suggest another fix.
Since  the datalink type is a global property of the pcap handle, `fuzz_ndpi_reader` could check it, via `ndpi_is_datalink_supported() `, before iterating the packets.
Something like calling `ndpi_is_datalink_supported()` in `fuzz_ndpi_reader::LLVMFuzzerTestOneInput()` before the main `while` loop. In this way, this define shouldn't be necessary.
Note that this is the exact behavior of `ndpi_reader`.
What do you think?

utoni: Your solution is the better one.
Although the fuzzer should should always test as many code path's as possible.
But this one is somehow a special case.

PR #1176, opened 2021-04-27
Allow to override build date with SOURCE_DATE_EPOCH

bmwiedemann:
Allow to override build date with `SOURCE_DATE_EPOCH`
in order to make builds reproducible.
See https://reproducible-builds.org/ for why this is good
and https://reproducible-builds.org/specs/source-date-epoch/
for the definition of this variable.

While this patch works, there are probably better ways to
deterministically generate the `NDPI_API_VERSION` value
by hashing relevant inputs.

This PR was done while working on [reproducible builds for openSUSE](https://en.opensuse.org/openSUSE:Reproducible_Builds).


lucaderi:
Thank you

PR #1177, opened 2021-04-27
Check datalink during fuzzing to prevent console / logfile spam.

utoni:
As adviced by @IvanNardi in #1175 the fuzz reader checks now if the datalink is valid before processing any packets.


lucaderi:
Thank you

Issue #1182, opened 2021-05-10
Certificate name mismatch risk not detected


simonemainardi:
visiting `https://wrong.host.badssl.com/` is meant to test name mismatches.

```
$ curl -vv "https://wrong.host.badssl.com/"
*   Trying 104.154.89.105...
* Connected to wrong.host.badssl.com (104.154.89.105) port 443 (#0)
* found 129 certificates in /etc/ssl/certs/ca-certificates.crt
* found 520 certificates in /etc/ssl/certs
* ALPN, offering http/1.1
* SSL connection using TLS1.2 / ECDHE_RSA_AES_128_GCM_SHA256
* 	 server certificate verification OK
* 	 server certificate status verification SKIPPED
* SSL: certificate subject name (*.badssl.com) does not match target host name 'wrong.host.badssl.com'
* Closing connection 0
curl: (51) SSL: certificate subject name (*.badssl.com) does not match target host name 'wrong.host.badssl.com'
```

![image](https://user-images.githubusercontent.com/3397663/117647818-c47d8e80-b18d-11eb-813d-ebbacd90cbff.png)

However, seems nDPI cannot detect this as certificate mismatch risk. Pcap enclosed.

[name_mismatch.pcap.zip](https://github.com/ntop/nDPI/files/6451455/name_mismatch.pcap.zip)

utoni:
Are wildcards within SSL cert subject names not allowed?

IvanNardi:
> Are wildcards within SSL cert subject names not allowed?

Wildcards are allowed in SSL certificates.
But, wildcard only covers one level of subdomains, AFAIK

See RFC 6125, section 6.4.3

```
   2.  If the wildcard character is the only character of the left-most
       label in the presented identifier, the client SHOULD NOT compare
       against anything but the left-most label of the reference
       identifier (e.g., *.example.com would match foo.example.com but
       not bar.foo.example.com or example.com).
```

Note that I am not sure if RFC 6125 is still the most updated reference for this matter.

Even if the RFC uses the "SHOULD NOT" expression (instead of "MUST NOT"), I am not aware of any browser/client which does not follow this rule.

lucaderi:
Fixed

PR #1183, opened 2021-05-11
Improved SSL certificate name wildcard handling and risk. #1182

utoni:


lucaderi:
Thank you

PR #1194, opened 2021-06-02
wireshark/lua: improve flow risk visualization

IvanNardi:
Create a separated proto field entry for each possible flow risk.
This way, filtering will be more natural: you can use something like
"ndpi.flow_risk.desktop_file_sharing_session"

### DIFF
--- a/wireshark/ndpi.lua
+++ b/wireshark/ndpi.lua
@@ -18,6 +18,10 @@
 -- Inc., 59 Temple Place - Suite 330, Boston, MA 02111-1307, USA.
 --
 
+function bit(p)
+   return 2 ^ p  -- 0-based indexing
+end
+
 
 local ndpi_proto = Proto("ndpi", "nDPI Protocol Interpreter")
 ndpi_proto.fields = {}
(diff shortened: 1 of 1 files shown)


IvanNardi:
There is nothing "wrong" with the current visualization of flow risk information in Wireshark.\
But using two fields (one with the raw bitmask, and one with the concatenation of the string representation of each bit) is not the natural way to visualize this kind of information.

I reworked that, creating a separated protocol field for each risk bit, allowing them to be "first class citizens" in the Wireshark interface.\
The major advantage is that you can filter them using an expression like: "ndpi.flow_risk.desktop_file_sharing_session"

Example:

Screenshot from 2021-05-18 20-19-51

This representation has the drawback of being quite large in the Packet Detail pane.\
Two knobs are provided to allow some (non-standard but) more compact views:

Screenshot from 2021-05-18 20-21-00
Screenshot from 2021-05-18 20-22-03

Please note that my knowledge of Lua (and Wireshark bindings) is quite elementary: there are for sure more elegant and efficient ways to do that.

Just an idea and some options...

lucaderi:
Great stuff @IvanNardi

PR #1198, opened 2021-06-04
Fixed uninitialized memory use.

vel21ripn:
The pointer "header" must be initialized before first call pcap_next_ex().
The problem shows up when using newer versions of the clang compiler.


lucaderi:
Thanks

PR #1204, opened 2021-06-09
Bug in ndpi_lru_find_cache

vel21ripn:
Key value check is missing.


lucaderi:
You're right, collisions can happen

Issue #1195, opened 2021-06-02
QUIC: (out-of-order) fragmented Client Hello 


IvanNardi:
From an historical point of view, splitting the ClientHello (CH) message (of TLS traffic) in multiple, out-of-order, fragments was one of the easiest, but very effective ways, to fool firewall, middleboxes and DPI engines...\
There is the same identical issue with QUIC, right now.

In the vast majority of QUIC traffic, in the first UDP packet sent by the client there is only one Crypto Frame and it contains the entire CH.\
Nonetheless, from a standard POV, it is perfectly fine to split the CH in multiple UDP packets (if the CH is very large, for example) or even in multiple frames in the same UDP payload (to help avoiding ossification of the public part of QUIC, for example).\
Google is doing exactly the latter with the latest Canary Chrome versions, splitting the CH in multiple out-of-order fragments in the same Initial Packet: you can easily detect this kind of traffic.\
See: https://quiche.googlesource.com/quiche/+/cb6b51054274cb2c939264faf34a1776e0a5bab7

Other information:
* https://github.com/quicwg/ops-drafts/issues/358
* https://datatracker.ietf.org/doc/html/draft-ietf-quic-manageability-11

Examples attached
[quic_frags_ch_in_multiple_packets.pcapng.gz](https://github.com/ntop/nDPI/files/6586647/quic_frags_ch_in_multiple_packets.pcapng.gz)
[quic_frags_ch_out_of_order_same_packet_craziness.pcapng.gz](https://github.com/ntop/nDPI/files/6586648/quic_frags_ch_out_of_order_same_packet_craziness.pcapng.gz)

IvanNardi:
Reopening the issue: f8fe3ee only handles in-order and non-overlapping fragments.
The most common case (out-of-order fragments, in the same UDP packet) is not handled yet

IvanNardi:
Current code should correctly handle fragmented Client Hello.
Thanks everyone

PR #1216, opened 2021-06-23
QUIC: add basic support for fragmented Client Hello

IvanNardi:
Only in-order and non overlapping fragments are handled
See #1195


lucaderi:
Thanks

PR #1218, opened 2021-06-27
reconcile PPSTREAM protocol category with content_match declarations

vpiserchia:
This PR tries to reconcile category definition between ```ndpi_main.c``` and ```lib/ndpi_content_match.c.inc``` and remove duplicated definition


vpiserchia:
It is very hard to find the error

lucaderi:
Thanks

PR #1227, opened 2021-07-04
TLS: fix a memory error in JA3 code

IvanNardi:
protocols/tls.c:1856:5: runtime error: index 256 out of bounds for type 'char [256]'
SUMMARY: UndefinedBehaviorSanitizer: undefined-behavior protocols/tls.c:1856:5

### DIFF
--- a/src/lib/protocols/tls.c
+++ b/src/lib/protocols/tls.c
@@ -1849,7 +1849,7 @@ int processClientServerHello(struct ndpi_detection_module_struct *ndpi_struct,
 		       duplicate_found);
 #endif
 
-		if (i == tot_signature_algorithms_len) {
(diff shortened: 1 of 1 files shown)


IvanNardi:
Proper (?) fix for the issue reported in #1222.
@lnslbrty , could you take a look, please?
[fuzz-uSbaCD.pcapng.gz](https://github.com/ntop/nDPI/files/6759876/fuzz-uSbaCD.pcapng.gz)

utoni:
will do ASAP

utoni:
Kudos! There was an error in my logic if `i > tot_signature_algorithms_len` which happens if the loop above ran `tot_signature_algorithms_len` times. Thanks!

PR #1233, opened 2021-07-06
Improved RTSP detection the second. (#1232)

utoni:
* RTSP is no subprotocol of HTTP (most of the time)
 * detection patterns should stay in rtsp.c
 * set detected HTTP protocol only if at least a valid HTTP method detected


lucaderi:
Thank you

PR #1238, opened 2021-07-10
TLS Risks - Certificate Validity Too Long 

pacant:


lucaderi:
This PR needs cleanup, and you need to remove binary files as well non-English words. Please clean it up first then we'll look at it

Issue #1196, opened 2021-06-03
TLS Risks - Certificate Validity Too Long


simonemainardi:
It would be good to add a mildly-scored TLS flow risk ERR_CERT_VALIDITY_TOO_LONG as 

> In a move that's meant to boost security, Apple, Google, and Mozilla are set to reject publicly rooted digital certificates in their respective web browsers that expire more than 13 months

(source: https://thehackernews.com/2020/09/ssl-tls-certificate-validity-398.html)

This can be reproduced visiting `https://wdcp.microsoft.com/`.

----

Moreover, visiting https://wdcp.microsoft.com/ yields errors on browsers and with curl but no risk is detected by nDPI.

Firefox yields an "Unkown Issuer" TLS error.

![image](https://user-images.githubusercontent.com/3397663/120598387-7edb7b00-c446-11eb-9740-8a2fe245a9e9.png)

Safari yields an untrusted CA error

![image](https://user-images.githubusercontent.com/3397663/120598789-fdd0b380-c446-11eb-84e4-0d09155df576.png)

Chrome says invalid CA

<img width="948" alt="image" src="https://user-images.githubusercontent.com/3397663/120599134-661f9500-c447-11eb-959b-7ce8e35b4b58.png">

```
$ curl -vvv "https://wdcp.microsoft.com/"
*   Trying 20.86.161.79...
* Connected to wdcp.microsoft.com (20.86.161.79) port 443 (#0)
* found 129 certificates in /etc/ssl/certs/ca-certificates.crt
* found 520 certificates in /etc/ssl/certs
* ALPN, offering http/1.1
* SSL connection using TLS1.2 / ECDHE_RSA_AES_256_GCM_SHA384
* server certificate verification failed. CAfile: /etc/ssl/certs/ca-certificates.crt CRLfile: none
* Closing connection 0
curl: (60) server certificate verification failed. CAfile: /etc/ssl/certs/ca-certificates.crt CRLfile: none
More details here: http://curl.haxx.se/docs/sslcerts.html

curl performs SSL certificate verification by default, using a "bundle"
 of Certificate Authority (CA) public keys (CA certs). If the default
 bundle file isn't adequate, you can specify an alternate file
 using the --cacert option.
If this HTTPS server uses a certificate signed by a CA represented in
 the bundle, the certificate verification probably failed due to a
 problem with the certificate (it might be expired, or the name might
 not match the domain name in the URL).
If you'd like to turn off curl's verification of the certificate, use
[...]

PR #1239, opened 2021-07-13
TLS Risks - Certificate Validity Too Long

pacant:
Ticket #1196


src/lib/protocols/tls.c
```diff
@@ -470,8 +474,13 @@ static void processCertificateElements(struct ndpi_detection_module_struct *ndpi
 		       flow->protos.tls_quic_stun.tls_quic.notAfter, utcDate);
 #endif
 	      }
+	      notAfterYear = utc.tm_year;
+		  notAfterMon = utc.tm_mon;
 	    }
-
```
lucaderi, line 474: This check is very complex. In the current code you have https://github.com/ntop/nDPI/blob/dev/src/lib/protocols/tls.c#L476
so it is enough to do something like
if((flow->protos.tls_quic_stun.tls_quic.notAfter-flow->protos.tls_quic_stun.tls_quic.notBefore) > threshold)
 ndpi_set_risk(flow, NDPI_TLS_CERT_VALIDITY_TOO_LONG);

Please rework this code

lucaderi (commented):
In the new PR please also add tests/result/tls_certificate_too_long.pcap.out that is used to verify the code

IvanNardi:
Might I suggest an improvement?
Only certificates issued after 01/09/2020 should have a validity not longer than 13 months.
It should be useful to avoid triggering any warnings for certificates issued earlier.
My 2 cents

src/lib/protocols/tls.c
```diff
@@ -454,7 +458,7 @@ static void processCertificateElements(struct ndpi_detection_module_struct *ndpi
 	    for(j=0; j<len; j++) printf("%c", packet->payload[offset+j]);
 	    printf("]\n");
 #endif
-
+
```
utoni, line 459: whitespace hell :)

src/lib/protocols/tls.c
```diff
@@ -470,8 +474,13 @@ static void processCertificateElements(struct ndpi_detection_module_struct *ndpi
 		       flow->protos.tls_quic_stun.tls_quic.notAfter, utcDate);
 #endif
 	      }
+	      notAfterYear = utc.tm_year;
+		  notAfterMon = utc.tm_mon;
```
utoni: confusing indentation

src/lib/protocols/tls.c
```diff
@@ -94,6 +94,8 @@ union ja3_info {
  */
 
 #define NDPI_MAX_TLS_REQUEST_SIZE 10000
+#define TLS_THRESHOLD 34186659 // Threshold for certificate validity
```
IvanNardi, line 97: @pacant: where does this value come from?
It seems wrong to me: max validity is 398 day -> 398 * 86400 = 34387200

PR #1241, opened 2021-07-13
Improved dnscrypt midstream detection and fixed Skype false-positive.

utoni:


lucaderi:
Thank you

PR #1245, opened 2021-07-14
Code cleanup.

vel21ripn:
ndpi_utils.c: use ndpi_malloc,ndpi_calloc,ndpi_free
genshin_impact.c, git.c, hpvirtgrp.c, http.c, z3950.c: removed "#include stdlib.h"
Added packet->iph check before using it. I have a bug report about a
crash in the tinc protocol due to packet->iph == NULL.

### DIFF
--- a/src/lib/ndpi_utils.c
+++ b/src/lib/ndpi_utils.c
@@ -1887,17 +1887,17 @@ u_int32_t ndpi_quick_16_byte_hash(u_int8_t *in_16_bytes_long) {
 /* ******************************************************************** */
 
 ndpi_str_hash* ndpi_hash_alloc(u_int32_t max_num_entries) {
-  ndpi_str_hash *h = (ndpi_str_hash*)malloc(sizeof(ndpi_str_hash));
+  ndpi_str_hash *h = (ndpi_str_hash*)ndpi_malloc(sizeof(ndpi_str_hash));
 
   if(!h) return(NULL);
   if(max_num_entries < 1024) max_num_entries = 1024;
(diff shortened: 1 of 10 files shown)


IvanNardi:
Are these `packet->iph` checks really necessary?
All the protocols you changed are using `NDPI_SELECTION_BITMASK_PROTOCOL_TCP_OR_UDP*` or similar, so they should be triggered only with ipv4 traffic, i.e. with a valid `packet->iph` pointer, IMO.
In the tinc case, for example, before accessing `packet->iph` we just check for `packet->tcp`: I think that if we have a valid TCP header but not an IPv4 header, the real bug is somewhere else.

The changes might be useful in case we handle IPv6 case, tough.

Am I missing something?

vel21ripn:
**Are these packet->iph checks really necessary?**
There was a problem with the TINC protocol. I have kernel crash. I have no traffic sample on which this error appears. At that time I could not find the source of the error.
Most likely this because of IPv6.
Is there a way to allow the application protocol to process only IPv4 protocol? 
Then we can remove the checks packet->iph.

IvanNardi:
AFAIK, TINC already processes only ipv4 traffic (because of `NDPI_SELECTION_BITMASK_PROTOCOL_TCP_OR_UDP`)

vel21ripn:
There is a problem with packet-> iph == NULL and packet-> tcp != NULL.
I just got a kernel panic in ipv4 code after enabling ipv6 support in ndpi-netfilter!
I will try to figure out the reason for this situation. Perhaps the problem is in the ndpi-netfilter code.
I will close this PR.

PR #1250, opened 2021-07-16
Bug fix.

vel21ripn:
Restore part of the fields flow->packet after cleaning flow structure.

### DIFF
--- a/src/lib/ndpi_main.c
+++ b/src/lib/ndpi_main.c
@@ -4245,26 +4245,63 @@ static int ndpi_init_packet_header(struct ndpi_detection_module_struct *ndpi_str
       if(flow->packet.tcp->syn != 0 && flow->packet.tcp->ack == 0 && flow->init_finished != 0 &&
 	 flow->detected_protocol_stack[0] == NDPI_PROTOCOL_UNKNOWN) {
 
-        u_int8_t backup;
-	u_int16_t backup1, backup2;
+	u_int16_t guessed_protocol_id, guessed_host_protocol_id;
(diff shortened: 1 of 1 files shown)


src/lib/ndpi_main.c
```diff
@@ -4249,11 +4249,35 @@ static int ndpi_init_packet_header(struct ndpi_detection_module_struct *ndpi_str
...
+	u_int64_t current_time_ms;
+	u_int16_t l3_packet_len;
+	u_int16_t l4_packet_len;
+	u_int16_t payload_packet_len;
+	u_int16_t actual_payload_len;
+	u_int8_t l4_protocol;
+
+#define packet_save(a) a = flow->packet.a
+#define packet_restore(a) flow->packet.a = a
+
```
IvanNardi: Is it feasible to use these macros for the fields we already restore, too? (guessed_protocol_id, num_processed_pkts....)
This way the code should be more uniform

vel21ripn:
It is quick and dirty. This is one of the possible options for correcting the error.
If you can offer a more elegant option, I close my PR.

IvanNardi:
> It is quick and dirty. This is one of the possible options for correcting the error.
> If you can offer a more elegant option, I close my PR.

Sorry, I have not been clear.
I like your solution! I just wonder if you can extend it to all the fields we restore (instead of using these macros only for the "new" fields and the old code for the "old"fields).
Anyway, your PR is ok, IMO!

vel21ripn:
New, more uniform version.

PR #1252, opened 2021-07-19
Added risk: TLS_EXTENSION_SUSPICIOUS

utoni:
In a malware-world where the bad guys mimic TLS extensions to transmit data to C&C infrastructure, we need to be aware of extensions and their content.

As you can see in the code, I've also included an entropy calculation but the code is currently not used somewhere.
I need to do some investigation if and where it makes sense and do some performance tests (maybe with libbenchmark ;).
But I think this function might also be coming handy in other parts of nDPI. 


src/lib/protocols/tls.c
```diff
@@ -1165,6 +1152,76 @@ static void ndpi_int_tls_add_connection(struct ndpi_detection_module_struct *ndp
...
+#ifdef DEBUG_TLS
+      printf("[TLS] suspicious extension id: %u\n", extension_id);
+#endif
+      ndpi_set_risk(flow, NDPI_TLS_EXTENSION_SUSPICIOUS);
+      return;
+    }
+  }
+
+  /* Check for DTLS-only extensions. */
+  if (packet->udp == NULL)
```
IvanNardi: It seems to me these two extensions shouldn't be used by QUIC.
Should this check be modified to something like `if (!is_dtls)` or similar (see at the beginning of `processClientServerHello()`...)?

utoni: Yeah, makes sense. AFAIK QUIC does not allow extension 53 and 54. Connection IDs are part of the QUIC protocol and not of the TLS handshake.

PR #1263, opened 2021-07-27
Improved libm check via Autoconf.

utoni:
* libm will now be part of @ADDITIONAL_LIBS@ (if required)

Not sure if the Autoconf macro `LT_LIB_M` will work for Darwin as well.
This fixes linking issues with pkg-config.

### DIFF
--- a/configure.seed
+++ b/configure.seed
@@ -32,6 +32,7 @@ AS_IF([test "${with_sanitizer+set}" = set],[
 ])
 
 LT_INIT
+LT_LIB_M
 
 SYSTEM=`uname -s`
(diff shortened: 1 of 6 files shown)


lucaderi:
> * libm will now be part of @ADDITIONAL_LIBS@ (if required)
> 
> Not sure if the Autoconf macro `LT_LIB_M` will work for Darwin as well.
> This fixes linking issues with pkg-config.
> 
> Signed-off-by: Toni Uhlig [matzeton@googlemail.com](mailto:matzeton@googlemail.com)

Seems to work on Darwin. Good work.

PR #1270, opened 2021-08-05
Added entropy calculation to check for suspicious (encrypted) payload.

utoni:
Currently available for ICMPv4 payload only. I'm planning to extend that feature for DNS and HTTP as well.

(this may be used to detect encrypted tunnels, but some linux hosts may also reply with ICMPs that contain the payload of a packet that could not delivered which may also be encrypted in case of QUIC)


lucaderi:
I will fix the remaining items. but please make sure whenever you do https://github.com/ntop/nDPI/blob/dev/src/include/ndpi_typedefs.h#L64

utoni:
Totally forgot about those items. Sorry.

PR #1273, opened 2021-08-09
Fixed Mingw64 build, SonerCloud-CI and more.

utoni:


lucaderi:
@lnslbrty Why do we need all windows/WpdPack/ files?

utoni:
> @lnslbrty Why do we need all windows/WpdPack/ files?

Probably not all files are required. For now I dunno which ones are superfluous besides Packet.lib
But for a succesful Mingw compilation, at least some header files and the import library wpcap.lib are required.

utoni:
I tried to remove as much as possible from `windows/WpdPack`. I've also removed the broken VisualStudio integration. Some serious issues, mostly path related. If there is a need for VS integration, I might be able to fix that (requires some time).

I did not test any configure/make with Darwin or *BSD.

lucaderi:
It works on Darwin, thank you.

PR #1275, opened 2021-08-17
wireshark/lua: restore full flow risks dissection

IvanNardi:
It was partially disabled in 3eba8cc5.
Wireshark and Wireshark-Lua bindings don't handle 64 bit integer very well.
(see https://www.wireshark.org/docs/wsdg_html_chunked/lua_module_Int64.html).

As workaround, only for visualization purpose, split the (64 bit) risk mask
into two 32 bit integer values.


IvanNardi:
The new visualization is something like that:
Screenshot from 2021-08-17 20-03-55
It is not perfect, but it should be good enough

lucaderi:
Thank you

PR #1276, opened 2021-08-18
Compile everything with "-W -Wall -Wno-unused-parameter" flags

IvanNardi:
Fix all the warnings.

Getting rid of "-Wno-unused-parameter" is quite complex because some
parameters usage depends on compilation variable (i.e.
`--enable-debug-messages`).

The "-Werror" flag has been added only in Travis builds to avoid
breaking the builds to users using uncommon/untested
OS/compiler/environment.

Tested on:
* x86_64; Ubuntu 20.04; gcc 7,8,9,10,11; clang 7,8,9,10,11,12
* x86_64; CentOS 7.7; gcc 4.8.5 (with "--disable-gcrypt" flag)
* Raspberry 4; Debian 10.10; gcc 8.3.0


IvanNardi:
I really like the "-Werror" option and I do think it is very useful.
Nonetheless, if you prefer otherwise or if its support will be too cumbersome, you can easily revert the travis part, now or in the future.

IvanNardi:
The idea was to set "-Werror" (only) during CI, but something is probably wrong....
Are Travis tests still triggered?

utoni:
Definitly a good idea. I aggree with -Werror in the CI as it might help to find potential dangerous errors before a Pull Request is merged.

// greetings from Istanbul

Issue #1282, opened 2019-05-17
Implement additional DNS Alerts


lucaderi:
DNS traffic is very important on networks as it one of the core Internet protocols. ntop (flow and host) alerts should be enriched with additional DNS alerts such as
- fragmented DNS traffic
- large DNS packets (pay attention to OARC support https://www.dns-oarc.net/oarc/services/replysizetest)

PR #1290, opened 2021-09-07
Progetto esame Gestione di Reti - Debora Cerretini

deboracerretini:


utoni:
I think `NDPI_DNS_DIMENSION_ALERT` is somehow confusing, because IMHO it seems more related to a size or length of a buffer then a dimension which I would interpret as something related to multi-dimensional arrays or matrices. But this is just a cosmetic thing.

I am more concerned about the `NDPI_DNS_FRAGMENTED`, which AFAIS is related to IP fragmentation rather then DNS. I do not know what could be the benefit of such a risk. I mean you can basicially fragment every IP packet with every underlying L3/L4 protocol.

But probably I do not get the full picture here.

Also an @ntop maintainer required to approve the PR so the CI can do some checks.

src/lib/protocols/dns.c
```diff
@@ -513,6 +517,26 @@ static void ndpi_search_dns(struct ndpi_detection_module_struct *ndpi_struct, st
 
   if(flow->packet_counter > 3)
     NDPI_EXCLUDE_PROTO(ndpi_struct, flow);
+    
+    
+  if(flow->packet.udp != NULL && flow->packet.payload_packet_len > PKT_LEN_ALERT)
```
IvanNardi: 1) Are you sure this limit applies to  LLMNR and MDNS, too (other than "standard" DNS)?
2) I don't know if it is in scope with your project, but it is more complex than that even for "standard" DNS: RFC 6891 allows DNS packets to be bigger than 512 bytes...

deboracerretini:
> I think `NDPI_DNS_DIMENSION_ALERT` is somehow confusing, because IMHO it seems more related to a size or length of a buffer then a dimension which I would interpret as something related to multi-dimensional arrays or matrices. But this is just a cosmetic thing.
> 
> I am more concerned about the `NDPI_DNS_FRAGMENTED`, which AFAIS is related to IP fragmentation rather then DNS. I do not know what could be the benefit of such a risk. I mean you can basicially fragment every IP packet with every underlying L3/L4 protocol.
> 
> But probably I do not get the full picture here.
> 
> Also an @ntop maintainer required to approve the PR so the CI can do some checks.

Ciao, grazie per il tuo feedback!
Questa pull request è un progetto che ho fatto per il corso Gestione di Reti con il professor Luca Deri, quindi i due flow risk sono stati implementati con il suo accordo. Ho iniziato con queste informazioni che indicano i problemi che possono derivare da queste condizioni non verificate: https://github.com/ntop/nDPI/issues/1282

Per quanto rigarda nello specifico `NDPI_DNS_DIMENSION_ALERT` è ovviamente un nome che ho creato io per il flow risk, certamente posso modificarlo per una comprensione migliore. Potrebbe andare meglio 'NDPI_DNS_LARGE_PACKET' ? Sono aperta ad ulteriori suggerimenti.

Grazie,
Debora

src/lib/protocols/dns.c
deboracerretini: Ciao, grazie per il tuo feedback!
Questa pull request è un progetto che ho fatto per il corso Gestione di Reti con il professor Luca Deri, quindi i due flow risk sono stati implementati con il suo accordo. Ho iniziato con queste informazioni che indicano i problemi che possono derivare da queste condizioni non verificate: #1282 Penso che questo risponda al punto n. 2.

Per il punto n. 1 invece sono d'accordo che potrebbe essermi sfuggito qualcosa nell'implementazione, mi informo su quanto mi hai indicato. Se avessi dei suggerimenti/riferimenti su cui posso partire andrebbe benissimo.

Grazie,
Debora

utoni:
I love the italian language. But I do not fully understand your reply.
Still, it helps to get the bigger picture.

Grazie,
Toni

src/lib/protocols/dns.c
IvanNardi: 1) rfc4795 sec 2.1 and rfc6762 sec 17
2) I was right to think it was out-of-scope for your project... But I think that you should, at least, add a comment about EDNS and its future implementation to avoid false positives

lucaderi:
@deboracerretini Can you please reply in English and address the comments reported?

deboracerretini:
> I love the italian language. But I do not fully understand your reply.
> Still, it helps to get the bigger picture.
> 
> Grazie,
> Toni

@lnslbrty lnslbrty I am very sorry, I didn't even notice I readed in English and replied in Italian!

What I meant is that this pull requests are implemented as a project for a university class with professor Luca Deri, and the specifics are here: #1282.

About 'NDPI_DNS_DIMENSION_ALERT', obviusly it's a descriprion I created, so if it confusing I can change it. Could it be better 'NDPI_DNS_LARGE_PACKET' ? I am open to further suggestions.

Thank you,
Debora

utoni:
`NDPI_DNS_LARGE_PACKET` sounds good.
But what about @IvanNardi comments about the packet length of DNS. Where does this magic number came from?

I still do not understand for what a *DNS fragmented packet* risk could be useful. I mean wouldn't it better to introduce a "generalized" *IP packet fragmented* risk

deboracerretini:
@lnslbrty reading in this page: https://www.dns-oarc.net/oarc/services/replysizetest, I came up with 512 bytes as a good measurement for a large DNS packet.

About the DNS fragmented packet risk, I readed here: https://blog.powerdns.com/2018/09/10/spoofing-dns-with-fragments/. Basically it says that DNS responses can be used for cache poisoning, and use fragmented DNS responses to acquire certificates for domain names whose nameservers you do not control. As I said, for my project it was specified the DNS protocol, so I implemented specifically for DNS, but I agree it could be implemented in a more generic way.

Debora

lucaderi:
please add test pcaps in your PR as well make sure that
- cd tests
- ./do.sh
does not return any error.

lucaderi:
> `NDPI_DNS_LARGE_PACKET` sounds good.
> But what about @IvanNardi comments about the packet length of DNS. Where does this magic number came from?
> 
> I still do not understand for what a _DNS fragmented packet_ risk could be useful. I mean wouldn't it better to introduce a "generalized" _IP packet fragmented_ risk

In DNS packets cannot be fragmented (I mean from the DNS standpoint), in IP this is possible,. If you have a large DNS request/response you need to resubmit it via TCP

utoni:
Got it. Thank you both.

IvanNardi:
> But what about @IvanNardi comments about the packet length of DNS. Where does this magic number came from?

It is not an arbitrary value: the original DNS RFC883  mandates the 512 bytes limit (for UDP).
If you can look at some DNS traffic you will notice that packet size is still nowadays lower than that value (usually).

However RFC6891 allows DNS UDP packets to be larger (via a protocol extension), so things are a little more complex than a trivial check on packet length... Nonetheless, this patch is a good first step in the right direction, IMO

utoni:
There is probably a bug somewhere. `steam_datagram_relay_ping.pcapng` get's a DNS related risk but I am pretty sure that it does not contain any DNS packets.

deboracerretini:
@lnslbrty I hope this changes fixed the bug. I also changed the name of the first flow risk, NDPI_DNS_LARGE_PACKET.

@lucaderi I updated all the files in tests/result, I hope that the outputs now are ok.

wireshark/ndpi.lua
```diff
@@ -74,6 +74,10 @@ flow_risks[32] = ProtoField.bool("ndpi.flow_risk.cert_validity_too_long", "TLS c
 flow_risks[33] = ProtoField.bool("ndpi.flow_risk.suspicious_extension", "TLS suspicious extension", num_bits_flow_risks, nil, bit(1), "nDPI Flow Risk: TLS suspicious extension")
 flow_risks[34] = ProtoField.bool("ndpi.flow_risk.fatal_alert", "TLS fatal alert detected", num_bits_flow_risks, nil, bit(2), "nDPI Flow Risk: TLS fatal alert")
 flow_risks[35] = ProtoField.bool("ndpi.flow_risk.suspicious_entropy", "Suspicious entropy", num_bits_flow_risks, nil, bit(3), "nDPI Flow Risk: suspicious entropy")
+
+flow_risks[36] = ProtoField.bool("ndpi.flow_risk.dns_large_packet", "DNS large packet", num_bits_flow_risks, nil, bit(0), "nDPI Flow Risk: DNS packet is larger than 512 bytes")
```
IvanNardi: bit(4)

wireshark/ndpi.lua
```diff
@@ -74,6 +74,10 @@ flow_risks[32] = ProtoField.bool("ndpi.flow_risk.cert_validity_too_long", "TLS c
...
+flow_risks[37] = ProtoField.bool("ndpi.flow_risk.dns_fragmented", "DNS fragmented", num_bits_flow_risks, nil, bit(0), "nDPI Flow Risk: DNS message is fragmented")
```
IvanNardi: bit(5)

wireshark/ndpi.lua
```diff
@@ -74,6 +74,10 @@ flow_risks[32] = ProtoField.bool("ndpi.flow_risk.cert_validity_too_long", "TLS c
...
+
 -- Last one: keep in sync the bitmask when adding new risks!!
 flow_risks[64] = ProtoField.new("Unused", "ndpi.flow_risk.unused", ftypes.UINT32, nil, base.HEX, bit(32) - bit(4))
```
IvanNardi: bit(32) - bit(6)

Very sorry I didn't notice these errors earlier

deboracerretini:
@IvanNardi updated file wireshark/ndpi.lua.

src/lib/protocols/dns.c
```diff
@@ -513,8 +517,30 @@ static void ndpi_search_dns(struct ndpi_detection_module_struct *ndpi_struct, st
 
   if(flow->packet_counter > 3)
     NDPI_EXCLUDE_PROTO(ndpi_struct, flow);
+    
+  if((flow->packet.detected_protocol_stack[0] == NDPI_PROTOCOL_DNS)
+	 || (flow->packet.detected_protocol_stack[1] == NDPI_PROTOCOL_DNS)) {
+	   
+  	if(flow->packet.udp != NULL && flow->packet.payload_packet_len > PKT_LEN_ALERT)
```
IvanNardi, line 524: Could you add a comment about EDNS above this check, please? Something like "TODO: add supporto to RFC6891 to avoid some false positives"

src/lib/protocols/dns.c
```diff
@@ -513,8 +517,30 @@ static void ndpi_search_dns(struct ndpi_detection_module_struct *ndpi_struct, st
...
+  	const u_int8_t *l3 = (const u_int8_t *) flow->packet.iph;
+  	const struct ndpi_ipv6hdr *iph_v6 = NULL;
+  	const u_int16_t ipsize = flow->packet.l3_packet_len;
+  	  
+  	if(iph != NULL && iph->version == 6 && ipsize >= sizeof(struct ndpi_ipv6hdr)) {
+  	  iph_v6 = (const struct ndpi_ipv6hdr *) l3;
+  	  iph = NULL;
+  	}
+  	  
+  	if((iph != NULL && (ipsize < iph->ihl * 4 || ipsize < ntohs(iph->tot_len) || ntohs(iph->tot_len) < iph->ihl * 4
```
IvanNardi, line 538: Could you add another comment here about ndpiReader?
Something like "Note that ndpiReader drops any IP-fragmented packets, so this check will be always false when using it"

deboracerretini: I'm editing the dns.c file like you suggested, but I don't understand this review.
Are you saying that the whole 'if' (from 537) always returns false? In this case, there must be something off because using the pcap file downloded here: https://weberblog.net/dns-capture-udp-tcp-ip-fragmentation-edns-ecs-cookie/ I see the flow risk about fragmented DNS packets.

IvanNardi: You are right, sorry for the noise. 
ndpiReader drops only ip-fragmented packets with ip-offset != 0, i.e. only the first fragment is passed to the nDPI code.
Ignore this comment; please add the other one about EDNS, please

PR #1294, opened 2021-09-10
Cassandra: fix compilation when "--enable-debug-messages" option is used

IvanNardi:
Let's try adding a dedicated compilation in GitHub Actions to easily
detect this kind of errors in the future

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -23,6 +23,10 @@ jobs:
         run: ./tests/do.sh
       - name: Test UNIT
         run: ./tests/unit/unit
+      - name: Configure (with debug logs)
+        run: make distclean && env CFLAGS='-Werror' ./autogen.sh --enable-debug-messages
+      - name: Build
+        run: make all
       - name: Configure ARM
         run: make distclean && env CFLAGS='-Werror' ./autogen.sh --host=arm-linux-gnueabihf --with-only-libndpi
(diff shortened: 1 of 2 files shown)


.github/workflows/build.yml
```diff
@@ -23,6 +23,12 @@ jobs:
         run: ./tests/do.sh
       - name: Test UNIT
         run: ./tests/unit/unit
+      - name: Configure (with debug logs)
+        run: make distclean && env CFLAGS='-Werror' ./autogen.sh --enable-debug-messages
+      - name: Build
+        run: make all
+      - name: Test DIFF
+        run: ./tests/do.sh
```
utoni: Is it necessary to do the diff tests twice?

utoni: Save ressources whenever possible. :)

IvanNardi: No strong opinion on that... I think that if we compile some more code, we should somehow check that it doesn't crash anything, at very least. But run the entire test suite might be a bit overwhelming...
Do you prefer to remove the tests anyway?

utoni: I mean it is just a "print some pretty debug messages"-mode. There shouldn't be lot's of additional code paths to check. And `./tests/do.sh` is always sooo slow.

IvanNardi: Ok!

PR #1297, opened 2021-09-11
Update the list of default ports for QUIC protocol

IvanNardi:
There are no reasons to register UDP/80 as a default port for QUIC


utoni:
Make sense as the "propagated" port for QUIC is UDP/443 only AFAIK.

PR #1299, opened 2021-09-11
Hangout: fix usage of STUN function

IvanNardi:


IvanNardi:
It is incredible that this error didn't trigger any errors so far!

utoni:
> It is incredible that this error didn't trigger any errors so far!

Google Hangout is in it's transition phase to Google Chat. It is probably not much used in these days. But this is just dangerous half-knowledge. 🦄

src/lib/protocols/hangout.c
```diff
@@ -84,12 +84,13 @@ static u_int8_t is_google_flow(struct ndpi_detection_module_struct *ndpi_struct,
 void ndpi_search_hangout(struct ndpi_detection_module_struct *ndpi_struct,
 			 struct ndpi_flow_struct *flow) {
   struct ndpi_packet_struct * packet = &flow->packet;
+  int matched_src = 0;
```
utoni: Scope of `matched_src` can be reduced if I am not wrong. (reading source code from a smartphone is sometimes confusing)

IvanNardi: Done in the last push. Thanks

PR #1300, opened 2021-09-11
TLS: avoid zeroing large structures

IvanNardi:
Zeroing large structures (i.e. size > KB) is quite costly (from a CPU point
of view): we can safely avoid doing that for a couple of big structures.

Standard and Valgrind tests have been diverging quite a lot: it is time
to re-sync them. Use the same script and enable Valgrind via an
environment variable:

  NDPI_TESTS_VALGRIND=1 ./tests/do.sh


lucaderi:
I think this PR is interesting. However, as you pointed out the real solution is to reduce memory usage and thus less data to zero out

PR #1289, opened 2021-09-02
Add libgcrypt as part of macOS CI.

aouinizied:
For consistency reasons, macOS CI should integrate libgcrypt. (I added also json-c which I think is present by default)

Zied

PR #1303, opened 2021-09-13
Improve CI

aouinizied:
This PR improves the current CI pipeline:
* Add macOS platform (both 11 and 10.15).
* Add Ubuntu 18.04 target.
* Add libgcrypt target (with and without on all targets).
* Keep building for ARM and mingw only on ubuntu latest target.

This PR addresses previously discussion in: https://github.com/ntop/nDPI/pull/1289

Zied


aouinizied:
@lnslbrty note that my changes do not affect other builds etc.
Here are some ideas where feedback is welcome:
* I can move fuzzing to this CI script too and then make all in one CI.
* I can set some GCC and CLANG versions as part of the matrix (maybe only on ubuntu target?).
* Should we definitively drop Travis?
* How about building on a Windows target too? (Note that I'm not an expert on building stuff on Windows and that if it looks ok for you I will need some help)

IvanNardi:
@aouinizied : I am not sure what's going on, but CI didn't run on your PR... 
Am I missing something? Is that expected?

aouinizied:
@IvanNardi I think this is normal as it's a PR changing the CI file. 
Already tested it using an nDPI like repo here https://github.com/nfstream/nDPI-actions/actions 
I already made the changes (remove nDPI repo cloning etc)  before opening the PR.

.github/workflows/build.yml
```diff
@@ -2,36 +2,56 @@ name: Build
 on:
   push:
     branches:
-      - master
+    - master
```
IvanNardi: I don't know the CI syntax, but nDPI doesn't have neither "master" nor "main" branch (but "dev"). Is that correct?

aouinizied: Changed to dev and now CI is triggered.

utoni:
> @lnslbrty note that my changes do not affect other builds etc.
> Here are some ideas where feedback is welcome:
> 
>     * I can move fuzzing to this CI script too and then make all in one CI.
> 
>     * I can set some GCC and CLANG versions as part of the matrix (maybe only on ubuntu target?).
> 
>     * Should we definitively drop Travis?
> 
>     * How about building on a Windows target too? (Note that I'm not an expert on building stuff on Windows and that if it looks ok for you I will need some help)

1. Unsure if we should move fuzzing to the CI script. Are there any benefits?
2. It would be probably worth to build nDPI at least twice with the lowest and highest supported CC.
3. Yeah, I would like to see the nDPI repository without any leftover files. We still have a GIT history. And with a meaningful commit message, we can always revert that commit.
4. I would advise to build at least with Mingw-w64. Building with Visual Studio won't probably as easy and may cause more maintaining work.

utoni:
I do not see any arm or mingw builds. What happened?

// EDIT: The first two jobs build for mingw.

Is it possible to rename jobs? instead of `ubuntu-* DISABLE_GCRYPT` something like `ubuntu-* DISABLE_GCRYPT arm mingw`?

.github/workflows/build.yml
```diff
@@ -2,36 +2,56 @@ name: Build
...
-      - name: Build Mingw-w64
-        run: make all
+    - uses: actions/checkout@v2
+      with:
+        fetch-depth: 0  # Shallow clones should be disabled for a better relevancy of analysis
+    - name: Install Ubuntu Prerequisites
+      if: startsWith(matrix.os, 'ubuntu')
+      run: |
+        sudo apt-get update
+        sudo apt-get install autoconf automake libtool pkg-config gettext libjson-c-dev flex bison libpcap-dev libgcrypt20-dev
```
utoni: Is it possible to install `libgcrypt20-dev` **only** if `DISABLE_GCRYPT=y`?

.github/workflows/build.yml
```diff
@@ -2,36 +2,56 @@ name: Build
...
+        sudo apt-get install gcc-arm-linux-gnueabihf gcc-mingw-w64 libc6-dev
+    - name: Installing MacOS prerequisites
+      if: startsWith(matrix.os, 'macOS')
+      run: |
+        brew install autoconf automake libtool pkg-config gettext json-c libgcrypt
```
utoni: same here

aouinizied: Nice catch (additionally, I was using the wrong flag)! I fixed it and now you can see in tests logs pcaps file (using libgcrypt) skipped. Thanks!

aouinizied:
> I do not see any arm or mingw builds. What happened?
> 
> // EDIT: The first two jobs build for mingw.
> 
> Is it possible to rename jobs? instead of `ubuntu-* DISABLE_GCRYPT` something like `ubuntu-* DISABLE_GCRYPT arm mingw`?

Name generation will be tricky. However I changed the script. Now, cross compilation is triggered for all ubuntu jobs and I added a notice to the step name displayed to make it more clear.

EDIT: Now reverted to ubuntu-latest only as compilation with mingw fails on ubuntu 18.04 with 
**/usr/bin/x86_64-w64-mingw32-ld: cannot find -lucrtbase**

.github/workflows/build.yml
utoni: Oh, I just recognized my typ0. Double negation can be confusing. :confused: 

s/`DISABLE_GCRYPT=y`/`DISABLE_GCRYPT=n`/

aouinizied:
@lnslbrty Travis dropped and badge icon replaced.

utoni:
> > I do not see any arm or mingw builds. What happened?
> > // EDIT: The first two jobs build for mingw.
> > Is it possible to rename jobs? instead of `ubuntu-* DISABLE_GCRYPT` something like `ubuntu-* DISABLE_GCRYPT arm mingw`?
> 
> Name generation will be tricky. However I changed the script. Now, cross compilation is triggered for all ubuntu jobs and I added a notice to the step name displayed to make it more clear.
> 
> EDIT: Now reverted to ubuntu-latest only as compilation with mingw fails on ubuntu 18.04 with
> **/usr/bin/x86_64-w64-mingw32-ld: cannot find -lucrtbase**

Explicit linkage of `-lucrtbase` was added by yourself in 812a3e7099ac644a5706bcc9dc52cfece6e48b1e. But I do not know why it should be necessary. As far as I can remember it is used for development/debugging purposes. Maybe you can try if it works for both by just removing `-lucrtbase` in `configure.seed`.

aouinizied:
> > > I do not see any arm or mingw builds. What happened?
> > > // EDIT: The first two jobs build for mingw.
> > > Is it possible to rename jobs? instead of `ubuntu-* DISABLE_GCRYPT` something like `ubuntu-* DISABLE_GCRYPT arm mingw`?
> > 
> > 
> > Name generation will be tricky. However I changed the script. Now, cross compilation is triggered for all ubuntu jobs and I added a notice to the step name displayed to make it more clear.
> > EDIT: Now reverted to ubuntu-latest only as compilation with mingw fails on ubuntu 18.04 with
> > **/usr/bin/x86_64-w64-mingw32-ld: cannot find -lucrtbase**
> 
> Explicit linkage of `-lucrtbase` was added by yourself in [812a3e7](https://github.com/ntop/nDPI/commit/812a3e7099ac644a5706bcc9dc52cfece6e48b1e). But I do not know why it should be necessary. As far as I can remember it is used for development/debugging purposes. Maybe you can try if it works for both by just removing `-lucrtbase` in `configure.seed`.

@lnslbrty it was probably a dummy fix for some windows CI I was trying to setup. Removed it and now it works on both. Thanks for your help.

.github/workflows/build.yml
```diff
@@ -2,36 +2,63 @@ name: Build
...
-      - name: Configure Mingw-w64
-        run: make distclean && ./autogen.sh --host=x86_64-w64-mingw32
-      - name: Build Mingw-w64
-        run: make all
+    - uses: actions/checkout@v2
+      with:
+        fetch-depth: 0  # Shallow clones should be disabled for a better relevancy of analysis
+    - name: Install Ubuntu Prerequisites
+      if: startsWith(matrix.os, 'ubuntu')
+      run: |
```
utoni: One last question: Is it possible to switch from multi-line statements to single-line statements?

Because with that it should be possible to see exactly which command failed.

(GitLab does support such superfluous things) :grinning:

aouinizied: @lnslbrty Done!

IvanNardi:
Is it possible to test multiple compilers, as suggested earlier?
No strong opinions but something like gcc-11, gcc-7, clang-12 and clang-7 on ubuntu-latest only... not sure about the versions: the idea is to test the newest compilers and something "old" (but easily available on the target)
In the old Travis configuration there was something like that.
Just an idea

utoni:
> Is it possible to test multiple compilers, as suggested earlier?
> No strong opinions but something like gcc-11, gcc-7, clang-12 and clang-7 on ubuntu-latest only... not sure about the versions: the idea is to test the newest compilers and something "old" (but easily available on the target)
> In the old Travis configuration there was something like that.
> Just an idea

I would like to see `libnDPI` compiling/testing on a bigendian architecture e.g. armhf.
The Debian project reported some issues on bigendian machines.

While compiling for `armhf -mbig-endian` is pretty easy, testing might be not as trivial. So this is just a "nice-to-have" (may be also part of a follow-up PR).

aouinizied:
@IvanNardi Yes, need to set it up. But we must implement at least 2 versions on each target for 2 compilers. I will rather make it part of the matrix. It will result in 24 jobs (8 current jobs * 4 versions (2 of CLANG and 2 of GCC let's say)) but this is not an issue as GitHub actions are limited to 255 jobs per matrix (AFAIK).

@lnslbrty We will have mainly two options if you want to build and test on arm using Github Actions:

* Have an RPI somewhere (do not need a static IP or whatever), install on it Github actions, and declare it within the ntop organization as a self-hosted runner. I have some RPIs left somewhere and I was thinking about such an option for nfstream project (to publish arm wheels). But as always, need to stop procrastinating and give it a try soon.
Examples: https://dev.to/l04db4l4nc3r/self-hosted-github-actions-using-raspberry-pi-212m 
* The second one will be to use QEMU to emulate such architectures and this can be set up on ubuntu-latest target for example.
Idea from: https://github.community/t/testing-against-multiple-architectures/17111/9

Zied

utoni:
> @IvanNardi Yes, need to set it up. But we must implement at least 2 versions on each target for 2 compilers. I will rather make it part of the matrix. It will result in 24 jobs (8 current jobs * 4 versions (2 of CLANG and 2 of GCC let's say)) but this is not an issue as GitHub actions are limited to 255 jobs per matrix (AFAIK).
> 
> @lnslbrty We will have mainly two options if you want to build and test on arm using Github Actions:
> 
>     * Have an RPI somewhere (do not need a static IP or whatever), install on it Github actions, and declare it within the ntop organization as a self-hosted runner. I have some RPIs left somewhere and I was thinking about such an option for nfstream project (to publish arm wheels). But as always, need to stop procrastinating and give it a try soon.
>       Examples: https://dev.to/l04db4l4nc3r/self-hosted-github-actions-using-raspberry-pi-212m
> 
>     * The second one will be to use QEMU to emulate such architectures and this can be set up on ubuntu-latest target for example.
>       Idea from: https://github.community/t/testing-against-multiple-architectures/17111/9
> 
> 
> Zied

While the first option sounds interesting and I will probably try it out for my own projects, for `libnDPI` I would stick to a "all-in-one" solution that does not depend on self-hosted machines. I am unsure if the performance drop for the second option is huge. My experience with `qemu-arm` is that it is somewhat slow compared to VM's with a hypervisor available. But I would still prefer option #2.

utoni:
@aouinizied There are also some (not serious) `Code Inspector` complaints. It is all about indentation and unnecessary whitespaces, but should be easy to fix.

aouinizied:
@IvanNardi @lnslbrty I added jobs on Ubuntu Latest using the following old and new versions logic:
* GCC: 7 and 10
* CLANG: 7 and 12

We continue to build on ubuntu-latest, ubuntu 18.04, macos11 and macos10.15 using the default compiler of the runner (gcc for ubuntu and clang for MacOS)

I fixed the CodeInspector complaints too.

As now we have improved our CI on Github Actions, Travis is definitively dropped.

ARM target using Qemu and build on Windows server using MinGW will be part of a follow PR as I need to investigate several issues. Consequently, this PR can be merged as it is.


Zied

.github/workflows/build.yml
```diff
@@ -2,73 +2,102 @@ name: Build
...
+          sudo apt-get install gcc-arm-linux-gnueabihf gcc-mingw-w64 libc6-dev
+      - name: Install Ubuntu Prerequisites (libgcrypt)
+        if: startsWith(matrix.os, 'ubuntu') && !startsWith(matrix.gcrypt, '--disable-gcrypt')
+        run: |
+          sudo apt-get install libgcrypt20-dev
+      - name: Setup Ubuntu specified compiler (gcc)
+        if: startsWith(matrix.os, 'ubuntu')  && startsWith(matrix.compiler, 'g')
+        run: |
+          sudo apt-get install ${{ format(matrix.compiler, 'cc') }}
+          sudo apt-get install ${{ format(matrix.compiler, '++') }}
```
utoni: Installing any C++ compiler is not required and superfluous.

.github/workflows/build.yml
```diff
@@ -2,73 +2,102 @@ name: Build
...
+      - name: Setup Ubuntu specified compiler (clang)
+        if: startsWith(matrix.os, 'ubuntu')  && startsWith(matrix.compiler, 'c')
+        run: |
+          sudo apt-get install ${{ format(matrix.compiler, '') }}
+          sudo apt-get install ${{ format(matrix.compiler, '++') }}
```
utoni: same here

.github/workflows/build.yml
```diff
@@ -2,36 +2,102 @@ name: Build
...
+        run: |
+          brew install libgcrypt
+      - name: Configure nDPI
+        if: startsWith(matrix.compiler, 'default-cc')
+        run: |
+          env CFLAGS='-Werror' ./autogen.sh --enable-debug-messages ${{ matrix.gcrypt }}
+      - name: Configure nDPI with specified GCC version on Ubuntu
+        if: startsWith(matrix.os, 'ubuntu') && startsWith(matrix.compiler, 'g')
+        run: |
+          env CC=${{ format(matrix.compiler, 'cc') }} CXX=${{ format(matrix.compiler, '++') }} CFLAGS='-Werror' ./autogen.sh --enable-debug-messages ${{ matrix.gcrypt }}
```
utoni: same here

.github/workflows/build.yml
```diff
@@ -2,36 +2,102 @@ name: Build
...
+      - name: Configure nDPI with specified CLANG on Ubuntu
+        if: startsWith(matrix.os, 'ubuntu') && startsWith(matrix.compiler, 'c')
+        run: |
+          env CC=${{ format(matrix.compiler, '') }} CXX=${{ format(matrix.compiler, '++') }} CFLAGS='-Werror' ./autogen.sh --enable-debug-messages ${{ matrix.gcrypt }}
```
utoni: and here

utoni (changes requested):
Good job!

.github/workflows/build.yml
```diff
@@ -2,36 +2,102 @@ name: Build
...
+      - name: Build nDPI
+        run: |
+          make all
```
utoni, line 138: Unfortunately, the `all` target does not build *all*.
It is a good idea to build as much as we can, so we have as less as broken code in the repository. An exception to the rule may be code only used for debugging purposes.
Things that are not build:
```
make -C example ndpiSimpleIntegration
make -C python
```

Not sure if we should also build some of the distro packages e.g. `packages/ubuntu`.

utoni: One job that configure's and build's nDPI with `--with-sanitizer`, `--with-pcre` and `--with-maxminddb` might be also interesting. What do you think?

aouinizied:
@lnslbrty Thanks for your feedbacks. 
* I fixed make all to take into account python wrapper and ndpiSimpleIntegration
* I fixed broken RCE detection within (ndpi_utils).
* Now tests take into account PCRE flag for WebattackRCE pcap. By default the result is the one with libpcre and when libpcre is not set, pcap is skipped (same logic done with gcrypt). [configure.seed + do.sh + pcap result file]
* I added jobs that enable: --with-sanitizer --with-pcre --with-maxminddb on all platform + each clang/gcc specific version job.


Here are some notes: 
* --with-sanitizer is broken on macOS (Apple LLVM CLANG) and MinGW.

Zied

aouinizied:
@lnslbrty also dropped appveyor.yml because it is not relevant anymore.

utoni (approved):

IvanNardi (approved):
Great and very useful wok. Thanks!

PR #1304, opened 2021-09-13
A (tiny) effort to reduce memory usage per flow

IvanNardi:
Follow-up of 22241a1d
Only trivial changes:
* remove completely unused fields
* remove fields only written (but never read)
* CSGO protocol only handles UDP traffic

### DIFF
--- a/src/lib/protocols/rtsp.c
+++ b/src/lib/protocols/rtsp.c
@@ -42,9 +42,6 @@ void ndpi_search_rtsp_tcp_udp(struct ndpi_detection_module_struct
 {
   struct ndpi_packet_struct *packet = &flow->packet;
 
-  struct ndpi_id_struct *src = flow->src;
-  struct ndpi_id_struct *dst = flow->dst;
-
   NDPI_LOG_DBG(ndpi_struct, "search RTSP\n");
 
   if (packet->parsed_lines == 0)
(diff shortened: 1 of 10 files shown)


src/lib/protocols/rtsp.c
```diff
@@ -84,20 +81,7 @@ void ndpi_search_rtsp_tcp_udp(struct ndpi_detection_module_struct
     if((memcmp(packet->payload, "RTSP/1.0 ", 9) == 0)
        || (strstr(buf, "rtsp://") != NULL)) {
       NDPI_LOG_DBG2(ndpi_struct, "found RTSP/1.0 \n");
-      if (dst != NULL) {
-	NDPI_LOG_DBG2(ndpi_struct, "found dst\n");
-	ndpi_packet_src_ip_get(packet, &dst->rtsp_ip_address);
-	dst->rtsp_timer = packet->current_time_ms;
-	dst->rtsp_ts_set = 1;
```
utoni, line 91: LoL, I did not know that OpenDPI used timestamps for flow correlation..
I am glad you removed that nonsense. :zany_face:

utoni (approved):

vel21ripn:
Then you need to remove these fields from ndpi_id_struct

IvanNardi:
> Then you need to remove these fields from ndpi_id_struct

I am not sure I understand your point. These fields *are* removed by this PR.
Am I missing something?

lucaderi:
I think this is good work towards cleaning nDPI. How did you do this? By hand or do you have some code analysis tool for this?

IvanNardi:
> I think this is good work towards cleaning nDPI. How did you do this? By hand or do you have some code analysis tool for this?

By hand, unfortunately... Focusing only on memory usage per flow, I had to check only a few structures, though.

PR #1309, opened 2021-09-19
Fix parsing of ipv6 packets with extension headers

IvanNardi:
Decoding of ipv6 traffic with extension headers was completely broken,
since the beginning of the L4 header was always set to a wrong value.

Handle the ipv6 fragments in the same way as the ipv4 ones: keep the first
one and drop the others.


src/lib/ndpi_main.c
```diff
@@ -4136,6 +4136,10 @@ int ndpi_handle_ipv6_extension_headers(struct ndpi_detection_module_struct *ndpi
       }
 
       *nxt_hdr = (*l4ptr)[0];
+      frag_offset = ntohs(*(u_int16_t *)((*l4ptr) + 2)) >> 3;
```
utoni, line 4139: Just a question: `l4ptr` does not point to any layer 4 header/payload but instead to an IPv6 extension header or am I wrong?

IvanNardi: Inside the function, `l4ptr` always points to the next header in the chain (a Ipv6 extension header or a proper L4 header).
Returning from the function, it points to the L4 header. I *think* this is the reason this parameter has been named in this way

utoni (commented):
Thanks <3

PR #1310, opened 2021-09-19
STUN: fix extraction of Realm attribute

IvanNardi:


src/lib/protocols/stun.c
```diff
@@ -349,14 +349,10 @@ static ndpi_int_stun_t ndpi_int_check_stun(struct ndpi_detection_module_struct *
...
 
-	      memset(flow->host_server_name, 0, sizeof(flow->host_server_name));
-
-	      for(j=0; j<i; j++) {
-		if((k+i) < payload_length)
-		  flow->host_server_name[j] = payload[k++];
-		else
-		  break;
+	      for(j=0; j<i && k<payload_length; j++) {
+		flow->host_server_name[j] = payload[k++];
```
utoni: Every time I see those buffer copy constructs in nDPI I am thinking if memcpy() could be faster, because of different optimizations done via builtins/compiler.
What do you think?

lucaderi: I think memcpy will be faster I agree

IvanNardi: I kept the for loop simply because it was already there (the bug was in the upper-limit check).
I am not so sure  if a memcpy is faster in this specific case (typically only  a dozen of bytes without any specific alignment), but I do think that memcpy is a cleaner solution anyway.
I'll push a new version

utoni: GCC with `-O2` tries to optimize the `memcpy` by using a builtin that instruments different buffer copy techniques depending on the length of a buffer. For example: `rep movsq` should be really fast for larger buffers.

See: [Godbolt Compiler Explorer: ndpi buffer copy test](https://godbolt.org/z/aM7nhWxT1)

utoni (approved):

src/lib/protocols/stun.c
IvanNardi: I agree that memcpy is definitely a better solution (cleaner and faster)

PR #1311, opened 2021-09-21
FTP: fix support for START-TLS sessions

IvanNardi:
When TLS-over-FTP is used, the credentials are encrypted. So we must not
wait for the username and the password commands, otherwise we elaborate a
lot of packets for nothing.


lucaderi:
I am not familiar with the protocol but I think we can merge this PR

python/ndpi_typestruct.py
```diff
@@ -507,6 +507,15 @@ class Http2(Structure):
         ("nat_ip", c_char * 24)
     ]
 
+class FtpImapPopSmtp(Structure):
```
utoni, line 510: How could this work before? The whole struct was missing in the flow struct. Python bindings via ctypes seems to be horribly broken.

src/lib/protocols/ftp_control.c
```diff
@@ -64,6 +64,11 @@ static int ndpi_ftp_control_check_request(struct ndpi_detection_module_struct *n
     return 1;
   }
 
+  if(ndpi_match_strprefix(payload, payload_len, "AUTH") ||
```
utoni, line 67: `ndpi_match_strprefix` calls `ndpi_match_prefix` which call `memcmp`, unsure if `strncasecmp`would be faster and if strings like `aUtH` are also valid

utoni (approved):

IvanNardi:
> I am not familiar with the protocol but I think we can merge this PR

Opportunistic TLS is an extension to introduce "TLS like" encryption in plain-text protocols (FTP/SMTP/IMAP/POP/...) without changing the default ports used (https://en.wikipedia.org/wiki/Opportunistic_TLS)
It was quite used once; now it is definitely discouraged (example: https://www.usenix.org/system/files/sec21-poddebniak.pdf)

PR #1313, opened 2021-09-22
QUIC: fix old GQUIC versions on big-endian machines

IvanNardi:
Not sure if now the QUIC core code is completely correct on big-endian machines: these were the easy fixes, though.


IvanNardi:
Is that the best way to handle MinGW?

utoni:
Probably. I was not able to find anything useful about byte order swapping on MSDN.

utoni (approved):

Issue #1312, opened 2021-09-22
nDPI and big endianess


utoni:
Although big endian architectures may not be very popular these days, nDPI has some serious issues dealing with them.

I've compiled nDPI for MIPS and ran `tests/do.sh` to verify that there are some.
Most of the issues are related to the recent ahocorasick library update in 65678dbeeabcf51e02ba3cc9ffbd36324a92a971. But there are also at least half a dozen endian issues in protocol dissectors.

I've created a guide (as Gist) how to reproduce those issues:
[HowTo-nDPI-MIPS.md](https://gist.github.com/lnslbrty/918abd33e7e99234d6c19b2800db8b53)

I would like to dig deeper in the ahocorasick library, but I was not able to find any upstream sources for the library which is used in nDPI. Any hints?
To prevent such issues in the future, a CI job building and testing for a big endian architecture may help.

Cheers!

utoni:
Some failing pcaps, caused by host patterns not found in ac automata:
```
qemu-mips -L /home/toni/x-tools/sysroot-mips-linux-gnu /home/toni/x-tools/sysroot-mips-linux-gnu/usr/bin/ndpiReader -p ../example/protos.txt -c ../example/categories.txt -r ../example/risky_domains.txt -j ../example/ja3_fingerprints.csv -S ../example/sha1_fingerprints.csv -q -t -i pcap/443-curl.pcap -w /tmp/reader.out -v 2 [old vs new]
1c1
< ntop	109	73982	1
---
> TLS	109	73982	1
8c8
[...]

PR #1314, opened 2021-09-22
Add arm64, armhf and s390x as part of CI.

aouinizied:
This PR add arm64, armhf and s390x as part of CI.
* We add 3 jobs where libgcrypt, pcre and maxminddb is enabled.
* --with-sanitizer is broken on macos, arm64 and s390x and thus, not set.
* @lnslbrty we cannot do single step output for these jobs (but it's better than not having these jobs).

IMHO, this PR should be merged even if it will fails when testing on s390x (big endian) as reported by @lnslbrty in https://github.com/ntop/nDPI/issues/1312.

Once merged, efforts fixing nDPI on big endian will be automatically tested using the CI.

Zied

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -80,21 +80,21 @@ jobs:
             pcre: "--with-pcre"
             maxminddb: "--with-maxminddb"
             msan: "" # Disable sanitizer on macos
-          - compiler:  "default-cc"
+          - compiler: "default-cc"
             os: ubuntu-latest
             arch: "arm64"
             gcrypt: ""
             pcre: "--with-pcre"
             maxminddb: "--with-maxminddb"
             msan: "" # Disable sanitizer on arm64
-          - compiler:  "default-cc"
+          - compiler: "default-cc"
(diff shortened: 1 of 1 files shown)


aouinizied:
We also detect now the following issue on ndpiReader using armhf CI:

ndpiReader.c:3602:36: runtime error: signed integer overflow: 1632342447 * 1000000 cannot be represented in type 'long int'
ndpiReader.c:3602:74: runtime error: signed integer overflow: 1632342447 * 1000000 cannot be represented in type 'long int'
ndpiReader.c:3603:33: runtime error: signed integer overflow: 1632342447 * 1000000 cannot be represented in type 'long int'
ndpiReader.c:3603:80: runtime error: signed integer overflow: 1632342444 * 1000000 cannot be represented in type 'long int'

Zied

utoni (approved):
Thanks for your awesome work @aouinizied
It is great to have a working CI.

IvanNardi (approved):
Useful improvements. Thanks!

PR #769, opened 2019-09-14
Fix non zero app data len increment.

rkerur:
Restrict SPLT score for first 32 packets in either dir
which should suffice for classification.

Signed-off-by: Ravi Kerur <ravi.kerur@viasat.com>

PR #1316, opened 2021-09-27
Fix IAT calculation

IvanNardi:
We shoudn't reset reference timestamps every 10 packets, i.e IAT
algorithm should be independent from SPLT and BD ones.

### DIFF
--- a/example/reader_util.c
+++ b/example/reader_util.c
@@ -1355,8 +1355,8 @@ static struct ndpi_proto packet_processing(struct ndpi_workflow * workflow,
     if((tcph != NULL) && (tcph->fin || tcph->rst || tcph->syn))
       begin_or_end_tcp = 1;
 
-    if(flow->entropy.flow_last_pkt_time.tv_sec) {
-      ndpi_timer_sub(&when, &flow->entropy.flow_last_pkt_time, &tdiff);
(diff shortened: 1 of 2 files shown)


IvanNardi:
This PR has been marked as Draft since it is only an incomplete POC: all the CI tests will fail.

I am not an expert of the algorithms implemented in nDPI, but I find strange that the packet timestamps used for IAT (Inter Arrival Time) stats are reset every 10 packets (via ndpi_clear_entropy_stats())

Is that correct?

Digging through the GIT history, I found that in the original implementation of the IAT algorithm (9a6f6d9fe) the timestamps were never reset.
Such a reset has been introduced in 4fd7e5734 (see #769), triggering a massive update of all the IAT values in the unit test results.
However, the comment https://github.com/ntop/nDPI/pull/769#issuecomment-537679757 seems to indicate that such change was unexpected.

@lucaderi: what do you think? That comment was made by you...

I can confirm that with this PR, the new IAT values are identical to the ones before 4fd7e5734.

Bottom line: if I am correct I will push a new version updating all the tests; otherwise, sorry for the noise, guys.

utoni:
Not sure, but IAT seems to me like an extracted feature usable for machine learning.

PR #1319, opened 2021-10-03
Remove `struct ndpi_packet_struct` from `struct ndpi_flow_struct`

IvanNardi:
There are no real reasons to embed `struct ndpi_packet_struct` (i.e. "packet")
in `struct ndpi_flow_struct` (i.e. "flow"). In other words, we can avoid
saving dissection information of "current packet" into the "flow" state,
i.e. in the flow management table.

The nDPI detection module processes only one packet at the time, so it is
safe to save packet dissection information in `struct ndpi_detection_module_struct`,
reusing always the same "packet" instance and saving a huge amount of memory.
Bottom line: we need only one copy of "packet" (for detection module),
not one for each "flow".

It is not clear how/why "packet" ended up in "flow" in the first place.
It has been there since the beginning of the GIT history, but in the original
OpenDPI code `struct ipoque_packet_struct` was embedded in
`struct ipoque_detection_module_struct`, i.e. there was the same exact
situation this commit wants to achieve.

Most of the changes in this PR are some boilerplate to update something
like "flow->packet" into something like "module->packet" throughout the code.
Some attention has been paid to update `ndpi_init_packet()` since we need
to reset some "packet" fields before starting to process another packet.

There has been one important change, though, in ndpi_detection_giveup().
Nothing changed for the applications/users, but this function can't access
"packet" anymore.
The reason is that this function can be called "asynchronously" with respect
to the data processing, i.e in context where there is no valid notion of
"current packet"; for example ndpiReader calls it after having processed all
the traffic, iterating the entire session table.

Mining LRU stuff seems a bit odd (even before this patch): probably we need
to rethink it, as a follow-up.

### DIFF
--- a/python/ndpi.py
+++ b/python/ndpi.py
@@ -645,6 +645,14 @@ struct ndpi_flow_udp_struct {
   /* NDPI_PROTOCOL_WIREGUARD */
   uint8_t wireguard_stage;
   uint32_t wireguard_peer_index[2];
+
+  /* NDPI_PROTOCOL_QUIC */
+  u_int8_t *quic_reasm_buf;
+  u_int32_t quic_reasm_buf_len;
+
+  /* NDPI_PROTOCOL_CSGO */
+  uint8_t csgo_strid[18],csgo_state,csgo_s2;
+  uint32_t csgo_id2;
 };
 
 struct ndpi_int_one_line_struct {
@@ -941,6 +949,9 @@ struct ndpi_detection_module_struct {
 
   uint8_t direction_detect_disable:1, /* disable internal detection of packet direction */
     _pad:7;
+
+  /* Current packet */
+  struct ndpi_packet_struct packet;
 };
 
 #define NDPI_CIPHER_SAFE                        0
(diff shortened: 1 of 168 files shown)


IvanNardi:
Hopefully, this patch is only the first of a series aiming to drastically reduce memory usage (per flow). Final numbers will be provided at the end, but this commit alone reduce such usage of **~50%** (from 2.77 KB to 1.41 KB, as reported by ndpiReader on x86_64)

lucaderi:
This is an important change. Looking at the code I think you are right but I didn't test it. If you did it, you can commit it, or if you want I can also test it when possible,

utoni:
This change is quite huge and I can not assure by just reviewing the code that there is no harmful bug or side-effect due to struct/code changes. I will do some tests with *nDPId* but additional tests might be useful @lucaderi.

IvanNardi:
> This is an important change. Looking at the code I think you are right but I didn't test it. If you did it, you can commit it, or if you want I can also test it when possible,

I run the following tests:
* ndpiReader (with ASAN and valgrind; same configuration used for unit tests) against a pretty big personal archive of traces, looking for memory errors/leaks -> No errors.
* ndpiReader (with ASAN and valgrind; same configuration used for unit tests) against some personal traces comparing the output, flow by flow, with and without this patch -> Same output.
* a minimal multi-thread environment (default compilation) with some real traffic. I was not able to analyze the classification flow-by flow, but global statistics were similar before and after applying this patch (and no crashes)

Any other suggestions?

lucaderi:
I think it's time to merge, do you agree?

IvanNardi:
> I think it's time to merge, do you agree?

Done. Let see what happens...

lucaderi:
Impressive

v4
nDPI Memory statistics:
	nDPI Memory (once):      221.05 KB    
	Flow Memory (per flow):  2.48 KB  

After this PR
nDPI Memory statistics:
	nDPI Memory (once):      223.38 KB    
	Flow Memory (per flow):  1.41 KB      <---

Great work

vel21ripn:
No! These are very bad changes!
nDPI can no longer work in multi-threaded mode.
Before this change, nDPI could work in multi-threaded mode and only the flow structure was needed to block, but now everything needs to be blocked!
This commit needs to be rewritten.

PR #1327, opened 2021-10-07
Fix `make dist` and add it to the CI along with `make -C doc text`.

utoni:
Not perfect because a simple `make dist` CI job does not ensure that the resulting tarball works (`./autogen.sh && make && ...`). But at least the packaging process does.


utoni:
`make -C doc release` is not part of any CI job as it ends up in an endless `make` loop, but this should be definitely part of a CI job later on

PR #1328, opened 2021-10-07
Get rid of `configure.seed` as it comes with some disadvantages.

utoni:
* using Autotools best-practices to achieve (hopefully) the same result

I know I will have hard times getting this PR merged. But this `autogen.sh <=> configure.seed` sed magic is not for what autotools was made for and does also break the configure/build system sometimes.
But now with a working CI it is possible to verify that it will work with MacOS. :smile:

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -139,6 +139,7 @@ jobs:
         if: startsWith(matrix.os, 'macOS') && startsWith(matrix.arch, 'x86_64')
         run: |
           brew install autoconf automake libtool pkg-config gettext json-c
+          brew install coreutils
       - name: Install MacOS Prerequisites (libgcrypt)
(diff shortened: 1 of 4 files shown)


utoni:
As mentioned in 3034e53d0061e1fa64f8081c4ca9ca26f2d34478, my previous attempt was reverted because CentOS6 does not have all AC_* macros available required for the `--enable-fuzztargets` checks. But CentOS6 does not get any updates since December 2020, so I think it is wise to drop support for such ancient OS's.

lucaderi:
This work looks good and it is going in the direction of simplifying the build process

utoni:
There was an issue with NDPI_MAJOR / NDPI_MINOR / NDPI_PATCH. I am now using PACKAGE_VERSION, which is set via AC_INIT() to get every part of the version triplet.

I've also added a `make install` job named **Install nDPI**. It install nDPI in a subdirectory and shows the content of same directory.

Last but not least, I've simplified some `make`'s in the CI e.g. replaced
```
cd ../python
make
cd .
```

with

```
make -C python
```

for the sake of readability.

PR #1330, opened 2021-10-09
Rework how hostname/SNI info is saved

IvanNardi:
Looking at `struct ndpi_flow_struct` the two bigger fields are
`host_server_name[240]` (mainly for HTTP hostnames and DNS domains) and
`protos.tls_quic.client_requested_server_name[256]`
(for TLS/QUIC SNIs).
    
This commit aims to reduce `struct ndpi_flow_struct` size, according to
two simple observations:
 1) maximum one of these two fields is used for each flow. So it seems safe
to merge them;
 2) even if hostnames/SNIs might be very long, in practice they are rarely
longer than a fews tens of bytes. So, using a (single) large buffer is a
waste of memory for all kinds of flows. If we need to truncate the name,
we keep the *last* characters, easing domain matching.
    
Analyzing some real traffic, it seems safe to assume that the vast
majority of hostnames/SNIs is shorter than 80 bytes.
    
Hostnames/SNIs are always converted to lowercase.
    
Attention was given so as to be sure that unit-tests outputs are not
affected by this change.
    
Because of a bug, TLS/QUIC SNI were always truncated to 64 bytes (the
*first* 64 ones): as a consequence, there were some "Suspicious DGA
domain name" and "TLS Certificate Mismatch" false positives.


IvanNardi:
This PR is marked as "Draft" since there are still some (minor) changes missing but I would like to have some preliminary feedback, nonetheless.

src/include/ndpi_typedefs.h
```diff
@@ -1240,10 +1240,14 @@ struct ndpi_flow_struct {
   /* Place textual flow info here */
   char flow_extra_info[16];
 
-  /* HTTP host or DNS query */
-  u_char host_server_name[240];
+  /* HTTP/DHCP/WHOIS hostname, DNS query, TLS/QUIC SNI, STUN realm, ... */
+  union {
+    char hostname_sni[80];
+    char *hostname_sni_ptr;
```
utoni: Is that union really necessary and required?
Why not using a `char *` with dynamic allocated memory only and skip the 80 bytes buffer?
As it is now, on x64 every dynamic allocation to `hostname_sni_ptr` wastes 72 bytes.
And additionally this is not error prone e.g. if someone writes text to the buffer `hostname_sni` and uses `hostname_sni_ptr` later on.

src/include/ndpi_typedefs.h
```diff
@@ -1240,10 +1240,14 @@ struct ndpi_flow_struct {
...
+  } h_sni;
```
utoni: `h_sni` seems too generic for me. When I read this, I would think about TLS/QUIC but not of HTTP/DHCP/DNS/etc..

example/ndpiReader.c
```diff
@@ -1481,7 +1488,16 @@ static void printFlow(u_int32_t id, struct ndpi_flow_info *flow, u_int16_t threa
...
   
   if(flow->ssh_tls.ssl_version != 0) fprintf(out, "[%s]", ndpi_ssl_version2str(flow->ndpi_flow, flow->ssh_tls.ssl_version, &known_tls));
-  if(flow->ssh_tls.client_requested_server_name[0] != '\0') fprintf(out, "[Client: %s]", flow->ssh_tls.client_requested_server_name);
+
+  if((flow->detected_protocol.master_protocol == NDPI_PROTOCOL_TLS)
+     || (flow->detected_protocol.app_protocol == NDPI_PROTOCOL_TLS)
+     || (flow->detected_protocol.master_protocol == NDPI_PROTOCOL_QUIC)
+     || (flow->detected_protocol.app_protocol == NDPI_PROTOCOL_QUIC)
+     || (flow->detected_protocol.master_protocol == NDPI_PROTOCOL_SSH)
+     || (flow->detected_protocol.app_protocol == NDPI_PROTOCOL_SSH)) {
```
utoni, line 1498: Are those protocol checks required? Even if you check additionally for the NUL below.

src/lib/ndpi_main.c
```diff
@@ -7783,3 +7758,92 @@ int ndpi_check_dga_name(struct ndpi_detection_module_struct *ndpi_str,
...
 }
+
+
+
+/* ******************************************************************** */
+
+char *ndpi_hostname_sni_get(const struct ndpi_flow_struct *flow) {
+  if(flow->h_sni_ptr)
+    return flow->h_sni.hostname_sni_ptr;
+  return (char *)flow->h_sni.hostname_sni;
```
utoni: The explicit cast is not required IMHO. `flow->h_sni.hostname_sni` may work depending on the CC / CC-version and CFLAGS and `&flow->h_sni.hostname_sni[0]` should always work.

IvanNardi: The cast seems to be required in both cases:
```
ndpi_main.c:7769:10: error: returning 'char const[80]' from a function with result type 'char *' discards qualifiers [-Werror,-Wincompatible-pointer-types-discards-qualifiers]
  return flow->h_sni.hostname_sni;
         ^~~~~~~~~~~~~~~~~~~~~~~~

```
```
ndpi_main.c:7769:10: error: returning 'const char *' from a function with result type 'char *' discards qualifiers [-Werror,-Wincompatible-pointer-types-discards-qualifiers]
  return &flow->h_sni.hostname_sni[0];
```
Probably changing the return type to ``const char *`` is a cleaner solution, but in that case I need to update a LOT of function prototypes (mainly in string matching stuff) to accept ``const char *`` parameter instead of ``char *``. 
I tried that way but there were too many changes and I didn't want to pollute this PR with (somehow) unrelated stuff. 
But, if you prefer, I could do it.

src/include/ndpi_typedefs.h
IvanNardi: The only point of this PR is to find a better tradeoff between memory usage per flow (i.e. size of ``struct ndpi_flow_struct``) and CPU utilization (via a a costly memory allocation).

In the current code we always avoid the malloc but we pre-allocate hundreds of bytes for every flow and we usually use only a few tens of them; it is a huge waste of memory.
With your proposal we significantly shrink  ``struct ndpi_flow_struct`` but we end up with another malloc for almost all the flows.

The proposed solution wants to be optimal for the vast majority of the flows (using small, but big enough, buffer pre-allocated; no malloc at runtime) and correctly handle the other flows too.
As a first iteration I simply used ``char hostname_sni[80]``, always truncating the hostname/SNI to 80 bytes; the union allows to have complete information even in the very rare cases where the hostname/SNI is longer, *without* any penalty for the common case. Yes, we waste 72 bytes in that scenario, but we probably use less memory than now anyway (and the dynamic allocation is an extreme rarely exception).

It always ends in the classic tradeoff memory vs CPU: nDPI is using too much memory per flow (IMO) and it would be nice to improve that without increasing CPU utilization too much.

The code is more complex, I agree, but I think that it is worth it. 
Such complexity is almost entirely hidden to user/application, tough. As reported in the commit message, some helpers has been added to correctly handle the union and the user/application only needs to care about `ndpi_hostname_sni_get()`

example/ndpiReader.c
IvanNardi: Yes, they are required to have the same outputs in unit tests.
This information is sometime logged as `[Host:...]` and sometime as `[Client: ...]` depending on the protocol
Having no changes in the unit tests greatly help to test/develop this PR (if really interested, we could try to simplify that as a follow-up)

src/include/ndpi_typedefs.h
IvanNardi: Any suggestions? `host_server_name_sni` (more or less, as the current names...)?

src/lib/ndpi_main.c
utoni: I overlooked the `const` qualifier for `flow`.

You could try changing the function signature of `ndpi_hostname_sni_get`:
`const struct ndpi_flow_struct *flow` to `struct ndpi_flow_struct *flow`.

This is not the best solution, but probably the one with least additional PR pollution. :)

example/ndpiReader.c
utoni: To keep future code modifications e.g. protocols that make use of `flow->host_server_name` as easy as possible, those additional checks make things unnecessary complicated.

A follow-up LGTM.

src/include/ndpi_typedefs.h
utoni: I like the naming you've used for TLS and QUIC: `struct {} tls_quic`, where the struct/union gives some information what protocols make use of such structs.

But in that case, it may even confuse people more: `union { ... } tls_quic_http_dhcp_dns_...`.
I am OOI (out-of-ideas) regarding the struct naming. Keep it as it is?

As long as the union has some comments giving the reader information about which protocols make use of the union, it should be fine I guess.

src/include/ndpi_typedefs.h
utoni: I would reeeeeally like to see a performance test and a diff between your PR and the dev branch. Unfortunately we do not have something available for now. :sob: 

I agree that dynamic memory allocations make things usually slower, but at the same time I would not underestimate `glibc`'s `malloc()` and heap management in general. But my `glibc` `malloc()` implementation knowledge is not that good.
Additionally I assume that most packets, which gets processed by `ndpi_detection_process_packet` using default preferences and application integration, do not contain any hostnames/SNIs.

src/include/ndpi_typedefs.h
IvanNardi: I'll update the existing comment in the next iteration

example/ndpiReader.c
IvanNardi: I agree we definitely need to simplify that code

src/lib/ndpi_main.c
IvanNardi: Ok, I'll try to update `ndpi_hostname_sni_get()` signature

src/include/ndpi_typedefs.h
IvanNardi: > Additionally I assume that most packets, which gets processed by `ndpi_detection_process_packet` using default preferences and application integration, do not contain any hostnames/SNIs.

That's exactly the point: nothing change with this PR for the packets without hostname/sni; nothing change for the vast majority of the packets with hostname/sni; for the very rare where hostname/sni are longer we pay the cost of a dynamic allocation.
But, contrary to what happens now, we always have the entire hostname/sni and we always (apart from pathological cases) use less memory (significantly less in the common cases), which is my goal.

I am even open to the possibility to use only `char hostname_sni[80]` (as described previously), to drastically simplify the PR: the memory usage improvements will still be there and we will lose very little respect the current code.

utoni: Ok, I get it. Let's stick to your current solution;

I am mostly concerned about that union.
What about:

```
struct {
```

?

We are wasting additional 4 / 8 bytes, but it is safer IMHO.

IvanNardi: It seems reasonable; I'll give it a try in the next patch iteration

IvanNardi:
This PR needs #1354 to be merged before moving on

utoni (approved):

IvanNardi:
With #1354  being merged, we can resume doing something useful here...

IvanNardi:
@lnslbrty I simplified this PR. Hostnames/SNIs are always saved in a pre-allocated buffer, trimming them if necessary.  This way, this change is greatly simplified.
If we really need the complete value, we can re-add the union/struct trick later

IvanNardi:
I analyze some real traffic, extracting all the hostnames/SNIs lengths, to determine the best value of the pre-allocated buffer.
80 bytes seems a fair tradeoff.
Two graphs below (same values, with linear and logarithmic scale)

sni_len
sni_len_log

utoni:
Interesting, thanks for that info.

utoni (approved):

utoni:
Maybe we can re-add the dynamic allocated buffer (as `char *`) at some point to struct. We are wasting 4 to 8 bytes for most host SNIs, but it provides some type safety.

What do you think?

src/lib/protocols/dns.c
```diff
@@ -368,6 +368,7 @@ static void ndpi_search_dns(struct ndpi_detection_module_struct *ndpi_struct, st
     int invalid = search_valid_dns(ndpi_struct, flow, &dns_header, payload_offset, &is_query);
     ndpi_protocol ret;
     u_int num_queries, idx;
+    char _hostname[256];
```
utoni, line 371: Why not `sizeof(host_server_name)` ?

IvanNardi: I think you mean `sizeof(flow->host_server_name)`... Anyway the basic idea is to always extract the full hostname in the dns code and than let `ndpi_hostname_sni_set()` to save the *last* 80 bytes (if longer). This way, domain matching shouldn't fail even with long hostnames.
If we use `_hostname[80]` we extract the *first* 80 bytes of the hostname.
We can obviously use `_hostname[80]` *and* modify DNS dissector to extract the *last* 80 characters itself but it seems unnecessarily complicated, IMO

utoni: Totally overlooked the string matching automata. Thanks for he hint.

PR #1336, opened 2021-10-11
Enable sanitizers on CIFuzz.

aouinizied:
Current CI Fuzz runs with default sanitizer which is "address".
In this PR, we enable all available sanitizers as a matrix.

Zied

### DIFF
--- a/.github/workflows/cifuzz.yml
+++ b/.github/workflows/cifuzz.yml
@@ -8,21 +8,21 @@ jobs:
       matrix:
         sanitizer: [address, undefined, memory]
     steps:
-    - name: Build Fuzzers - ${{ matrix.sanitizer }}
-      uses: google/oss-fuzz/infra/cifuzz/actions/build_fuzzers@master
-      with:
-        oss-fuzz-project-name: 'ndpi'
-        dry-run: false
-        sanitizer: ${{ matrix.sanitizer }}
-    - name: Run Fuzzers - ${{ matrix.sanitizer }}
-      uses: google/oss-fuzz/infra/cifuzz/actions/run_fuzzers@master
-      with:
-        oss-fuzz-project-name: 'ndpi'
-        fuzz-seconds: 600
-        dry-run: false
-        sanitizer: ${{ matrix.sanitizer }}
-    - name: Upload Crash
-      uses: actions/upload-artifact@v1
-      with:
-        name: ${{ matrix.sanitizer }}-artifacts
-        path: ./out/artifacts
+      - name: Build Fuzzers - ${{ matrix.sanitizer }}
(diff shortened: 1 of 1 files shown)


utoni:
Shouldn't the asan fuzzing job fail?

aouinizied:
@IvanNardi @lnslbrty Yes, this is weird.
I checked  the logs and we have errors and artifacts are created, so do not know why Run Fuzzers always pass.
I added a temporary hack (removed failure condition) to be able to have artifacts on all cases, so after this PR at least you will have the pcaps :).

IvanNardi:
@lnslbrty , here you are the QUIC pcap triggering the crash!
[quic-fuzz-crash.zip](https://github.com/ntop/nDPI/files/7322584/quic-fuzz-crash.zip)

Crash confirmed on current master

aouinizied:
pr_fuzz

And now we have all artifacts uploaded each time the CI is triggered. One artifact by sanitizer.

Zied

IvanNardi:
> @IvanNardi @lnslbrty Yes, this is weird. I checked the logs and we have errors and artifacts are created, so do not know why Run Fuzzers always pass. I added a temporary hack (removed failure condition) to be able to have artifacts on all cases, so after this PR at least you will have the pcaps :).

I confirm that the jobs don't fail but we have the artifacts!
Thank you very much. very useful!!-

utoni:
Thank you both! Mercí beaucoup and Mille grazie. :smile:

.github/workflows/cifuzz.yml
```diff
@@ -1,23 +1,28 @@
...
-       dry-run: false
-   - name: Run Fuzzers
-     uses: google/oss-fuzz/infra/cifuzz/actions/run_fuzzers@master
-     with:
-       oss-fuzz-project-name: 'ndpi'
-       fuzz-seconds: 600
-       dry-run: false
-   - name: Upload Crash
-     uses: actions/upload-artifact@v1
-     if: failure()
```
IvanNardi, line 20: @aouinizied, could you add a comment here about why we disabled this statement, please? So we don't forget it, hopefully....

aouinizied:
@IvanNardi Added the comment.
I will try to add a step where we count crash artifacts and if they are greater than 0, trigger failure and make uploading conditioned by this failure. It's a more clean workaround.

aouinizied:
@IvanNardi @lnslbrty I fixed this strange behavior by adding a step that checks manually for crashes. 
As GitHub action use the exit code as status. Now the condition if failure() will work as expected when uploading crashes.
Ready to merge even if Fuzz CIs will fail as expected :).

Zied

IvanNardi:
Thanks a lot!

PR #1337, opened 2021-10-11
QUIC: fix an integer overflow

IvanNardi:
Long standing bug: credits to @lnslbrty for digging into it and to
@aouinizied for the CI improvements


utoni (approved):
Nice one!

PR #1366, opened 2021-11-03
Fix use-of-uninitialized-value

### LINKED ISSUE #1346: Undefined behavior in hostname extraction when custom defined rules are set in protos.txt
Further to hostname extration from a packet applied to a custom protocol detection rule there is an undefined behavior as the hostname is occasionally extracted during subsequent runs.

Should we expect always an empty hostname when a packet is under ip based custom protocol detection?

## nDPI Environment:
* OS name: Ubuntu
* OS version: 12.04, 16.04
* Architecture: x86_64
* nDPI version or commit hash: 4.0-stable 
*  [config.log](https://github.com/ntop/nDPI/files/7344255/config.log)

##  Bug Reproduce Procedure
After the successful initialization of the library (as described in the quick start guide) I created a unit test in order to assert the behavior of the library. I used the attached "test.pcap" and the "protos.txt" to define some custom rules. After some continuous runs, the hostname "www.ntop.org" appearing in the fourth packet which is also applied to the third rule in protos.txt sometimes is returned during subsequent runs.

Also if you come up with valgring and rerun this test, two warnings are emerged when the librady tries to classify the already mentioned fourth packet in the attached pcap and the eight packet applied to the fourth rule in protos.txt. The warnings describe a `conditional jump or move depends on uninitialed value(s)`  in ndpi_main.c : 4625 and in ndpi_main.c: 4644

The reported bug is not reproducible using ndpiReader.

[attachments.zip](https://github.com/ntop/nDPI/files/7344048/attachments.zip)

IvanNardi:
Close #1346

### DIFF
--- a/src/lib/ndpi_main.c
+++ b/src/lib/ndpi_main.c
@@ -5376,7 +5376,7 @@ static int ndpi_do_guess(struct ndpi_detection_module_struct *ndpi_str, struct n
 
   if(flow->guessed_host_protocol_id >= NDPI_MAX_SUPPORTED_PROTOCOLS) {
     //u_int32_t num_calls;
-    NDPI_SELECTION_BITMASK_PROTOCOL_SIZE ndpi_selection_packet;
+    NDPI_SELECTION_BITMASK_PROTOCOL_SIZE ndpi_selection_packet = {0};
 
     /* This is a custom protocol and it has priority over everything else */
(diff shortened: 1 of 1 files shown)


vanggelis:
Hello Ivan,

Could you please merge the above mentioned commit to the 4.0-stable branch?

Thank you

IvanNardi:
> Could you please merge the above mentioned commit to the 4.0-stable branch?

Done

PR #1369, opened 2021-11-03
Differentiate between standard Amazon stuff (i.e market) and AWS

IvanNardi:


lucaderi:
Thanks, great idea to classify the Amazon protocols

PR #1371, opened 2021-11-07
Reduce memory used by `ndpiReader`

IvanNardi:
`ndpiReader` is only an example, aiming to show nDPI capabilities
and integration, without any claim about performances.

Nonetheless its memory usage per flow is *huge*, limiting the kinds
of traces that we can test on a "normal" hardware (example: scan
attacks).

The key reason of that behaviour is that we preallocate all the memory
needed for *all* the available features.

Try to reduce memory usage simply allocating some structures only
when they are really needed. Most significant example: JOY algorithms.

This way we should use a lot less memory in the two most common
user-cases:
 * `ndpiReader` invoked without any particular flag (i.e `ndpiReader -i
$FILENAME_OR_IFACE`)
 * internal unit tests

Before (on x86_64):
```
struct ndpi_flow_info {
[...]
	/* size: 7320, cachelines: 115, members: 72 */
```
After:
```
struct ndpi_flow_info {
[...]
	/* size: 2128, cachelines: 34, members: 75 */
```

### DIFF
--- a/example/ndpiReader.c
+++ b/example/ndpiReader.c
@@ -367,47 +367,50 @@ flowGetBDMeanandVariance(struct ndpi_flow_info* flow) {
   uint32_t tmp[256], i;
   unsigned int num_bytes;
   double mean = 0.0, variance = 0.0;
-  struct ndpi_entropy last_entropy = flow->last_entropy;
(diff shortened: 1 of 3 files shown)


lucaderi:
I think we could remove JOY at all

PR #1372, opened 2021-11-08
Update ndpi_main.c for HTTP line parser optimization and cleaning

ragostino:
HTTP line parser: optimization and cleaning.

### DIFF
--- a/src/lib/ndpi_main.c
+++ b/src/lib/ndpi_main.c
@@ -5884,27 +5884,6 @@ void ndpi_parse_packet_line_info(struct ndpi_detection_module_struct *ndpi_str, 
 
 	packet->http_num_headers++;
       } else
-
-      /* "Content-Type:" header line in HTTP. */
-      if(packet->line[packet->parsed_lines].len > 14 &&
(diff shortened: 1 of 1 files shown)


src/lib/ndpi_main.c
```diff
@@ -6020,6 +6004,19 @@ void ndpi_parse_packet_line_info(struct ndpi_detection_module_struct *ndpi_str,
 		      "Upgrade-Insecure-Requests: ", 27) == 0)) {
 	/* Just count. In the future, if needed, this if can be splited to parse these headers */
 	packet->http_num_headers++;
+      } else
```
IvanNardi, line 5986: @ragostino , while this "if/else list" seems fine, I am not sure that this code block that you moved should be put in it.\
Scenario: if we set `packet->content_line.ptr` around line 5900 (in that "else" block), we will need to check for these separators anyway, IMO.
Could you check, please?

ragostino: Good morning.
Yes, you are right! I edited and pushed a change.

PR #1374, opened 2021-11-17
Bittorrent: lookup into the cache only as a last effort

IvanNardi:
First, try to detect BT flows in the "standard way". Only if that fails,
try to lookup into the common cache.
This way:
* we are able to extract hash metadata from properly detected BT flows
* we should avoid some false positives since we allow other dissectors
to run


lucaderi:
Merged with https://github.com/ntop/nDPI/commit/4c39ed293e03c59af9d69a412107d41b00ebc21d

Thanks for the contribution

PR #1377, opened 2021-11-23
Make serialize risk and proto not dependant on any flow.

utoni:
Useful for other implementations that want to serialize risks/protos, but do not have flow memory allocated.


IvanNardi (approved):

lucaderi:
Thank you

PR #1379, opened 2021-12-03
QUIC: add support for QUICv2 (draft 00)

IvanNardi:
It is already time to start looking at the new QUIC version.
See: https://datatracker.ietf.org/doc/html/draft-ietf-quic-v2-00


lucaderi:
Great add-on: QUIC is the new TCP IMHO

Issue #1279, opened 2021-08-20
Purpose of  `struct ndpi_id_struct`


IvanNardi:
Looking at ways to reduce memory usage, I bumped into the structure `ndpi_id_struct`. We have 2 of this variables for each `struct ndpi_flow_struct`: `src` and `dst` (plus another pointer `server_id`: let ignore it for the time being)

What is their purpose?

Looking at the comments it seems they are a way to correlate different flows originated or terminated  from/to the same "host/user", i.e a way to collect some information about "host/user". And that makes sense...
However... the lifetime of this structures is the same of the lifetime of the flow they are used in!!!
In other worlds, `src` and `dst` are -always- unconditionally allocated (and freed) when the corresponding `struct ndpi_flow_struct` is allocated (or freed): they are not persistent objects across different sessions. This is the behavior of `ndpiReader` and of `ntop`, too.

I quickly run the unit tests with a slightly modified version of `ndpiReader` where these two structures are never allocated: no difference at all.

Right now, using this structure (in the current way) )seems to be only a huge waste of memory.

Note that the definition of `struct ndpi_id_struct` pre-date git history (~6 years ago) and it has never been touched ever since...

Note also that a similar (but not identical) functionality is achieved, in recent code, via the LRU caches.

Before talking about what should be done to improve the current situation, am I missing something obvious here? Have I mis-read the code?

vel21ripn:
In ndpi-netfilter, these structures are stored in a separate RB-tree. Each ndpi_id_struct has a usage counter and only deletes when the counter is zero.
The ndpi_id_struct number is approximately 60-70% of the number of connections.
Another question is how much ndpi_id_struct is used.

lucaderi:
I agree that this is something inherited from OpenDPI that needs to be cleaned up

IvanNardi:
> I agree that this is something inherited from OpenDPI that needs to be cleaned up

Yeah, this code is present in the original OpenDPI repository. From its manual (emphasis mine):




```
A similar state buffer is maintained for every subscriber. In most situations, a subscriber is identified by its
internal IP address. In this case, ****a memory buffer for every internal IP address must be maintained****.

OpenDPI is designed to work in different network situations. There are two different situations for the subscriber
mapping.
In an access gateway situation, every packet belongs to one subscriber. The subscriber is either the sender or
the receiver of the packet. In this case, either the source or the destination subscriber is known.

[...]

PR #1380, opened 2021-12-03
Improve IPv6 support, enabling IPv6 traffic on (almost) all dissectors.

IvanNardi:
There are no valid reasons for a (generic) protocol to ignore IPv6
traffic.
Note that:
* I have not found the specifications of "CheckPoint High Availability
Protocol", so I don't know how/if it supports IPv6
* all LRU caches are still IPv4 only

Even if src_id/dst_id stuff is probably useless (see #1279), the right
way to update the protocol classification is via `ndpi_set_detected_protocol()`


lucaderi:
yes caching need to support IPv6 you're right.

PR #1390, opened 2021-12-22
Add support for ICloud Private Relay

IvanNardi:
See:
https://www.apple.com/privacy/docs/iCloud_Private_Relay_Overview_Dec2021.PDF

TODO: an up-to-date list of egress IP ranges is publicly available. Can
we use it somehow?


lucaderi:
If you share the IP list I will try to do something similar to what I have done with Azure

IvanNardi:
The list is available at: https://mask-api.icloud.com/egress-ip-ranges.csv
I am not sure how to handle these ips. They are the addresses of the egress relays, NOT of the servers.
For example: if a TLS.Youtube flow has the source address in this list, how should we classify the flow? iCloudRelay? iCloudRelay/Youtube? Should we ignore the relay stuff if we already have a proper classification?

It should be the same as for the TOR case, where we have the list of the egress nodes... the problem is that I don't know how nDPi exactly handle TOR flows :-)

PR #1395, opened 2022-01-08
Add a "confidence" field about the reliability of the classification.

IvanNardi:
As a general rule, the higher the confidence value, the higher the
"reliability/precision" of the classification.

In other words, this new field provides an hint about "how" the flow
classification has been obtained.
For example, the application may want to ignore classifications "by-port"
(they are not real DPI classifications, after all) or give a second
glance at flows classified via LRU caches (because of false positives).

Setting only one value for the confidence field is a bit tricky: more
work is probably needed in the next future to tweak/fix/improve the logic.


lucaderi:
I think this is a great addition I wanted to implement for a long time. In particular, I was thinking about protocols such as BitTorrent that sometimes are detected using a few bytes that might be a false positive. So the code is ok but in the future I want to extend it to cover DPI "reliability levels" too.

PR #1405, opened 2022-01-14
XBox, Diameter: fix dissectors initialization

IvanNardi:
These dissectors have *never* been triggered because their registration
functions use the wrong parameter/bitmask.
Diameter code is buggy since the origianl commit (1d108234), while
XBox code since 5266c726.

Fix some false positives in Xbox code.


IvanNardi:
Thanks to oss-fuzz for pointing out that there was something really wrong with these two protocols

Screenshot from 2022-01-13 15-01-33a

PR #1410, opened 2022-01-16
Zattoo: fix Null-dereference READ with ipv6 traffic

IvanNardi:
Fix: 20b5f6d7
Detected by oss-fux:
https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=43700


lucaderi:
I am wondering if this protocol is still in use or if we can remove it

IvanNardi:
> I am wondering if this protocol is still in use or if we can remove it

A Zattoo streaming service still exists (according to Google/Wikipeadia; I don't personally know it).
I doubt this OpenDPI code (based on plain text HTTP) still detects it, though

Issue #1386, opened 2021-12-16
Incorrect comparison of nDPI_flow_info structures in ndpiSimpleIntegration


ol-andreyizrailev:
https://github.com/ntop/nDPI/blob/8e512ff50331e2750d13b3d983f3b3216134ad18/example/ndpiSimpleIntegration.c#L362-L402
I believe the logic in `ip_tuples_compare()` is incorrect.
First of all, on the first line of the function (which is 365) `B->l3_type` should obviously be checked to be `L3_IP`, not `L3_IP6`.  The same problem presents in `ip_tuples_equal()`. 
Another issue here is the comparion logic itself. Let's take a look at the 366 to 375. You simplified code looks like this:
```
if (A->src < B->src || A->dst < B->dst)
{
    return -1;
}
if (A->src > B->src || A->dst > B->dst)
{
    return 1;
}
```
Let's now imagine that we pass to the function two structure pointers X and Y which have the following values: `X->src = 1, X->dst = 2, Y->src = 3, Y->dst = 1`.
If X is the first argument of the function and Y is the second one, the function will return -1 since X->src is less than Y->src.
But if Y is the first argument and X is the second one, the function will also return -1 since Y->src is greater than X->src and Y->dst is less than X->dst.
The first case implies that X is less than Y and the second case implies that Y is less than X. They are contradictory. The result of comparison must not be affected by the order of arguments passed to the function.
The same problem exists further on with IPv6 case and ports.
I suggest you do something like this:
```
if (A->src < B->src)
    return -1;
else if (A->src > B->src)
    return 1;

if (A->dst < B->dst)
    return -1;
else if (A->dst > B->dst)
    return 1;
```

This will also allow you to get rid of `ip_tuples_equal()` call in `ndpi_workflow_node_cmp()`.

utoni:
I agree, these are two valid bugs. The first one was probably a typ0, but the second one is a logic fail.
Will fix that ASAP.
Thanks for reporting those!

utoni:
Fixed with #1418.

PR #1418, opened 2022-01-26
Fixed wrong ip tuple comparison. #1386

utoni:
* Added u32 pads to `union ip_tuple` so btree search should now work as expected.
   The bug caused new flow's when the remote answers, resulting in two Flows per direction. Fail.
 * Fixed a race condition during shutdown phase.

Two more race conditions in ndpiSimpleintegration exist. They will be part of a follow-up as libnDPI is also affected by a data race. Maybe we should think about adding `-fsanitize=thread` to the CI.

### DIFF
--- a/example/ndpiSimpleIntegration.c
+++ b/example/ndpiSimpleIntegration.c
@@ -16,6 +16,7 @@
 #include <string.h>
 #include <unistd.h>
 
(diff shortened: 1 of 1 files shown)


IvanNardi (approved):
+1 on `-fsanitize=thread `

PR #1429, opened 2022-02-01
Added lightweight implementation of libgcrypt.

vel21ripn:
Implementation borrowed from the https://github.com/theicfire/aes-gcm.git project

Speed testing (Xeon(R) CPU E3-1230 V2 @ 3.30GHz):

gcrypt-gnu        Test md 2.89 sec enc 2.78 sec dec 0.94 sec
gcrypt-int        Test md 3.62 sec enc 1.90 sec dec 4.12 sec
gcrypt-gnu-nonopt Test md 3.74 sec enc 4.96 sec dec 4.00 sec

gcrypt-gnu-nonopt - libgcrypt compiled without hardware acceleration
  --disable-padlock-support --disable-aesni-support \
  --disable-shaext-support --disable-pclmul-support \
  --disable-sse41-support --disable-drng-support \
  --disable-avx-support --disable-avx2-support \
  --disable-neon-support --disable-arm-crypto-support \
  --disable-ppc-crypto-support --disable-amd64-as-feature-detection


src/lib/third_party/src/gcrypt.c
```diff
@@ -0,0 +1,325 @@
...
+#include "gcrypt/cipher.c"
+#include "gcrypt/cipher_wrap.c"
+#include "gcrypt/digest.c"
+
+int gcry_control (int ctl,int val) {
+    return 0;
+}
+
+const char *gcry_check_version(void *unused) {
+    return "1.8.6";
```
IvanNardi, line 94: Maybe "internal" or something similar? We shouldn't cheat here, IMO

IvanNardi (changes requested):
I really like this PR!
I can't comment on the crypto code itself, but I have a few questions:
1) could you add the necessary code to support liggpg-error too? Right now, compilation fails if you enable the two debug defines at the beginning of `quic.c` file. Even an "empty" wrapper should be fine, IMO
```
protocols/quic.c:252:18: error: unknown type name ‘gpg_error_t’; did you mean ‘gcry_error_t’?
  252 | char *__gcry_err(gpg_error_t err, char *buf, size_t buflen)
      |                  ^~~~~~~~~~~
      |                  gcry_error_t
```

2) Did you test your code on a big-endian machine? A lot of efforts has been made lately to try to support big endian architectures. You can enable the tests directly in GitHubActions adding `./tests/do.sh` at the end of `.github/workflows/build.yml` (there is a comment near the the end). The tests are expected to fail, but I do prefer that this code doesn't add any new issues in this area

3) Could you take a look at the compilation errors on Mingw, please?

4) Could you take a look at the warnings reported by Codiga: https://github.com/ntop/nDPI/pull/1429/checks?check_run_id=5033301932?

Great work!

vel21ripn:
@IvanNardi 
Why do all build tests use mingw64 and not gcc or clang?
What's with the weird version of glibc? man 3 bswap
````
        #include <byteswap.h>

        uint16_t bswap_16(uint16_t x);
        uint32_t bswap_32(uint32_t x);
        uint64_t bswap_64(uint64_t x);
````

**Comment by vel21ripn** [comment]
1. I added  "empty" wrapper for __gcry_err()
2. s390 is BigEndian. We need an additional build with "--disable-gcrypt" option for s390.
3. fixed.
4. Work in  progress

**Comment by IvanNardi** [comment]
> 1. I added  "empty" wrapper for __gcry_err()
> 
>     2. s390 is BigEndian. We need an additional build with "--disable-gcrypt" option for s390.
> 
>     3. fixed.
> 
>     4. Work in  progress

@vel21ripn, thanks!
Could you push a new commit with the following change, please?
```
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -101,6 +101,13 @@ jobs:
             pcre: "--with-pcre"
             maxminddb: "--with-maxminddb"
             msan: ""
+          - compiler: "default-cc"
+            os: ubuntu-latest
+            arch: "s390x"
+            gcrypt: "--disable-gcrypt"
+            pcre: "--with-pcre"
+            maxminddb: "--with-maxminddb"
+            msan: ""
     steps:
       - name: Setup multiarch/qemu-user-static
         if: startsWith(matrix.os, 'ubuntu') && !startsWith(matrix.arch, 'x86_64')
@@ -285,5 +292,6 @@ jobs:
             make -C example ndpiSimpleIntegration &&
             make -C rrdtool &&
             make -C python &&
+            ./tests/do.sh &&
             ./tests/do-unit.sh
             "
```
the idea is let CI to test the big-endian stuf...

utoni:
I am a bit concerned about possible security issues of [https://github.com/theicfire/aes-gcm](https://github.com/theicfire/aes-gcm).
The repository was last updated in 2016 and the mbedTLS code seems to originate from a mbedTLS library version back in 2015. The repository is not well maintained and (probably) not widely used.
[https://tls.mbed.org/security](https://tls.mbed.org/security) shows some serious security vulnerabilities between 2015 till today. I did not verify if the code of `aes-gcm` inherited some of mbedTLS vulnerabilites, but I would assume so.

Wouldn't it be better to integrate mbedTLS in libnDPI's third party modules, because it is well maintained and actively used on lot's of different platforms. What do you guys think?

vel21ripn:
I agree that the code is old, but I couldn't find another simple and cross-platform implementation of aes128-cbc and aes128-gcm.
Everything modern (like libgcrypt, openssl, libressl) is a monsters!
I can compare the added code with the actual code https://github.com/ARMmbed/mbedtls.
Now we need to check the correct operation on platforms with big endian.

IvanNardi:
> I am a bit concerned about possible security issues of https://github.com/theicfire/aes-gcm. The repository was last updated in 2016 and the mbedTLS code seems to originate from a mbedTLS library version back in 2015. The repository is not well maintained and (probably) not widely used. https://tls.mbed.org/security shows some serious security vulnerabilities between 2015 till today. I did not verify if the code of `aes-gcm` inherited some of mbedTLS vulnerabilites, but I would assume so.
> 
> Wouldn't it be better to integrate mbedTLS in libnDPI's third party modules, because it is well maintained and actively used on lot's of different platforms. What do you guys think?

I really don't like to use old, unmaintained code because of the usual reasons: algorithm correctness, memory bugs, missing "upstream" to pull bug-fixes from... So, I do prefer to use something newer and well maintained.

Nonetheless, I am not worried from a security point of view. Let me explain. Even if we use a crypto library, we don't do any "real" crypto stuff, IMO: we don't handle/generate private keys and we don't have access to any private data, then we don't risk to leak any sensitive information.
I personally doubt that decrypting a QUIC Client Hello with a well-know key/salt could be considered an operation involving any sensitive data (at least, not more sensitive than any other protocols handled by nDPI...)
Am I too naive?

vel21ripn:
It makes sense to upgrade to the current implementation https://github.com/ARMmbed/mbedtls.git (v3.1.0).
We will get error code decryption and aesni support.
This will take several days.

utoni:
> Am I too naive?

Maybe you are not naive. Maybe I'm just paranoid.

IvanNardi:
> 
> Maybe you are not naive. Maybe I'm just paranoid.

Yours are valid concerns. I was thinking if there are reasons to treat this crypto code differently from all other code in nDPI: I don't have a proper reply to that, but I don't usually handle crypto/security code in the first place

utoni:
I just remembered that I once used [libtomcrypt](https://github.com/libtom/libtomcrypt) for a similar problem. Not sure if it is easier to integrate then mbedTLS.

viniciussn:
What was the conclusion? This PR cannot be merged?

vel21ripn:
@viniciussn I am preparing a new version of the lightweight implementation of libgcrypt.

PR #1430, opened 2022-02-02
Added `--enable-code-coverage' build using lcov for coverage generation.

utoni:
* CI job generates a coverage report

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -10,6 +10,39 @@ on:
   release:
     types: [created]
 jobs:
+  coverage:
+    name: Coverage (ubuntu-latest)
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+        with:
+          fetch-depth: 0
+      - name: Install Ubuntu Prerequisites
+        run: |
+          sudo apt-get update
+          sudo apt-get install autoconf automake libtool pkg-config gettext libjson-c-dev flex bison libpcap-dev
+          sudo apt-get install rrdtool librrd-dev
+          sudo apt-get install libgcrypt20-dev libpcre3-dev libmaxminddb-dev lcov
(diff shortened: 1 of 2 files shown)


utoni:
This PR generates a HTML based code coverage report and uploads it as CI artifact. It can be viewed locally for now. It is also possible to use third party services e.g. [codecov.io](https://codecov.io) but not implemented yet. The report literally shows how badly libnDPI's code is tested.

lucaderi:
Interesting contribution

IvanNardi:
> This PR generates a HTML based code coverage report and uploads it as CI artifact. It can be viewed locally for now. It is also possible to use third party services e.g. [codecov.io](https://codecov.io) but not implemented yet. The report literally shows how badly libnDPI's code is tested.

@lnslbrty : I haven't yet looked at this report.
oss-fuzz generates something similar (https://storage.googleapis.com/oss-fuzz-coverage/ndpi/reports/20220227/linux/report.html; the link should be public).
The results seem similar to what you have found: good average coverage for dissector code; non-existent coverage for common code

utoni:
> > This PR generates a HTML based code coverage report and uploads it as CI artifact. It can be viewed locally for now. It is also possible to use third party services e.g. [codecov.io](https://codecov.io) but not implemented yet. The report literally shows how badly libnDPI's code is tested.
> 
> @lnslbrty : I haven't yet looked at this report. oss-fuzz generates something similar (https://storage.googleapis.com/oss-fuzz-coverage/ndpi/reports/20220227/linux/report.html; the link should be public). The results seem similar to what you have found: good average coverage for dissector code; non-existent coverage for common code

I think that a code coverage percentage generated from fuzzing may be misleading, because it is basically a non-reproducible random value that may vary with every execution / on every device.
I would prefer generating code coverage from integration or better unit tests.

PR #1436, opened 2022-02-04
Add few scripts to easily update some IPs lists

IvanNardi:
Some IPs lists should be updated frequently: try to easy the process.
The basic idea is taken from d59fefd0 and a8fe74e5 (for Azure
addresses): one specific .c.inc file and one script for each protocol.

Add the possibility to don't load a specific list.

Rename the old NDPI_PROTOCOL_HOTMAIL id to NDPI_PROTOCOL_MS_OUTLOOK,
to identify Hotmail/Outlook/Exchange flows.

TODO: ipv6

Remove the 9 addresses associated to BitTorrent: they have been added in
e2f21116 but it is not clear why all the traffic to/from these ips
should be classified as BitTorrent.


lucaderi:
This looks a great work to automate nDPI

IvanNardi:
> This looks a great work to automate nDPI

Thanks for taking care of the "quotes" issues: I should have noticed the Codiga warnings....

PR #1438, opened 2022-02-07
Fix memory access in ndpi_strnstr()

IvanNardi:
Found by oss-fuzz:
https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=44482

It should be the same error reported (and only partially fixed) in
79968f32


lucaderi:
Good catch!

PR #1440, opened 2022-02-08
HSRP: add support for IPv6

IvanNardi:


lucaderi:
Thank you (I didn't have IPv6 HSRP to test with)

PR #1447, opened 2022-02-19
Add support for Google Cloud

IvanNardi:
Differentiate between Google its own apps/services and Google Cloud.
We already do something similar for Amazon vs AWS and Microsoft vs Azure.


lucaderi:
A good step towards automation, and service differentiation

PR #1448, opened 2022-02-19
EthernetIP: add missing initialization

IvanNardi:
Fix: 1e1cfb89


lucaderi:
This was my fault: thank you for fixing it

PR #1449, opened 2022-02-20
Add some scripts to easily update some IPs lists

IvanNardi:
While the lists in a6ff0dd0 and 2f5f445f are somehow provided by the
companies themselves (or by some interested parties), these new lists
are directly extracted from BGP information, via AS prefixes.

*Usually*, these new lists are far more stable than the previous ones.

TODO:
* add some other ASNs (see `src/lib/ndpi_content_match.c.inc`)
* IPv6, as usual :-(


utoni (commented):
LGTM, but it would be awesome if nDPI had the capability to update such things during runtime as well. That would make things much easier for me.

Why not merge all those `asn_*.sh` into a single one?

IvanNardi:
> LGTM, but it would be awesome if nDPI had the capability to update such things during runtime as well. That would make things much easier for me.

I agree with you but I think that is a separate issue.

> Why not merge all those `asn_*.sh` into a single one?

Good idea. Done in the last iteration

utoni (approved):

PR #1459, opened 2022-02-26
Bug fixing.

vel21ripn:
The '--enable-debug-messages' option works again.
Fixed warning in ahocorasick.c
Fixed integer overflow in ndpiReader.c for 32bit systems.

### DIFF
--- a/src/include/ndpi_main.h
+++ b/src/include/ndpi_main.h
@@ -25,6 +25,10 @@
 #define __NDPI_MAIN_H__
 
 #include "ndpi_includes.h"
+#ifdef NDPI_LIB_COMPILATION
+/* for macros NDPI_LOG_* in ndpi_define.h */
+#include "ndpi_config.h"
+#endif
 #include "ndpi_define.h"
 #include "ndpi_protocol_ids.h"
 #include "ndpi_typedefs.h"
--- a/example/ndpiReader.c
+++ b/example/ndpiReader.c
@@ -2958,7 +2958,7 @@ static void printResults(u_int64_t processing_time_usec, u_int64_t setup_time_us
 	struct tm result;
 
 	if(live_capture) traffic_duration = processing_time_usec;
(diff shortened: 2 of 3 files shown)


src/include/ndpi_main.h
```diff
@@ -25,6 +25,10 @@
 #define __NDPI_MAIN_H__
 
 #include "ndpi_includes.h"
+#ifdef NDPI_LIB_COMPILATION
+/* for macros NDPI_LOG_* in ndpi_define.h */
+#include "ndpi_config.h"
+#endif
```
IvanNardi, line 31: @lucaderi, could you take a look, please? You removed a similar include in https://github.com/ntop/nDPI/commit/246d115d7bf126c284df739c35cd9b790bf140cf to "prevents ndpi_config.h to be included everywhere in apps using nDPI that might leade to #define redefinitions".
Could you check if this fix is fine or if it triggers some regressions? (it shoudn't, because of the #ifdef)

vel21ripn: If we build the library with the "--enable-debug-messages" option, then we want debug printing to work.
Perhaps this can be done using configure.ac by adding -DNDPI_ENABLE_DEBUG_MESSAGES to CFLAGS.
There is another hidden problem: the size of the ndpi_detection_module_struct structure depends on this option.
Let's leave the "user_data" field unconditional. 8 bytes in 250 kilobytes won't change anything.

IvanNardi: You patch clearly fix the debug log. My only concern is if it breaks whatever @lucaderi  fixed in https://github.com/ntop/nDPI/commit/246d115d7bf126c284df739c35cd9b790bf140cf.

> There is another hidden problem: the size of the ndpi_detection_module_struct structure depends on this option.
> Let's leave the "user_data" field unconditional. 8 bytes in 250 kilobytes won't change anything.

No strong opinion on that

lucaderi: For me we can accept it, try to compile on all platforms and see what happens.

PR #1461, opened 2022-02-27
Removed some unused fields

IvanNardi:


python/ndpi_typestruct.py
```diff
@@ -229,7 +229,6 @@ class CustomCategories(Structure):
     ("detection_bitmask", NDPIProtocolBitMask),
     ("generic_http_packet_bitmask", NDPIProtocolBitMask),
```
utoni:
```suggestion
```

utoni (changes requested):

python/ndpi_typestruct.py
IvanNardi: Thanks for spotting that! Python bindings are a completely mess...
Fixed

utoni: True. We might need some sort of automation there.

utoni (approved):

lucaderi:
@aouinizied I am going to merge this code but your feedback on Python bindings is welcome

PR #1462, opened 2022-02-27
Add a new flow risk `NDPI_ANONYMOUS_SUBSCRIBER`

IvanNardi:
The main goal of a DPI engine is usually to determine "what", i.e. which
types of traffic flow on the network.
However the applications using DPI are often interested also in "who",
i.e. which "user/subscriber" generated that traffic.

The association between a flow and a subscriber is usually done via some
kind of DHCP/GTP/RADIUS/NAT mappings. In all these cases the key element
of the flow used to identify the user is the source ip address.

That usually happens for the vast majority of the traffic.

However, depending on the protocols involved and on the position on the net
where the traffic is captured, the source ip address might have been
changed/anonymized. In that case, that address is useless for any
flow-username association.

Example: iCloud Private Relay traffic captured between the exit relay and
the server.
See the picture at page 5 on:
https://www.apple.com/privacy/docs/iCloud_Private_Relay_Overview_Dec2021.PDF

This commit adds new generic flow risk `NDPI_ANONYMOUS_SUBSCRIBER` hinting
that the ip addresses shouldn't be used to identify the user associated
with the flow.
As a first example of this new feature, the entire list of the relay ip
addresses used by Private Relay is added.

A key point to note is that list is NOT used for flow classification
(unlike all the other ip lists present in nDPI) but only for setting this
new flow risk.

TODO: IPv6


doc/flow_risks.rst
```diff
@@ -272,4 +272,9 @@ NDPI_HTTP_CRAWLER_BOT
 ===================================
 The risk is set whenever a crawler/bot/robot has been detected
 
+.. _Risk 045:
 
+NDPI_ANONYMOUS_USER
```
utoni: `NDPI_ANONYMOUS_USER` may be misleading.
Is `NDPI_ANONYMOUS_ADDRESS` or `NDPI_ANONYMOUS_SUBSCRIBER` more appropriate?

utoni (commented):
1. Does it make sense to set this risk not only for iCloud relays, but also for other anonymization services e.g. TOR or if a HTTP Proxy is detected?
2. I do not see any `NDPI_SET_BIT`. How is this risk set?

doc/flow_risks.rst
IvanNardi: `NDPI_ANONYMOUS_SUBSCRIBER` is a better name. Thanks for the suggestion

src/lib/ndpi_main.c
```diff
@@ -5899,6 +5933,27 @@ ndpi_protocol ndpi_detection_process_packet(struct ndpi_detection_module_struct
...
+        ndpi_risk_enum net_risk;
+
+        addr.s_addr = packet->iph->saddr;
+        net_risk = ndpi_network_risk_ptree_match(ndpi_str, &addr);
+        if(net_risk == NDPI_NO_RISK) {
+          addr.s_addr = packet->iph->daddr;
+          net_risk = ndpi_network_risk_ptree_match(ndpi_str, &addr);
+        }
+        if(net_risk != NDPI_NO_RISK)
+          ndpi_set_risk(ndpi_str, flow, net_risk);
```
IvanNardi, line 5952: The risk should be set here. Am I missing something?

IvanNardi:
> 1. Does it make sense to set this risk not only for iCloud relays, but also for other anonymization services e.g. TOR or if a HTTP Proxy is detected?

Yeah, the idea is to incrementally extend this feature with other services.
The difficult part is how to detect them, IMO
Let's talk about TOR for example: we should have an ip list with the "ingress" nodes (to classify the flow as `NDPI_PROTOCOL_TOR`) and another list with the "egress" nodes (to set this new risk). I don't know if that is even feasible in the first place, I am not TOR expert.
Similarly for the HTTP Proxy: are we able to distinguish the two "legs", before and after the proxy?

With PrivateRelay we are able to identify the flows before the "proxies" (via SNI) and the (different) flows after it (via this new ip list)

I think that when nDPI classified a flow as `NDPI_PROTOCOL_HTTP_PROXY` it is the first leg, i.e. the source address really identified the user.
Not idea which TOR list are we using right now...

src/lib/ndpi_main.c
utoni: Nope. I've overlooked that line.

utoni:
> Let's talk about TOR for example: we should have an ip list with the "ingress" nodes (to classify the flow as NDPI_PROTOCOL_TOR) and another list with the "egress" nodes (to set this new risk). I don't know if that is even feasible in the first place, I am not TOR expert.

It should be possible to use the publicly available list of TOR exit nodes and set the risk according to their address tuple. Or is that not how the feature is meant to be used? Anyway, it might be better to solve this in a follow-up, so this PR does not get bloated.

utoni (approved):

IvanNardi:
> It should be possible to use the publicly available list of TOR exit nodes and set the risk according to their address tuple. Or is that not how the feature is meant to be used?

You got that right; that is the general idea. I know that there are lists with (only) the exit nodes, but I don't know if there are lists only with the "ingress" nodes (to flow classification).
I will try something...

lucaderi:
I think the idea and the PR are great and I believe they can be merged. As you said probably some extra work will be necessary in order to complete it

Issue #719, opened 2019-06-14
nDPI fails to recognize SSH in ICMP


vpiserchia:
Provided the great work you're doing, thanks for sharing with the community.

During test runs, we've found a detection mismatch with the online available pcap: 

`
# ndpiReader -i c37c0d3084675ed9b9d63a4e5e50e8da.pcap 

-----------------------------------------------------------
* NOTE: This is demo app to show *some* nDPI features.
* In this demo we have implemented only some basic features
* just to show you what you can do with the library. Feel 
* free to extend it and send us the patches for inclusion
------------------------------------------------------------

Using nDPI (2.9.0) [1 thread(s)]
Reading packets from pcap file /bcache0/icmp_tunneling/c37c0d3084675ed9b9d63a4e5e50e8da.pcap...
Running thread 0...

nDPI Memory statistics:
        nDPI Memory (once):      203.62 KB    
        Flow Memory (per flow):  2.08 KB      
        Actual Memory:           2.74 MB      
        Peak Memory:             2.74 MB      
        Setup Time:              1367 msec
        Packet Processing Time:  1 msec

Traffic statistics:
        Ethernet bytes:        211522        (includes ethernet CRC/IFC/trailer)
        Discarded bytes:       4998         
        IP packets:            863           of 961 packets total
        IP bytes:              190810        (avg pkt size 198 bytes)
        Unique flows:          1            
        TCP Packets:           0            
        UDP Packets:           0            
        VLAN Packets:          0            
        MPLS Packets:          0            
        PPPoE Packets:         0            
        Fragmented Packets:    0            
        Max Packet size:       1041         
        Packet Len < 64:       100          
        Packet Len 64-128:     158
        Packet Len 128-256:    533
        Packet Len 256-1024:   11
        Packet Len 1024-1500:  61
        Packet Len > 1500:     0
        nDPI throughput:       665.38 K pps / 1.22 Gb/sec
        Analysis begin:        07/Feb/2013 10:04:26
        Analysis end:          07/Feb/2013 10:23:08
        Traffic throughput:    0.77 pps / 1.47 Kb/sec
        Traffic duration:      1122.515 sec
        Guessed flow protos:   0

Detected protocols:
        ICMP                 packets: 863           bytes: 190810        flows: 1

Protocol statistics:
        Acceptable                  190810 bytes
`

The same pcap generates alerts in suricata with ET Open rulesets installed:

`
[...]

PR #1464, opened 2022-03-01
Add ICMP checksum check and set risk if mismatch detected.

utoni:


src/lib/ndpi_main.c
```diff
@@ -3030,6 +3030,15 @@ u_int16_t ndpi_guess_protocol_id(struct ndpi_detection_module_struct *ndpi_str,
 	    if (NDPI_ENTROPY_ENCRYPTED_OR_RANDOM(flow->entropy) != 0) {
 	      ndpi_set_risk(ndpi_str, flow, NDPI_SUSPICIOUS_ENTROPY);
 	    }
+
+	    struct ndpi_icmphdr * const icmphdr = (struct ndpi_icmphdr *)packet->payload;
+	    u_int16_t hdr_chksm = icmphdr->checksum;
+	    icmphdr->checksum = 0;
```
utoni: The packet buffer will be modified, but the old state is restored later on.
It is required to calculate the correct checksum. I've decided against copying the whole packet buffer in to a new buffer before modification for the sake of performance.

IvanNardi:
Curiosity: is that feature somehow related to #719?

utoni:
> Curiosity: is that feature somehow related to #719?

It is indeed. But it does not fix the SSH in ICMP tunnel detection in a "stable" way so I've decided against any reference to #719. :)

src/lib/ndpi_utils.c
```diff
@@ -2295,6 +2295,26 @@ float ndpi_entropy(u_int8_t const * const buf, size_t len) {
   return entropy;
 }
 
+/* ******************************************************************** */
+
+u_int16_t ndpi_calculate_icmp_checksum(u_int8_t const * const buf, size_t len) {
```
IvanNardi: I didn't checked, but this algorithm seems identical to the TCP/UDP ones. **If** that is true, should we use a more generic name for the function (especially since we export it to the public API)?

src/lib/ndpi_main.c
IvanNardi: Not very fond of changing (even temporarily) the real packet, but I do understand the performance goal. Let's try this way anyway

utoni: Solved those two issues by modifying the checksum calculation. Killing two birds with one stone. There is no need anymore for any buffer modification and the name of the function should make more sense now as the calculation is much more different to other ones.

IvanNardi (approved):

PR #1469, opened 2022-03-02
Drop support for non-gcrypt builds.

utoni:
* As there is now a builtin, lightweight libgcrypt
   there is no need to disable tls-clho decryption.
 * It is still possible to use a host libgcrypt
   with `--with-local-libgcrypt'.

Signed-off-by: Toni Uhlig <matzeton@googlemail.com>

PR #1470, opened 2022-03-02
Follow up on using internal crypto code

IvanNardi:
Two possible configurations:
* using internal crypto library (default)
* using external libgrypt (via `--with-libgcrypt`). In that case, if
libgrypt/libgpg-error are not available (or if it is too old) an error
is returned.

Update CI configuration to tests both.

Bottom line: QUIC flows are always decoded.

The define `HAVE_LIBGCRYPT` should be set only when the external
libgcrypt is used.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -240,25 +240,9 @@ AC_CHECK_LIB(pthread, pthread_setaffinity_np, AC_DEFINE_UNQUOTED(HAVE_PTHREAD_SE
 
 AS_IF([test "${with_libgcrypt+set}" = set],[
 dnl> GCRYPT
-GCRYPT_ENABLED=1
-AC_ARG_ENABLE([gcrypt],
-  [AS_HELP_STRING([--disable-gcrypt], [Avoid compiling with libgcrypt/libgpg-error, even if they are present. QUIC sub-classification may be missing])],
-  [GCRYPT_ENABLED=0],
-  [AC_CHECK_LIB(gcrypt, gcry_cipher_checktag)])
-AS_IF([test ${GCRYPT_ENABLED} -eq 1], [
-  if test "x$ac_cv_lib_gcrypt_gcry_cipher_checktag" = xyes; then :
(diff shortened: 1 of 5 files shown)


src/lib/protocols/quic.c
```diff
@@ -246,21 +241,14 @@ static uint16_t gquic_get_u16(const uint8_t *buf, uint32_t version)
 }
 
 
-#if defined(HAVE_LIBGCRYPT)
-
 #ifdef DEBUG_CRYPT
 char *__gcry_err(gpg_error_t err, char *buf, size_t buflen)
 {
-#if defined(HAVE_LIBGPG_ERROR) || defined(LIBGCRYPT_INTERNAL)
```
utoni, line 254: forgot to adapt that line in #1469

configure.ac
```diff
@@ -240,25 +240,9 @@ AC_CHECK_LIB(pthread, pthread_setaffinity_np, AC_DEFINE_UNQUOTED(HAVE_PTHREAD_SE
...
-    if test "x$ac_cv_lib_gcrypt_gcry_cipher_checktag" = xyes -a "x$ac_cv_lib_gpg_error_gpg_strerror_r" = xyes; then :
-      ADDITIONAL_LIBS="${ADDITIONAL_LIBS} -lgcrypt -lgpg-error"
-    else
-      GCRYPT_ENABLED=0
-    fi
-  fi
-])
+  AC_CHECK_LIB(gpg-error, gpg_strerror_r, [], [AC_MSG_ERROR([libgpg-error required (because of --with-libgcrypt) but not found or too old.])])
+  AC_CHECK_LIB(gcrypt, gcry_cipher_checktag, [], [AC_MSG_ERROR([libgcrypt required (because of --with-libgcrypt) but not found or too old.])])
+  ADDITIONAL_LIBS="${ADDITIONAL_LIBS} -lgcrypt -lgpg-error"
```
utoni, line 245: Better approach. Will use that slightly modified.

IvanNardi:
#1469 is a more complete solution. Closing.

PR #1474, opened 2022-03-04
Improved ASN/IP update scripts and CI integration.

utoni:


.github/workflows/build.yml
```diff
@@ -43,6 +43,29 @@ jobs:
...
+        run: |
+          echo 'Running ./utils/bitcoinnodes.sh'
+          ./utils/bitcoinnodes.sh >/dev/null
+          echo 'Running ./utils/get_routes_by_asn.sh AS714'
+          ./utils/get_routes_by_asn.sh AS714 >/dev/null
+          echo 'Running ./utils/update_every_lists.sh'
+          ./utils/update_every_lists.sh
+          echo 'Checking for changes in the git tree..'
+          git update-index --refresh || echo "::warning file=utils/update_every_lists.sh::Please re-run utils/update_every_lists.sh and commit any changes."
+          git diff-index --quiet HEAD -- || true
```
IvanNardi, line 67: Is that the best solution? 
I am fine with checking in the CI that the scripts run smoothly and report a message if something is outdated. But forcing the update otherwise the CI fails?
Some of this lists change very often, even daily (TOR). That's mean that anyone pushing even the most trivial PR should update some lists in the PR itself, i.e. we pollute (almost) every commit with completely unrelated stuff.
Maybe we ("we" as maintainers) could manually update the list once in a week/month/before the release/whenever in a separated commit. Or add a note in the README (or configure output) about that feature and let the users update the lists themselves  (if they care) when they compile the code. 
That requires more work, but we haven't updated some of this list for years and I don't know if there is any value on updating them so often.
What do you think?

utoni: > Is that the best solution?

This is definitly not the best solution.

> I am fine with checking in the CI that the scripts run smoothly and report a message if something is outdated. But forcing the update otherwise the CI fails?

For now the CI won't fail, it just prints a warning at the "Files changed" tab. This is also far from perfect. I would rather see a comment or an opened issue. 

> Some of this lists change very often, even daily (TOR). That's mean that anyone pushing even the most trivial PR should update some lists in the PR itself, i.e. we pollute (almost) every commit with completely unrelated stuff. Maybe we ("we" as maintainers) could manually update the list once in a week/month/before the release/whenever in a separated commit.

It should be possible to use something like Github Actions cron job that runs every week/month and creates an issue if something changed (which will be most likely the case).

> Or add a note in the README (or configure output) about that feature and let the users update the lists themselves (if they care) when they compile the code. That requires more work, but we haven't updated some of this list for years and I don't know if there is any value on updating them so often. What do you think?

In my opinion the very best solution would be if nDPI could update those lists w/o recompiling any code. There should be an API that enables every developer to update it during runtime. But that requires some implementation effort.

IvanNardi: > For now the CI won't fail, it just prints a warning at the "Files changed" tab. This is also far from perfect. I would rather see a comment or an opened issue.

Sorry, I misunderstood the code. If the CI doesn't fail, I am ok with this change (and with all your consideration about possible improvements)

IvanNardi (approved):

PR #1478, opened 2022-03-05
Fixed a bug for BE architectures

vel21ripn:
Fixed a bug in the internal implementation of libgcrypt for bigendian architectures


IvanNardi:
Thanks for fixing this issue so promptly!

PR #1482, opened 2022-03-07
Errors fixed

vel21ripn:
Fixed errors for bigendian platforms in ndpiReader.
All address and port comparisons and hash calculations are done with
endian in mind.
The get_ndpi_flow_info() function searched for an existing flow for the
forward and reverse direction of the packet.
The ndpi_workflow_node_cmp() function looked for a flow regardless of
the packet's direction. This is what led to an error in determining the
direction of transmission of the packet.

Fixed error in "synscan" test: the number of packets in the forward and
reverse direction is incorrectly defined (verified via tcpdump).


IvanNardi:
@vel21ripn , could you please push a new version with the following patch?
```
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -305,10 +305,10 @@ jobs:
             "uname -a &&
             lscpu | grep Endian
             "
-      - name: Configure and compile (no tests) using qemu for the specified architecture (s390x - big endian)
+      - name: Configure and compile using qemu for the specified architecture (s390x - big endian)
         if: startsWith(matrix.os, 'ubuntu') && startsWith(matrix.arch, 's390x')
         uses: docker://multiarch/ubuntu-core:s390x-bionic
-        with: #./tests/do.sh disabled because we know we have some problems with big-endian machines
+        with:
           args: >
             bash -c
             "apt-get -y update &&
@@ -318,5 +318,6 @@ jobs:
             make -C example ndpiSimpleIntegration &&
             make -C rrdtool &&
             make -C python &&
+            ./tests/do.sh &&
             ./tests/do-unit.sh

```

IvanNardi (approved):

utoni (approved):

utoni:
Thanks for fixing this long term issue.

PR #1483, opened 2022-03-08
[WIP]  Implement CI on Windows.

aouinizied:
Now that all issues on big endian were solved by talented people and that we finally saw this green light near s390 line, we can move to new issues and a new red line.
This PR is marked as WIP as I need several attempts to make it work but it will add build and test process on **Windows server**.
I already did that within nfstream project and saw several issues related to new lightweight encryption library an so on.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -170,7 +170,7 @@ jobs:
         if: startsWith(matrix.os, 'ubuntu') && startsWith(matrix.arch, 'x86_64') && startsWith(matrix.compiler, 'clang')
         run: |
           sudo apt-get install ${{ matrix.compiler }}
-      - name: Set up MSYS2
+      - name: Setup Windows prerequisites
         if: startsWith(matrix.os, 'windows')
         uses: msys2/setup-msys2@v2
         with:
@@ -212,27 +212,27 @@ jobs:
         run: |
           env CC=${{ matrix.compiler }} CFLAGS='-Werror' ./autogen.sh --enable-option-checking=fatal --enable-debug-messages ${{ matrix.gcrypt }} ${{ matrix.msan }} ${{ matrix.pcre }} ${{ matrix.maxminddb }} --enable-tls-sigs
       - name: Build nDPI
-        if: startsWith(matrix.arch, 'x86_64')
+        if: startsWith(matrix.arch, 'x86_64') & !startsWith(matrix.os, 'windows')
         run: |
           make all
           make -C example ndpiSimpleIntegration
           make -C rrdtool
           make -C python
       - name: Install nDPI
-        if: startsWith(matrix.arch, 'x86_64')
+        if: startsWith(matrix.arch, 'x86_64') & !startsWith(matrix.os, 'windows')
         run: |
           make install DESTDIR=$(realpath _install)
           ls -alhHR _install
       - name: Test nDPI [DIFF]
-        if: startsWith(matrix.arch, 'x86_64')
+        if: startsWith(matrix.arch, 'x86_64') & !startsWith(matrix.os, 'windows')
         run: |
           ./tests/do.sh
       - name: Test nDPI [UNIT]
-        if: startsWith(matrix.arch, 'x86_64')
+        if: startsWith(matrix.arch, 'x86_64') & !startsWith(matrix.os, 'windows')
         run: |
           ./tests/do-unit.sh
       - name: Test nDPI [DGA]
-        if: startsWith(matrix.arch, 'x86_64')
+        if: startsWith(matrix.arch, 'x86_64') & !startsWith(matrix.os, 'windows')
         run: |
           ./tests/do-dga.sh
       - name: Generate tarball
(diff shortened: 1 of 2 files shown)


.github/workflows/build.yml
```diff
@@ -47,9 +47,9 @@ jobs:
     name: ${{ matrix.os }} ${{ matrix.arch }} ${{ matrix.gcrypt }} ${{ matrix.compiler }} ${{ matrix.pcre }} ${{ matrix.maxminddb }} ${{ matrix.msan }}
     runs-on: ${{ matrix.os }}
     strategy:
-      fail-fast: false # Debugging purposes, switch to true when all issues on big endian is solved
+      fail-fast: true # Debugging purposes, switch to true when all issues on big endian is solved
```
utoni:
```suggestion
      fail-fast: true
```

AFAIK the comment is not necessary anymore.

aouinizied: Not sure as we will have issues on windows for sure. So I will update it but I'll not remove it.

aouinizied:
@lnslbrty now It compiles for both libgcrypt and internal one. 
There is a lot of warnings for both of them. We have two options:
* Treat warning as errors and once all warnings during compilation step are treated, move to test.
* Enable testing, and then (it will fail for sure) start debugging.

Note also that ndpiSimpleIntegration compilation step is disabled as it fails too, rrd stuff too.

aouinizied:
@IvanNardi @lnslbrty what do you think is the best option? 
Maybe we should merge in the actual state and then people can enable/disable according to the progress.

utoni:
1. merge it right away
2. merge it, but let the job fail until some takes pity to solve the problem
3. Is it possible to let the job neither succeed nor fail? Like a warning or so? In real life there are always different shades of grey between black and white. CI's should be like that.

.github/workflows/build.yml
```diff
@@ -195,7 +221,15 @@ jobs:
       - name: Configure nDPI on MacOS
         if: startsWith(matrix.os, 'macOS') && startsWith(matrix.arch, 'x86_64') && startsWith(matrix.compiler, 'default-cc')
         run: |
-          env CC=clang CFLAGS='-Werror' ./autogen.sh --enable-option-checking=fatal --enable-debug-messages ${{ matrix.gcrypt }} ${{ matrix.msan }} ${{ matrix.pcre }} ${{ matrix.maxminddb }} --enable-tls-sigs
+          env CC=clang CFLAGS='-Werror' ./autogen.sh --enable-option-checking=fatal --enable-debug-messages ${{ matrix.gcrypt }} ${{ matrix.msan }}
```
IvanNardi: Why this change on MacOS?

.github/workflows/build.yml
```diff
@@ -170,6 +170,32 @@ jobs:
...
+        with:
+          msystem: MINGW64
+          update: true
+          install: git mingw-w64-x86_64-toolchain automake1.16 automake-wrapper autoconf libtool make mingw-w64-x86_64-json-c mingw-w64-x86_64-crt-git mingw-w64-x86_64-pcre
+      - name: Setup Windows prerequisites (libgcrypt)
+        if: startsWith(matrix.os, 'windows') && startsWith(matrix.gcrypt, '--with-local-libgcrypt')
+        run: |
+          msys2 -c 'git clone --branch libgpg-error-1.42 https://github.com/gpg/libgpg-error'
+          cd libgpg-error
+          msys2 -c './autogen.sh'
```
utoni, line 185: Is it possible to merge it with the `./configure` step e.g.:
```suggestion
          msys2 -c './autogen.sh --enable-maintainer-mode --enable-static --enable-shared --with-pic --disable-doc --disable-nls'
```
So we can safe one additional `./configure` step.

IvanNardi:
> * Enable testing, and then (it will fail for sure) start debugging.

Just to be sure I fully understand the current situation: ndpi compiles on Windows (with lots of warnings) but the executable "doesn't work": does that mean that some unit tests fail or that it always crashes?

.github/workflows/build.yml
utoni: Sorry, this isn't libndpi...

aouinizied:
@lnslbrty We can merge it as it is. There is warnings but build is passing.

@IvanNardi build pass, but tests do not run (diff).
Based on what I observed on nfstream project these pcaps test fails on Windows:
* ocs.pcap
* quic-mvfst-22_decryption_error.pcap
* http-crash-content-disposition.pcap

IvanNardi:
> @IvanNardi build pass, but tests do not run (diff). Based on what I observed on nfstream project these pcaps test fails on Windows:
> 
>     * ocs.pcap
> 
>     * quic-mvfst-22_decryption_error.pcap
> 
>     * http-crash-content-disposition.pcap

Thanks for the clarification. Could you share a link with the logs, please?

IvanNardi (approved):
I think we can merge this PR in the current state and we can fix the issues step-by-step later. Exactly like we did for the big-endian stuff

PR #1488, opened 2022-03-15
ndpiReader leaks memory when to free idle flows

wofanli:
Looks like some memory happens in live-capture scenario. Some metadata is leaked during flow freeing.


utoni:
Could you please share the ./configure parameters and ndpiReader command line arguments which lead to this leakage? Thanks.

wofanli:
@lnslbrty 
Actually, I integrate the ndpi_reader into my own app. I found this leak lies on some typo within my code.  
Forget about this PR.

PR #1490, opened 2022-03-16
Reducing the size of the ndpi_detection_module_struct structure.

vel21ripn:
The ndpi_detection_module_struct structure contains 5 arrays "struct
ndpi_call_function_struct" size 286*144=41 kB size,
which are occupied by a small number of elements.

At the moment we have callback_buffer_size = 172, tcp_with_payload=114,
tcp_no_payload=8, udp=93, other 8.
NDPI_MAX_SUPPORTED_PROTOCOLS = 285.

Size of struct ndpi_detection_module_struct is 253136 bytes.
Size of all structs ndpi_call_function_struct 5*286*144=205920 bytes.
Real use memory size for struct ndpi_call_function_struct is
(173+224)*144=57168 bytes.

### DIFF
--- a/src/include/ndpi_typedefs.h
+++ b/src/include/ndpi_typedefs.h
@@ -991,20 +991,17 @@ struct ndpi_detection_module_struct {
   void *user_data;
 #endif
   char custom_category_labels[NUM_CUSTOM_CATEGORIES][CUSTOM_CATEGORY_LABEL_LEN];
+
   /* callback function buffer */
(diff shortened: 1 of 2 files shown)


utoni (approved):
LGTM. I think this is a good memory-performance trade-off. The additional *deref for every dissection routine called may not a performance issue although not verified.

IvanNardi (approved):
Fine for me

IvanNardi:
Thanks!

PR #1491, opened 2022-03-16
Fixed msys2 build and re-activated CI Mingw-w64 build.

utoni:
* Removed Visual Studio leftovers. Maintaining an autotools project with VS integration requires some additional overhead.


aouinizied:
@lnslbrty amazing work!
Just some remarks based on what I already tested:
* Keep libgcrypt and libgperror (clone build from source etc.) steps as these are not available targets (didn't succeed to find them at least).
* Also no need for libpcap as it is already on Windows runners with WinPcap.
* For rrd, python, pcre and maxmind, I guess we want an initial support on CI with libndpi compiled with and without libgcrypt and tested. Then we can add stuff for these targets. What do you think?

With your fixes you will be able to add testing. Note that it fails if you keep `--enable-tls-sigs` and I remove it on my local windows build in order to run the following and see pcaps when tests fails.

Hope this helps.

```bash
msys2 -c 'cd tests'
msys2 -c 'do.sh'
```

utoni:
> @lnslbrty amazing work! Just some remarks based on what I already tested:

Thanks! <3

>     * Keep libgcrypt and libgperror (clone build from source etc.) steps as these are not available targets (didn't succeed to find them at least).

According to [msys2 pkgs](https://packages.msys2.org/base/mingw-w64-libgcrypt) it is. But I am not sure if you need to add some special repo components/url somewhere. (similiar to debian based distro's e.g. base/contrib/non-free/etc)

>     * Also no need for libpcap as it is already on Windows runners with WinPcap.

True. Totally forgot about that.

>     * For rrd, python, pcre and maxmind, I guess we want an initial support on CI with libndpi compiled with and without libgcrypt and tested. Then we can add stuff for these targets. What do you think?

I will comment that rrd and python stuff out. Support for libpcre and libmaxmind was there before.

> With your fixes you will be able to add testing. Note that it fails if you keep `--enable-tls-sigs` and I remove it on my local windows build in order to run the following and see pcaps when tests fails.

Interesting. It shouldn't. Will investigate.

> Hope this helps.
> 
> ```shell
> msys2 -c 'cd tests'
> msys2 -c 'do.sh'
> ```

It does indeed. Thanks!

utoni:
@aouinizied 
The CI shows some pcap diff's, but looking at those diff's it seems that there aren't any. Is this a classy '\r\n' issue?

IvanNardi:
> @aouinizied The CI shows some pcap diff's, but looking at those diff's it seems that there aren't any. Is this a classy '\r\n' issue?

I briefly looked at https://github.com/ntop/nDPI/runs/5612568828?check_suite_focus=true and it seems that some diffs are real.
For example:
* `4in4tunnel.pcap` and `4in6tunnel.pcap` are clearly different: it seems that the decapsulation code is not triggered
* `1kxun.pcap` flow 1: the URL field is one byte longer: the last bytes were "656" and are "6569" now
* `netflix.pcap` flows 10, 13, 16: the URL field is one byte longer

src/lib/protocols/kerberos.c
```diff
@@ -200,17 +200,21 @@ static int krb_decode_asn1_blocks_skip(struct ndpi_detection_module_struct *ndpi
...
 {
    int dst_len = ndpi_min(src_siz, dst_siz - 1);
 
-   strncpy(dst, src, dst_len);
    dst[dst_len] = '\0';
 
    for(int i = 0; i < dst_len; ++i)
    {
-     dst[i] = tolower(dst[i]);
+     if (ndpi_isprint(src[i]) == 0)
```
utoni, line 212: Not sure how we should deal with non printable characters in a printable string. The issue could be an encoding related one or a diff executable one. But printing binary data in a text terminal is in general not a good idea can lead to different results on different platforms. This caused `tests/pcap/kerberos_fuzz.pcap` and `tests/pcap/telnet.pcap` to fail.

We could implement an "escape" function that prints something e.g. "\\x90" if a buffer contains 0x90.
I've just saw, that my solution is pretty much the same as in `src/lib/protocols/dns.c:494`.

src/lib/protocols/telnet.c
```diff
@@ -99,7 +99,19 @@ static int search_telnet_again(struct ndpi_detection_module_struct *ndpi_struct,
...
     if(packet->packet_direction == 0) /* client -> server */ {
       if(flow->protos.telnet.character_id < (sizeof(flow->protos.telnet.username)-1))
-	flow->protos.telnet.username[flow->protos.telnet.character_id++] = packet->payload[i];
+      {
+        if (i>=packet->payload_packet_len-2 &&
+            (packet->payload[i] == '\r' || packet->payload[i] == '\n'))
+        {
+          continue;
+        }
+        else if (ndpi_isprint(packet->payload[i]) == 0)
```
utoni, line 108: same as src/lib/kerberos.c:212

src/lib/ndpi_utils.c
```diff
@@ -2433,3 +2433,36 @@ void ndpi_set_tls_cert_expire_days(struct ndpi_detection_module_struct *ndpi_str
...
+
+int ndpi_vsnprintf(char * str, size_t size, char const * format, va_list va_args)
+{
+#ifdef WIN32
+  if (str == NULL || size == 0 || format == NULL)
+  {
+    return -1;
+  }
+
+  int ret = vsnprintf_s(str, size, _TRUNCATE, format, va_args);
```
utoni, line 2447: The issue here is a Windoze one. The Windows c library does not obey C99. `snprintf` does not guarantee a `NUL` terminated string if the output buffer is not big enough for the formatted string. That results in test diff's e.g. for `tests/pcap/1kxun.pcap`.

I thought about using the Mingw-w64 builtin, but decided against it, because that will also reduce portability for e.g. MSVC (if we ever want to do that).

utoni:
The last two remaining diff's are `tests/pcap/4in4tunnel.pcap` and `tests/pcap/4in6tunnel.pcap`.
As I am not able to use a decent Windows debugger, I am pretty much lost.
@IvanNardi

IvanNardi:
> The last two remaining diff's are `tests/pcap/4in4tunnel.pcap` and `tests/pcap/4in6tunnel.pcap`. As I am not able to use a decent Windows debugger, I am pretty much lost. @IvanNardi

Is it possible that the macro/define `IPPROTO_IPIP` is not defined in Windows (like `IPPROTO_SCTP` if I am not wrong)?

utoni:
> > The last two remaining diff's are `tests/pcap/4in4tunnel.pcap` and `tests/pcap/4in6tunnel.pcap`. As I am not able to use a decent Windows debugger, I am pretty much lost. @IvanNardi
> 
> Is it possible that the macro/define `IPPROTO_IPIP` is not defined in Windows (like `IPPROTO_SCTP` if I am not wrong)?

It was indeed. Thanks!

.github/workflows/build.yml
```diff
@@ -347,6 +341,7 @@ jobs:
             bash -c
             "apt-get -y update &&
             apt-get -y install git autoconf automake libtool pkg-config gettext libjson-c-dev flex bison libpcap-dev libgcrypt20-dev libpcre3-dev libmaxminddb-dev rrdtool librrd-dev &&
+            git config --global --add safe.directory $(realpath .) &&
```
IvanNardi, line 365: A curiosity: do you know why this line:
a) is needed in the first place?
b) is needed only on same target?
c) has been triggering some errors (on dev) since today (and not before)?

utoni: a) probably due to CVE-2022-24765
b) I guess that the ubuntu-core docker container has other permissions in the source/build folder (e.g. not owned by the user, who executes this bash script)
c) the git ubuntu package was updated on 8th of april with the patch for CVE-2022-24765 (the containers ubuntu-latest and ubuntu-core may have different update cycles, dunno)

IvanNardi: Great reference! Thank you very much!

configure.ac
```diff
@@ -7,7 +7,7 @@ AM_INIT_AUTOMAKE([foreign subdir-objects])
 
 AC_PREFIX_DEFAULT(/usr)
 
-EXTRA_TARGETS="example tests tests/unit tests/dga"
+EXTRA_TARGETS="example tests tests/dga"
```
IvanNardi, line 10: Is it correct to remove `tests/unit`?

utoni: The `tests/unit` will only be added to `EXTRA_TARGETS` if json-c was found via `pkg-config` and can be successfully used for linkage. See lines 102-104.

IvanNardi: Thanks!

aouinizied:
@utoni Impressive work. Thanks for this contribution.

PR #1497, opened 2022-03-25
Improved ASN.1 parsing for Keberos. Fixes #1492.

### LINKED ISSUE #1492: Unexpected Characters in Kerberos output from ndpi/ndpiReader
## Describe the bug
I have a pcap that I feed into a flow sensor that leverages the ndpi library for dissection. When dissecting some kerberos traffic I see some unexpected non alphanumeric characters within some of the fields.

### Expected behavior
I expect printable alphanumeric characters to be contained within the kerberos fields within the ndpi flow struct
### Obtained behavior
There are unexpected characters in the output of some Kerberos fields. In this
field, only printable, alphanumeric characters were expected but other symbols
were observed present.

00000510: 2c30 2c2c 302c 2c2c 2c2c 7465 7374 6265  ,0,,0,,,,,testbe
00000520: 6431 2e63 615c 39e2 b7fa a275 fe64 0f78  d1.ca\9....u.d.x
00000530: d1f2 ae31 ccb1 fcb1 325f c4c6 2c30 3b30  ...1....2_..,0;0

Note the characters after '.ca' in the above dump (output is from -dev).

* OS name: Ubuntu 
* OS version: 20.04
* Architecture: x86_64
* nDPI version or commit hash: 4.0, 4.2, dev we use 4.0 primarily but tested with 4.2 and dev
.
* nDPI compilation flags used: if you are building from source 
Steps: ./autogen.sh ./configure make & make install (default compilation flags)
* Attach the `config.log` file generated after `./configure` ran (if you are building from source).
*  [config.log](https://github.com/ntop/nDPI/files/8306819/config.log)
This is generated from the building of version 4.0
### Reproducible using ndpiReader? Yes
- [x] The reported bug is reproducible using ndpiReader.

### If applicable, the used ndpiReader options:
Default options

I've attached a zip file below with the pcap in question as well as output from the provided reader command for 4.0 4.2 and dev
Example: kerberos.pcap

### Steps to reproduce the behavior:
Run examples/pcapReader on pcap that contains Kerberos events. This is a subset
of pcap taken from https://www.unb.ca/cic/datasets/ids-2017.html
'ThursdayWorkingHours.pcap'.

1. ./pcapReader -i kerberos.pcap -C kerberos.csv

## Additional context
Output seems to differ across versions which may make sense based on the changelog
[ndpi-bug.zip](https://github.com/ntop/nDPI/files/8306886/ndpi-bug.zip)
Please contact me if more details are required 
email: hleta@liveaction.com
-Henry Leta

utoni:
* This is a quick fix, the Kerberos protocol dissector requires some refactoring effort.


src/lib/protocols/kerberos.c
```diff
@@ -402,6 +617,15 @@ void ndpi_search_kerberos(struct ndpi_detection_module_struct *ndpi_struct,
...
 	      koffsetp = koffset + 4;
+	      if (krb_parse(ndpi_struct, flow, koffsetp) == 0)
+	      {
+	        NDPI_LOG_DBG(ndpi_struct,
+	                     "[TGS-REP][s/dport: %u/%u][Kerberos Hostname,Domain,Username][%s,%s,%s]\n",
+	                     sport, dport, flow->protos.kerberos.hostname, flow->protos.kerberos.domain,
+	                     flow->protos.kerberos.username);
+	        flow->extra_packets_func = NULL;
+	        return;
+	      }
```
IvanNardi, line 690: Let's see if I got that right.
You wrote a new decoder for the TGS-REP message. If your code succeeds, dissection ends. But if your code fails, the existing code is triggered. Is that the expected behaviour?

src/lib/protocols/kerberos.c
```diff
@@ -36,6 +36,221 @@ static int ndpi_search_kerberos_extra(struct ndpi_detection_module_struct *ndpi_
...
+    return -1;
+  }
+
+  length = krb_decode_asn1_string_type(ndpi_struct, &kasn1_offset, &text);
+  if (length < 0)
+  {
+    return -1;
+  }
+
+  kasn1_offset += length;
```
IvanNardi, line 310: Codiga warning...

src/lib/protocols/kerberos.c
utoni: This is a more or less quick fix for #1492. I was pretty much scared about unforseen consequences. That is why the old code is still triggered.

src/lib/protocols/kerberos.c
utoni: Seems like a false-positive to me. But I can change it in a way that the function parameter won't be used as a local scope variable. That should fix it.

utoni:
Further Kerberos refactoring will be part of a follow-up.

IvanNardi (approved):

PR #1499, opened 2022-03-25
Change the way NDPI_API_VERSION generated.

socketpair:


configure.ac
```diff
@@ -73,21 +73,17 @@ if test -d ".git" || test -f ".git" ; then :
...
-     fi
-     NDPI_API_VERSION=`echo $SOURCE_DATE_EPOCH | cut -c7-10`
 fi
 
-NDPI_API_VERSION=`echo $NDPI_API_VERSION | sed 's/^0*//'`
+NDPI_API_VERSION=0x$( cat \
+    src/include/ndpi_typedefs.h \
+    src/include/ndpi_protocol_ids.h \
+    src/include/ndpi_api.h.in \
+    | md5sum | cut -c1-16 \
```
IvanNardi, line 85: Is it better to be extra cautious here and add all the headers in `src/include/`?
For example, `ndpi_define.h.in` and `ndpi_patricia_typedefs.h`are definitely needed, not sure about the others...

example/ndpiSimpleIntegration.c
```diff
@@ -1108,7 +1109,7 @@ int main(int argc, char ** argv)
   printf("usage: %s [PCAP-FILE-OR-INTERFACE]\n"
 	 "----------------------------------\n"
 	 "nDPI version: %s\n"
-	 " API version: %u\n"
+	 " API version: %" PRIu64 "\n"
```
utoni, line 1112: `PRIu64` may not work for all Windows CCs.
Better use `%llu` and cast (if necessary) the format param to `(unsigned long long int)`.

configure.ac
utoni: Am I right that by replacing the time based approach with a checksum based one, the API version is now not constantly increasing? That may break (sooner or later) applications which use nDPI's dev branch for compilation.

utoni (changes requested):

configure.ac
socketpair: It always has been random. (by number of lines in files). So I did not make it worse.

example/ndpiSimpleIntegration.c
socketpair: How is it possible it may not work? It's the standard. No? Also, what is Windows CC ?

configure.ac
socketpair: 1. @IvanNardi Please give me comprehensive list of files to include.
2. @lnslbrty Time-based logic does not work in presence of branches and backports. Or, we should use "epoch" increased by hand. Actually, this is not needed because backward-compatibiliy is absent, and also API versions are never compared in the code. (only == and not < or >).

example/ndpiSimpleIntegration.c
utoni: [MSVC wants %I64u :/](https://docs.microsoft.com/en-us/previous-versions/visualstudio/visual-studio-6.0/aa261215(v=vs.60)?redirectedfrom=MSDN)

`%llu` should be the most portable way to print 64-bit uint's

configure.ac
utoni: > Actually, this is not needed because backward-compatibiliy is absent, and also API versions are never compared in the code. (only == and not < or >).


The API version may not be used by libnDPI itself. But other developers may use it for their own applications. The current API version is pretty much obsolete. But making it even more obsolete shouldn't be the goal, instead I would pledge to remove the API version.

socketpair: Yes, removing API version looks OK to me.

utoni:
I am closing this PR due to inactivity.
There is also a decision to make:

1. keep the API_VERSION as it is
2. remove API_VERSION w/o replacement
3. increase the size of the API_VERSION, but let it constantly increase (may not work as API_VERSION generation differs for Git-Sources and Dist-Tarballs)
4.  increase the size of the API_VERSION and use a hash based approach (basically this PR)
5. remove API_VERSION and use semantic versioning (the method I would prefer, but results in some additional overhead)

PR #1504, opened 2022-03-27
Fix CI tests results

IvanNardi:
CI integration is failing since 856d7d2.


IvanNardi:
@lucaderi , @aouinizied, I am not sure if that is a proper fix for the CI failures, but I am going to merge it to have again meaningful CI results...

PR #1508, opened 2022-03-31
ndpireader: add json output back.

vitalivanov:
- partial revert of:
    commit 51cfdfb0d80a7bbcc11bc3b95d1696d8dae900c2
    Author: Luca Deri <deri@ntop.org>
    Date:  Sun Nov 17 17:51:45 2019 +0100
        Removed unused JSON-C code

- Json option is changed from 'j' to 'k' as it's used in the new codebase.
- use HAVE_LIBJSON_C instead of HAVE_JSON_C.
- tabs vs spaces clean ups.

Signed-off-by: Vitaliy Ivanov <vitaliyi@interfacemasters.com>

### DIFF
--- a/example/ndpiReader.c
+++ b/example/ndpiReader.c
@@ -54,6 +54,9 @@
 #include <sys/stat.h>
 #include <fcntl.h>
 #include <libgen.h>
+#ifdef HAVE_LIBJSON_C
+#include <json.h>
+#endif
 
 #include "reader_util.h"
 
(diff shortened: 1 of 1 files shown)


vitalivanov:
@lucaderi this is for 4.2-stable.
i'm preparing one for dev also

lucaderi:
Good work but for the future please avoid changing the indentation as the diff file is huge

PR #1509, opened 2022-03-31
ndpireader: add json output back.

vitalivanov:
- partial revert of:
    commit 51cfdfb0d80a7bbcc11bc3b95d1696d8dae900c2
    Author: Luca Deri <deri@ntop.org>
    Date: Sun Nov 17 17:51:45 2019 +0100
      Removed unused JSON-C code

- Json option is changed from 'j' to 'k' as it's used in the new codebase.
- use HAVE_LIBJSON_C instead of HAVE_JSON_C.
- tabs vs spaces clean ups.

Signed-off-by: Vitaliy Ivanov <vitaliyi@interfacemasters.com>

 Conflicts:
	example/ndpiReader.c

### DIFF
--- a/example/ndpiReader.c
+++ b/example/ndpiReader.c
@@ -54,6 +54,9 @@
 #include <sys/stat.h>
 #include <fcntl.h>
 #include <libgen.h>
+#ifdef HAVE_LIBJSON_C
+#include <json.h>
+#endif
 
 #include "reader_util.h"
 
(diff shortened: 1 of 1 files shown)


lucaderi:
Thanks you

IvanNardi:
This patch broke the `ndpiReader` output when "-v2" option is used.
That means that unit tests are now completely broken.
There are also some (trivial) compilation errors that need to be fixed.
@vitalivanov , could you take a look, please?

lucaderi:
I am also looking at it as I have realized it too

lucaderi:
@vitalivanov I have fixed the dev branch, please send a PR for the stable branch

vitalivanov:
> please send a PR for the stable branch
ok i will

@lucaderi 
@IvanNardi 

a.
4 for Potentially overflowing call to snprintf
1 for Local variable hides global variable

these are part of old npireader json patch...

b.
tabs vs spaces - yes, agree, but reading that was a pain...

c. v2 - if luca addressed this in dev - I would update to 4.2-stable.
it still works with v4:)

PR #1516, opened 2022-04-18
Kerberos: fix Undefined-shift error

IvanNardi:
Similar to the error fixed in 4775be3d
Found by oss-fuzz.
See: https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=46713


utoni (approved):
my fault

PR #1518, opened 2022-04-19
Improved sflow protocol detection false-positives.

utoni:

### DIFF
--- a/src/lib/protocols/sflow.c
+++ b/src/lib/protocols/sflow.c
@@ -35,10 +35,19 @@ void ndpi_search_sflow(struct ndpi_detection_module_struct *ndpi_struct, struct 
   if((packet->udp != NULL)
      && (payload_len >= 24)
      /* Version */
-     && (packet->payload[0] == 0) && (packet->payload[1] == 0) && (packet->payload[2] == 0)
-     && ((packet->payload[3] == 2) || (packet->payload[3] == 5))) {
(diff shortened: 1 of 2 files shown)


src/lib/protocols/sflow.c
```diff
@@ -35,10 +35,20 @@ void ndpi_search_sflow(struct ndpi_detection_module_struct *ndpi_struct, struct
   if((packet->udp != NULL)
      && (payload_len >= 24)
      /* Version */
-     && (packet->payload[0] == 0) && (packet->payload[1] == 0) && (packet->payload[2] == 0)
-     && ((packet->payload[3] == 2) || (packet->payload[3] == 5))) {
-    NDPI_LOG_INFO(ndpi_struct, "found sflow\n");
-    ndpi_set_detected_protocol(ndpi_struct, flow, NDPI_PROTOCOL_SFLOW, NDPI_PROTOCOL_UNKNOWN, NDPI_CONFIDENCE_DPI);
+     && (ntohl(get_u_int32_t(packet->payload, 0)) == 0x00000002 ||
```
utoni: There is no pcap available for that version. Maybe remove it?

IvanNardi: SFlow versions different from 5 are deprecated. I have not been able to find a v2 trace.
I am not a user of SFlow stuff, but I think we can remove it

IvanNardi (approved):

PR #1521, opened 2022-04-20
Improved whatsapp detection/dissection.

utoni:


src/lib/protocols/whatsapp.c
```diff
@@ -36,6 +81,25 @@ void ndpi_search_whatsapp(struct ndpi_detection_module_struct *ndpi_struct,
...
+    ndpi_set_detected_protocol(ndpi_struct, flow, NDPI_PROTOCOL_WHATSAPP, NDPI_PROTOCOL_UNKNOWN, NDPI_CONFIDENCE_DPI);
+    return;
+  }
+
+  if (packet->payload_packet_len >= 32 &&
+      ntohs(get_u_int16_t(packet->payload, 0)) == 0xc2fe &&
+      packet->payload[3] == 0x05 &&
+      ntohl(get_u_int32_t(packet->payload, 8)) == 0x00020016 &&
+      packet->payload[16] == 0x08)
+  {
```
IvanNardi, line 96: I have some doubts about that signature... could you give me a few days to run some tests before merging, please?

utoni: Sure, this signature is based on flows captured in a period of ~3 months. But could be still false-positive deficient.

IvanNardi: I don't think this signature matches WA traffic, but some kind of Xiaomi proprietary protocol, used (only) in Xiaomi phones.
To check that, I took a Xiaomi MiA3 phone, factory reset it and capture the traffic at the boot, without any WA (or Meta) apps installed: I found that pattern (note the DNS request)
[xiaomi_mia3.zip](https://github.com/ntop/nDPI/files/8542925/xiaomi_mia3.zip)

I checked that pattern in some live traffic I have access to (with a lots of different users), and the user agents and domains names *always* point to Xiaomi 

As you found out, it seems this traffic contains some kind of domain name, a user agent, the radio access technology and the APN. Not sure about the format, though.

What do you think?

utoni: Thanks for the investigation. My assumption was only port based. A simple `whois` and `nslookup` would be enough to realize that the traffic is not related to Meta.

What now? Create a new protocol or ignore such packets?

IvanNardi: > My assumption was only port based.

Yes. This traffic use the same port used by WA

If no one bets me on time, I'll push a PR with a new, generic,  Xiaomi protocol

PR #1522, opened 2022-04-21
Add some scripts to easily update some IPs lists

IvanNardi:
Follow-up of 8b062295

Add a new protocol id for generic Tencent/Wechat flows


utoni (approved):
Currently fixing `utils/whatsapp_ip_addresses_download.sh` so it will use the correct URL automatically.

PR #1523, opened 2022-04-21
Reduce ndpiReader's `-h' spam. `-H' does this job now.

utoni:


IvanNardi (approved):
Small change but very.. helpful :)

PR #1529, opened 2022-04-24
XIAOMI: add detection of Xiaomi traffic

IvanNardi:
Most of the credits should go to @utoni (see #1521)


utoni:
Can you merge your pcap with the one from PR #1521? I've just wanna know if the dissection works as expected. :)

PR #1536, opened 2022-04-30
Improved AES-NI check.

utoni:
* A library should not open a subshell


IvanNardi (approved):
e4318ffc2d3b9b7f481bf1fd9aac713218e65bd9 introduced a ***huge*** performance drop (on ubuntu 20.04, x86_64, at least).
This patch seems to restore the old performance.
Therefore, even if I am not able to test it on different environments, I am going to merge it

PR #1537, opened 2022-05-02
Improved suspicious http user agent detection.

utoni:


src/lib/protocols/http.c
```diff
@@ -420,19 +420,44 @@ static void ndpi_http_parse_subprotocol(struct ndpi_detection_module_struct *ndp
...
+      if (isupper(ua[i]) != 0)
+      {
+        upper_case_count++;
+      }
+    }
+
+    if (i == ua_len)
+    {
+      float upper_case_ratio = (float)upper_case_count / (float)ua_len;
+      if (upper_case_ratio >= 0.2f)
```
IvanNardi, line 452: Empiric value?

utoni: observation of some malware traffic, but not yet empiric

PR #1540, opened 2022-05-05
Added Edgecast and Cachefly CDNs.

utoni:


src/lib/ndpi_content_match.c.inc
```diff
@@ -717,6 +717,31 @@ static ndpi_network host_protocol_list[] = {
   { 0x3353EED8 /* 51.83.238.216/30 */, 30, NDPI_PROTOCOL_ANYDESK },
   { 0x3353EEDC /* 51.83.238.220/31 */, 31, NDPI_PROTOCOL_ANYDESK },
 
+  /*
+   * Edgecast CDN
+   * Uses captcha to prevent web scraping.
```
IvanNardi, line 722: A captcha to protect a list of 20 addresses, last updated 8 years ago.... :(

src/lib/ndpi_content_match.c.inc
```diff
@@ -717,6 +717,31 @@ static ndpi_network host_protocol_list[] = {
...
+   * https://support.globaldots.com/hc/en-us/articles/115004003849-Edgecast-IP-addresses
+   */
```
IvanNardi, line 724: Reading that page, it seems to me these addresses are used only from Edgecast servers to connect to the origin.
In other words, if I am right, a client connecting to a site served by this CDN does not connect to one of this addresses...

Probably the real list we want is https://bgp.he.net/AS15133#_prefixes (and that seems confirmed by https://developer.edgecast.com/cdn/api/Content/Media_Management/Get-CDN-IP-Blocks.htm)

Anyway, this list might be useful anyway.

src/lib/ndpi_main.c
```diff
@@ -1877,6 +1878,14 @@ static void ndpi_init_protocol_defaults(struct ndpi_detection_module_struct *ndp
                           "Xiaomi", NDPI_PROTOCOL_CATEGORY_WEB,
                           ndpi_build_default_ports(ports_a, 0, 0, 0, 0, 0) /* TCP */,
                           ndpi_build_default_ports(ports_b, 0, 0, 0, 0, 0) /* UDP */);
+  ndpi_set_proto_defaults(ndpi_str, 0 /* encrypted */, NDPI_PROTOCOL_ACCEPTABLE, NDPI_PROTOCOL_EDGECAST,
+                          "Edgecast", NDPI_PROTOCOL_CATEGORY_WEB,
```
IvanNardi: `NDPI_PROTOCOL_CATEGORY_CLOUD` like all the other CDN?

src/lib/ndpi_main.c
```diff
@@ -1877,6 +1878,14 @@ static void ndpi_init_protocol_defaults(struct ndpi_detection_module_struct *ndp
...
+                          ndpi_build_default_ports(ports_a, 0, 0, 0, 0, 0) /* TCP */,
+                          ndpi_build_default_ports(ports_b, 0, 0, 0, 0, 0) /* UDP */);
+  ndpi_set_proto_defaults(ndpi_str, 0 /* encrypted */, NDPI_PROTOCOL_ACCEPTABLE, NDPI_PROTOCOL_CACHEFLY,
+                          "Cachefly", NDPI_PROTOCOL_CATEGORY_WEB,
```
IvanNardi: As above

utils/asn_update.sh
```diff
@@ -12,104 +10,137 @@ function processing_list() {
...
 function create_list() {
+	RETVAL=0
 	LIST=/tmp/list
 
 	for i in "${@:3}"; do
-		./get_routes_by_asn.sh "$i" >> $LIST
+		if ! ./get_routes_by_asn.sh "$i" >> $LIST; then
+			echo "Could not fetch route for ${i} (${1})"
+			RETVAL=$(( RETVAL + $? ))
+		fi
```
IvanNardi, line 23: You removed  all the ASs without any prefixes.
Do we have any guarantees that these ASs don't start announcing some prefixes in the next future? If so, this change make sense. Otherwise I think it is more "future proof" to keep them and keep polling them

utils/asn_update.sh
```diff
@@ -12,104 +10,137 @@ function processing_list() {
...
 
-echo "(1) Downloading Hotspot Shield routes..." #AnchorFree
-DEST=../src/lib/inc_generated/ndpi_asn_hotspotshield.c.inc
-create_list NDPI_PROTOCOL_HOTSPOT_SHIELD $DEST "AS26642"
-echo "(3) Hotspot Shield IPs are available in $DEST"
+#echo "(1) Downloading Hotspot Shield routes..." #AnchorFree
+#DEST=../src/lib/inc_generated/ndpi_asn_hotspotshield.c.inc
+#create_list NDPI_PROTOCOL_HOTSPOT_SHIELD $DEST "AS26642"
+#RETVAL=$(( RETVAL + $? ))
+#echo "(3) Hotspot Shield IPs are available in $DEST"
```
IvanNardi: As above

src/lib/ndpi_content_match.c.inc
utoni: agreed

src/lib/ndpi_content_match.c.inc
utoni: I have only skimmed that page, unfortunately. You're right, AS15133 seems promising.

utils/asn_update.sh
utoni: The goal was to let the script fail if e.g. ASN update does not work anymore. I will re-add them and look for another solution.

IvanNardi (approved):
Thanks for the changes!

PR #1546, opened 2022-05-10
Improved Xiaomi HTTP detection.

utoni:
* Merged Xiaomi pcap files


IvanNardi:
How did you merge the two pcaps (with different snapshot lengths), creating a new one that ndpiReader is able to open? Plain mergecap didn't work for me in the past...

utoni:
I've used `mergecap` to merge those pcaps, but indeed `ndpiReader` did not like the result.
So I tried to open it with `wireshark` and exported it again as pcap.
It seems that when merging different pcap (formats? versions?) it can result in pcaps that can not be used for `ndpiReader`.
Not sure if the issue is related to `mergecap` or `ndpiReader` (or more precisely `libpcap`).

IvanNardi:
> So I tried to open it with `wireshark` and exported it again as pcap.

Thanks for that advice!!

IvanNardi (approved):

PR #1548, opened 2022-05-12
DNS-over-QUIC: update default port

IvanNardi:
Final RFC (https://www.rfc-editor.org/rfc/rfc9250.txt) changed the
default port to 853 (from the 784 used on the early drafts).
Since there is at least one deployed implementation still stuck at
draft-02, keep also the old value, for the time being.


src/lib/ndpi_main.c
```diff
@@ -1154,7 +1154,7 @@ static void ndpi_init_protocol_defaults(struct ndpi_detection_module_struct *ndp
   ndpi_set_proto_defaults(ndpi_str, 0 /* encrypted */, NDPI_PROTOCOL_FUN, NDPI_PROTOCOL_DOH_DOT,
 			  "DoH_DoT", NDPI_PROTOCOL_CATEGORY_NETWORK /* dummy */,
 			  ndpi_build_default_ports(ports_a, 853, 0, 0, 0, 0) /* TCP */,
-			  ndpi_build_default_ports(ports_b, 784, 0, 0, 0, 0) /* UDP */);
+			  ndpi_build_default_ports(ports_b, 853, 0, 0, 0, 0) /* UDP */);
```
utoni: Isn't better to keep the old port together with the new one? I guess older protocol versions can be still seen in the wild.

utoni (approved):

PR #1549, opened 2022-05-16
Some maintenance tasks.

utoni:


IvanNardi (approved):
I never noticed the trace in the wrong place...

utoni:
Took me a while as well. I did not even notice it while reviewing your PR.

PR #1552, opened 2022-05-25
Fixed windows-latest build error.

utoni:

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -280,7 +280,7 @@ jobs:
       - name: Configure nDPI on Windows
         if: startsWith(matrix.os, 'windows') && startsWith(matrix.arch, 'x86_64') && startsWith(matrix.compiler, 'default-cc')
         run: |
-          msys2 -c 'env CFLAGS="-Werror" ./autogen.sh --enable-option-checking=fatal --enable-debug-messages --enable-tls-sigs --disable-npcap ${{ matrix.gcrypt }}'
+          msys2 -c 'env CFLAGS=\"-Werror -Wuse-after-free=1\" ./autogen.sh --enable-option-checking=fatal --enable-debug-messages --enable-tls-sigs --disable-npcap ${{ matrix.gcrypt }}'
       - name: Build nDPI on Windows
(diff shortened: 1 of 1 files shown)


.github/workflows/build.yml
```diff
@@ -280,7 +280,7 @@ jobs:
       - name: Configure nDPI on Windows
         if: startsWith(matrix.os, 'windows') && startsWith(matrix.arch, 'x86_64') && startsWith(matrix.compiler, 'default-cc')
         run: |
-          msys2 -c 'env CFLAGS="-Werror" ./autogen.sh --enable-option-checking=fatal --enable-debug-messages --enable-tls-sigs --disable-npcap ${{ matrix.gcrypt }}'
+          msys2 -c 'env CFLAGS=\"-Werror -Wno-use-after-free\" ./autogen.sh --enable-option-checking=fatal --enable-debug-messages --enable-tls-sigs --disable-npcap ${{ matrix.gcrypt }}'
```
IvanNardi: I think we should add something like "-Wuse-after-free=1" (the default seems to be 2) to CFLAGS in configure.ac generic code. Not sure, how, thought.
This error/warnings is not related to Windows, but to gcc-12. I am able to trigger it on ubuntu with gcc-12.

BTW, do you agree that this is a false positive warnings?

utoni: You're right. My manpage did not state anything about `-Wuse-after-free=1`. I was using gcc-11 and forgot to upgrade...

Indeed, I think this is a false positive at least if we assume a POSIX compatible `realloc()`.

utoni: Seems to work, @IvanNardi

IvanNardi: Sorry if I was not clear. What about setting `-Wuse-after-free=1` in `configure.ac` (and not in `.github/workflows/build.yml`) for (kind of) every build? Is that even possible? This way we don't get the warnings whenever we use gcc-12.
(This issue is not specific to windows or to the CI)

utoni: It is possible, but requires lot more work.
My hope is that this false-positive will be fixed ASAP, so we do not need that flag anymore.

IvanNardi: Ok, it is reasonable!

IvanNardi (approved):

PR #1555, opened 2022-05-26
Added MPEG-DASH dissector. Fixes #1223.

### LINKED ISSUE #1223: add signature to identify Mpeg-DASH(Dynamic Adaptive Streaming over HTTP) application
mpeg-DASH is one of video streaming application. I am planning to add signature to identify DASH,
I had tried to create new dissector of dash i was following steps in https://github.com/ntop/nDPI#how-to-add-a-new-protocol-dissector but NDPI still don't know Dash,
please can any one help me to add new signature DASH

utoni:
* Improved HTTP POST detection
 * Refactored subprotocol detection


utoni:
Subprotocol handling for HTTP is weird..
Most of the time HTTP is a master protocol, but in some rare cases it is an app protocol.

src/lib/ndpi_main.c
```diff
@@ -5037,6 +5045,47 @@ void ndpi_connection_tracking(struct ndpi_detection_module_struct *ndpi_str,
...
+    if ((ndpi_str->callback_buffer[subproto_index].ndpi_selection_bitmask & ndpi_selection_packet) ==
+         ndpi_str->callback_buffer[subproto_index].ndpi_selection_bitmask &&
+        NDPI_BITMASK_COMPARE(flow->excluded_protocol_bitmask,
+                             ndpi_str->callback_buffer[subproto_index].excluded_protocol_bitmask) == 0 &&
+        NDPI_BITMASK_COMPARE(ndpi_str->callback_buffer[subproto_index].detection_bitmask,
+                             detection_bitmask) != 0)
+    {
+      ndpi_str->callback_buffer[subproto_index].func(ndpi_str, flow);
+      num_calls++;
+    }
```
IvanNardi, line 5081: Why we don't "break" from the "for" loop if we got a classification (like in the original code)?

src/lib/protocols/mpegdash.c
```diff
@@ -0,0 +1,96 @@
...
+  return;
+}
+
+void init_mpegdash_dissector(struct ndpi_detection_module_struct *ndpi_struct,
+                             u_int32_t *id, NDPI_PROTOCOL_BITMASK *detection_bitmask)
+{
+  ndpi_set_bitmask_protocol_detection("MpegDash", ndpi_struct, detection_bitmask, *id,
+				      NDPI_PROTOCOL_MPEGDASH,
+				      ndpi_search_mpegdash_http,
+				      NDPI_SELECTION_BITMASK_PROTOCOL_V4_V6_TCP_WITH_PAYLOAD,
```
IvanNardi: `NDPI_SELECTION_BITMASK_PROTOCOL_V4_V6_TCP_WITH_PAYLOAD_WITHOUT_RETRANSMISSION`?

src/lib/ndpi_main.c
utoni: The issue with the loop break is that it creates a kind of dissector hierarchy.
But that was the exact issue I've wanted to fix with the sub-protocol check in the first place.

More specific: Protocols set with `ndpi_set_proto_subprotocols()` would had have a hierarchy again which already exists for "normal" protocol detection/dissection, basically a first-come-first-serve kind of thing depending on their protocol id. But that can lead to false-positives/negatives.

IvanNardi: Sorry but I don't get it: don't you have change the logic from "first-come-first-serve" to "the last match win"? Isn't that some kind of "dissector hierarchy" anyway?

But the main point I don't understand is: why a protocol hierarchy is fine with the "standard" classification but not for the "sub-protocol" classification (i.e. mainly/only the HTTP sub-classification)? Do you have any examples triggering any issue?

As usual, my main concern is performance: `check_ndpi_detection_func()` is already one of the hottest functions; I'd like to avoid calling more callbacks, if not necessary. Obviously, not problem from me if we need to...

utoni: I probably need to rephrase my thoughts so it may be more clear.

A common detection phase works like this:
The classification routines are called in the order they are in the callback buffer.
After the first classification routine set's a detected protocol, no more routines are invoked from the callback buffer.
So the hierarchy I was talking about is the index of the protocol callback in the callback buffer. A lower index in that buffer will dominate every higher index.
But there are some protocols that are embedded in another protocol e.g. mpeg-dash is usually transferred over http(s).

Some classification routines are now called after a protocol was already detected as long as they are not excluded (and were previously set as possible sub-protocol via `ndpi_set_proto_subprotocols()`).
The system is still far from perfection e.g. dissectors that rely on packet numbers of a flow may have a negative performance impact (and can still result in false positives/negatives).

I hope that I was able to reduce the "noise" of my previous statement. :)

IvanNardi: Thanks, and I agree with your general description.
But I don't get this point:
* with the previous code (with the break), the final classification was determined by the first match (i.e. the first matching protocol of the sub list)
*  without the break, the final classification is determined by the last match

I don't know which solution is the best one; but it seems to me that the classification results still depends on the sub-protocols list order.
Do you agree with that?

utoni: >     * with the previous code (with the break), the final classification was determined by the first match (i.e. the first matching protocol of the sub list)

Exactly.

>     * without the break, the final classification is determined by the last match

Not necessarily. If a protocol sets a classification, others are still possible to overwrite that as long as they did not exclude themselves. That way the decision for a classification is delegated to the dissector instead the core.

IvanNardi: Ok, I am not entirely convinced, but this patch is an improvements compare to the present code.
Let's merge it and we will face any fallback in the future (if any)
Thanks

IvanNardi (approved):

PR #1556, opened 2022-05-27
Provide a generic reassembler interface.

utoni:
* Shall be used for stream based protocols e.g. TCP, QUIC, etc.


utoni:
Will be part of a new PR.

PR #1557, opened 2022-05-28
Fixed HTTP lower/upper protocol mess for Aimini/IPP.

utoni:
* If HTTP was already detected, there is no need to overwrite the detection again


IvanNardi (approved):
Thanks for fixing this issue!

IvanNardi:
A similar patch might be needed for mpegdash (the attached session is classified as MpegDash/HTTP insted of HTTP/MpegDash).
Unfortunately a trivial fix in `ndpi_int_mpegdash_add_connection()` triggers a regression in the first flow of `mpeg-dash.pcap`: MpegDash.AmazonAWS -> HTTP.AmazonAWS (the other 2 flows are fine)

[mpeg-dash-2.zip](https://github.com/ntop/nDPI/files/8793404/mpeg-dash-2.zip)

PR #1561, opened 2022-05-29
Fix dissection of IPv4 header

IvanNardi:
See: https://github.com/ntop/nDPI/runs/6643914510?check_suite_focus=true

Convert al the `MIN(a,b)` calls to `ndpi_min(a,b)`


src/lib/ndpi_main.c
```diff
@@ -4664,7 +4664,7 @@ static u_int8_t ndpi_detection_get_l4_internal(struct ndpi_detection_module_stru
 
   /* 0: fragmented; 1: not fragmented */
   if(iph != NULL && ndpi_iph_is_valid_and_not_fragmented(iph, l3_len)) {
-    u_int16_t len = ntohs(iph->tot_len);
+    u_int16_t len = MIN(ntohs(iph->tot_len), l3_len);
```
utoni: We should probably merge those `MIN`s with `ndpi_min`.

utoni (approved):

PR #1563, opened 2022-05-30
Certificate timestamps should be printed in UTC

IvanNardi:
This way, the CI results are consistent


utoni:
Thanks! I did not know what was going on because the CI output did not help me and I was not able to reproduce it on my local machine.

IvanNardi:
> Thanks! I did not know what was going on because the CI output did not help me and I was not able to reproduce it on my local machine.

It take some times to me to figure it out, since it was not reproducible in my machine neither. And I needed to triple check before seeing the real differences reported by the CI

utoni (approved):

PR #1570, opened 2022-05-31
Removed LGTM ql query for packet payload integer arithmetic.

utoni:
* Detection is unreliable and results in too much false-positives (and false-negatives!)


IvanNardi:
I couldn't agree more.
I'll happily merge this PR

IvanNardi (approved):

PR #1577, opened 2022-06-03
Fixed syslog false positives.

utoni:
* syslog: removed unnecessary/unreliable printable string check
 * added `ndpi_isalnum()`
 * splitted `ndpi_is_printable_string()` into `ndpi_is_printable_buffer()` and `ndpi_normalize_printable_string()`


IvanNardi:
Do you have any example of false positive with the old code?
AFAIK, all the flows in the removed `syslog.pcapng` file were real syslog messages.
BTW, with the new code and the new `syslog.pcap` 2 sessions are classified only "by-port", even if it seems they are real syslog flows

src/include/ndpi_utils.h
```diff
@@ -25,6 +25,7 @@ extern u_int8_t ndpi_ends_with(char *str, char *ends);
 
 #define ndpi_isalpha(ch) (((ch) >= 'a' && (ch) <= 'z') || ((ch) >= 'A' && (ch) <= 'Z'))
 #define ndpi_isdigit(ch) ((ch) >= '0' && (ch) <= '9')
+#define ndpi_isalnum(ch) (ndpi_isalpha(ch) != 0 && ndpi_isdigit(ch) != 0)
```
utoni:
```suggestion
#define ndpi_isalnum(ch) (ndpi_isalpha(ch) != 0 || ndpi_isdigit(ch) != 0)
```

fixes the issue ;)

utoni:
> Do you have any example of false positive with the old code? AFAIK, all the flows in the removed `syslog.pcapng` file were real syslog messages. BTW, with the new code and the new `syslog.pcap` 2 sessions are classified only "by-port", even if it seems they are real syslog flows

I accidentally deleted the `rsh` false-positive unfortunately. 👎 
`syslog.pcap` is a merge of the old `syslog.pcapng` and some newer flows.

IvanNardi:
This is an example of a flow classified "by-dpi" with the old code and "by-port" with the new one.
[syslog.zip](https://github.com/ntop/nDPI/files/8833280/syslog.zip)
Could you take a look, please?

IvanNardi:
> This is an example of a flow classified "by-dpi" with the old code and "by-port" with the new one. [syslog.zip](https://github.com/ntop/nDPI/files/8833280/syslog.zip) Could you take a look, please?

Thanks!

IvanNardi (approved):

PR #1582, opened 2022-06-04
Fixed syslog false negatives.

utoni:


src/lib/protocols/syslog.c
```diff
@@ -107,7 +108,7 @@ void init_syslog_dissector(struct ndpi_detection_module_struct *ndpi_struct, u_i
   ndpi_set_bitmask_protocol_detection("Syslog", ndpi_struct, detection_bitmask, *id,
 				      NDPI_PROTOCOL_SYSLOG,
 				      ndpi_search_syslog,
-				      NDPI_SELECTION_BITMASK_PROTOCOL_V4_V6_TCP_OR_UDP_WITH_PAYLOAD_WITHOUT_RETRANSMISSION,
+				      NDPI_SELECTION_BITMASK_PROTOCOL_V4_V6_UDP_WITH_PAYLOAD,
```
utoni: In theory, syslog over TCP is possible. But in practice, I've never seen such aliens.

IvanNardi: According to https://twitter.com/webernetz/status/1401883154298908673 it shoudn't be so rare and that is a pcap example [syslog_over_tcp.zip](https://github.com/ntop/nDPI/files/8839546/syslog_over_tcp.zip)
Anyway, I agree with your proposal: remove TCP support and let's fight this battle another day, only if really needed.

IvanNardi: Could you add a comment here about why TCP is missing, please? Just as a reminder....

IvanNardi (approved):
Could you remove 514 as default port for Syslog for TCP from `ndp_main.c`, please?

utoni:
> Could you remove 514 as default port for Syslog for TCP from `ndp_main.c`, please?

But isn't 514 the default port for Syslog?

src/lib/protocols/syslog.c
utoni: I will revert to UDP&TCP detection and try something else.

IvanNardi: As you like. I am fine with both solutions (Syslog TCP+UDP or only UDP)

IvanNardi:
> > Could you remove 514 as default port for Syslog for TCP from `ndp_main.c`, please?
> 
> But isn't 514 the default port for Syslog?

UDP, yes. No official TCP default port, AFAIK. (and 514 is reserved for RSH).

utoni:
I've decided to ignore the Syslog false positive for RSH, because this will only affect midstream flows under certain and very rare conditions.

PR #1586, opened 2022-06-06
Dissect host line if SSDP contains such.

utoni:


IvanNardi:
@utoni , could you add something like "SSDP host" to the list in the comment about `struct ndpi_flow_struct->host_server_name` in `src/include/ndpi_typedefs.h`, please?

IvanNardi (approved):

PR #1594, opened 2022-06-13
DNS: fix TTL check and sync unit test results

IvanNardi:


src/lib/protocols/dns.c
```diff
@@ -299,7 +299,7 @@ static int search_valid_dns(struct ndpi_detection_module_struct *ndpi_struct,
 	  } else
 	    x += data_len;
 
-	  if((x+2) >= packet->payload_packet_len) {
+	  if((x+8) >= packet->payload_packet_len) {
```
utoni, line 306: Shouldn't be `x+6` sufficient or do I miss something?

IvanNardi: No, we need 8. From `x`, we read the type (2 bytes); ignore the class (2 bytes) and read the ttl (4 bytes).
Please note that `get16()` increments `x` by 2 internally...

utoni: Yep, I did ignore this DNS specific `get16()`.

utoni (approved):

PR #1601, opened 2022-06-16
Added collectd dissector (again).

utoni:


IvanNardi:
@utoni , all the attached flows are classified as collectd; is that correct or are they false positives?
[collectd.zip](https://github.com/ntop/nDPI/files/8927970/collectd.zip)

utoni:
false positives, need to increase `COLLECTD_MIN_BLOCKS_REQUIRED` or instead of setting a malformed packet risk, just exclude it immediately =)

IvanNardi:
> false positives, need to increase `COLLECTD_MIN_BLOCKS_REQUIRED` or instead of setting a malformed packet risk, just exclude it immediately =)

Thanks for the improvements!

src/lib/protocols/collectd.c
```diff
@@ -0,0 +1,209 @@
...
+
+  if (num_blocks < COLLECTD_MIN_BLOCKS_REQUIRED)
+  {
+    NDPI_EXCLUDE_PROTO(ndpi_struct, flow);
+    return;
+  }
+
+  if (hostname_length > 0 &&
+      ndpi_int_collectd_dissect_hostname(flow, packet, hostname_offset,
+                                         hostname_length) != 0)
```
IvanNardi, line 189: Just a nit: `ndpi_int_collectd_dissect_hostname` never fails, because `ndpi_hostname_sni_set` never returns NULL; so this check is useless.
Up to you if removing the check or not; I am fine in both cases

IvanNardi (approved):

src/lib/protocols/collectd.c
utoni: I know, but I think it makes sense to extend `ndpi_hostname_sni_set` at some point e.g. so that it will set a risk if a non-printable character is seen.

PR #1603, opened 2022-06-17
gprof test/CI integration

utoni:
CI test, not ready to merge


utoni:
The CI shows that `ndpi_load_malicious_sha1_file` wastes a lot of memory by using an AhoCorasick automata, which is useful for sub string matching, which we do not need for a simple sha1 lookup. I will replace that with a memory saving hash map approach asap.

IvanNardi (approved):

IvanNardi:
@utoni, could you rebase also this one, please?

PR #1605, opened 2022-06-18
Improved SOAP via HTTP.

utoni:


src/lib/protocols/soap.c
```diff
@@ -37,6 +44,28 @@ void ndpi_search_soap(struct ndpi_detection_module_struct *ndpi_struct,
...
+      {
+        ndpi_int_soap_add_connection(ndpi_struct, flow);
+        return;
+      }
+    }
+
+    NDPI_EXCLUDE_PROTO(ndpi_struct, flow);
+    return;
+  }
+
```
IvanNardi, line 65: Shouldn't **all** this new code be within an `if (flow->detected_protocol_stack[0] == NDPI_PROTOCOL_HTTP)` kind of check?

utoni: True, but the http dissector has an issue detecting some POSTs. I can try to fix it here or in a follow-up.

IvanNardi: As you prefer... I noticed that because the attached flow is no more classified as SOAP
[soap3.zip](https://github.com/ntop/nDPI/files/8933722/soap3.zip)

utoni: I forgot that the packet line parser just searches for `\r\n` ...
`if (flow->detected_protocol_stack[0] == NDPI_PROTOCOL_HTTP)` would be the better solution, but needs some http dissector fixes e.g. handling HTTP POSTs like the one `tests/pcap/soap.pcap`.

IvanNardi (approved):

PR #1615, opened 2022-07-01
Added TunnelBear VPN detection patterns.

utoni:


mmanoj:
@utoni 

Do we need to add Packet communication patterns, like from first packet bytes, etc? or are DNS entries enough?

utoni:
@mmanoj 

If you want to detected the DNS flow which may occur either before or not at all due to DoH/DoT, yes.
But more important is the VPN communication itself, be it via TLS, OpenVPN or something proprietary.

mmanoj:
@utoni 
Thanks for the clarifications, I captured Ultraserf communication, and I will upload it to the original communication thread.

PR #1621, opened 2022-07-03
Fix category for mail sessions

### LINKED ISSUE #629: Incorrect flow category
nDPI performs category detection with the rule "app protocol first". In the image below, the app protocol is Google, so nDPI determines "Web" (Google category) as the flow category, but this is incorrect since the correct category should be "Email".

IvanNardi:
Close #629


utoni:
Thanks! I was encountering this inconsistency as well.

utoni (approved):

PR #1632, opened 2022-07-03
Add two new confidence values: confidence by partial DPI

IvanNardi:
Used for all classifications based on partial/incomplete DPI
information, i.e. all classifications done in `ndpi_detection_giveup()`.


src/include/ndpi_typedefs.h
```diff
@@ -829,7 +829,9 @@ typedef enum {
   NDPI_CONFIDENCE_UNKNOWN = 0,		/* Unknown classification */
   NDPI_CONFIDENCE_MATCH_BY_PORT,	/* Classification obtained looking only at the L4 ports */
   NDPI_CONFIDENCE_MATCH_BY_IP,		/* Classification obtained looking only at the L3 addresses */
-  NDPI_CONFIDENCE_DPI_CACHE,		/* Classification results based on same LRU cache (i.e. correlation among sessions) */
+  NDPI_CONFIDENCE_DPI_PARTIAL,		/* Classification results based on partial/incomplete DPI information */
+  NDPI_CONFIDENCE_DPI_PARTIAL_CACHE,	/* Classification results based on some LRU cache with partial/incomplete DPI information */
+  NDPI_CONFIDENCE_DPI_CACHE,		/* Classification results based on some LRU cache (i.e. correlation among sessions) */
```
utoni, line 834: I am always asking this question to myself when I stumble on these LRUs: How are LRU caches supposed to work in a multi-threaded environment? It basically means loss of DPI detection, right?

utoni (approved):

src/include/ndpi_typedefs.h
IvanNardi: In general, yes.
However, most of these caches use (or should use...) the "client ip" as primary element of the key.
If you balance your traffic in a way that all the traffic of the same "client ip" is elaborated from the same thread, you should not loss DPI detection.

Having all these caches per-thread means that nDPI performance scale linearly (i.e. optimally) with the number of available cores: this is an extreme important feature if you handle tons of trafffic

utoni: Thanks for clarification.

PR #1636, opened 2022-07-05
Fix handling of NDPI_UNIDIRECTIONAL_TRAFFIC risk

IvanNardi:


IvanNardi:
Not a perfect solution, but a step in the right direction (hopefully)

utoni (approved):

src/include/ndpi_main.h
```diff
@@ -150,6 +150,8 @@ extern "C" {
   void ndpi_set_risk(struct ndpi_detection_module_struct *ndpi_str,
 		     struct ndpi_flow_struct *flow, ndpi_risk_enum r,
 		     char *risk_message);
+  void ndpi_unset_risk(struct ndpi_detection_module_struct *ndpi_str,
```
utoni: Not sure if we should export that function. It should be used with caution.

IvanNardi: Ok, let 's make it private. We can make it public again later, if necessary

PR #1637, opened 2022-07-05
Detect SMTPs w/ STARTTLS as TLS and dissect client/server hello. Fixe…

### LINKED ISSUE #1630: nDPI multi-layer identification
Im trying to to do multi-level protocol detection with nDPI.  For example, with SMTPS, currently I am only able to get NDPI to classify the top-level as SMTP.Google and it disregards the subsequent TLS handshake even though I see it processing those packets.  If I remove the beginning SMTP handshake from the pcap, then nDPI sees the TLS.  Any ideas on how to achieve both SMTP.Google and TLS/JA3 classification?  Sample pcap attached.  
[smtp-starttls.zip](https://github.com/ntop/nDPI/files/9035175/smtp-starttls.zip)

utoni:
…s #1630.

 * more packets need to be processed -> bad
 * FTP has to get updated as well as it has similiar STARTTLS semantics -> follow-up

// not a perfect solution, but sufficient to fix #1630 

### DIFF
--- a/src/lib/ndpi_main.c
+++ b/src/lib/ndpi_main.c
@@ -8034,7 +8034,9 @@ u_int8_t ndpi_extra_dissection_possible(struct ndpi_detection_module_struct *ndp
   switch(proto) {
   case NDPI_PROTOCOL_TLS:
   case NDPI_PROTOCOL_DTLS:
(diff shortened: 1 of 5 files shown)


lucaderi:
Looks good as a solution. Not sure if TLS.Google is better than SMPTS.Google, this is my only concern.

PR #1638, opened 2022-07-05
Run regression tests from different locations at the same time w/o si…

utoni:
…de effects on the results.


IvanNardi (approved):
Extremely useful!

PR #1644, opened 2022-07-06
SMB: add (partial) support for messages split into multiple TCP segments

IvanNardi:


utoni (commented):
Do you have a PCAP triggering this code path by any chance?

IvanNardi:
> Do you have a PCAP triggering this code path by any chance?

Unfortunately, I can't share the trace I have because it contains sensitive domain(s).
Let me check if I am able to find another one...

IvanNardi:
> > Do you have a PCAP triggering this code path by any chance?
> 
> Unfortunately, I can't share the trace I have because it contains sensitive domain(s). Let me check if I am able to find another one...

Found it

utoni (approved):

PR #1648, opened 2022-07-07
Avoid spurious calls to extra dissection

IvanNardi:
If the extra callabck is not set, calling the extra dissection is only a
waste of resources...


utoni (approved):
I am in general not a big fan of `ndpi_extra_dissection_possible()`, because it should be possible to achieve the same result with `extra_packets_func()`. It decreases readability and increases confusion.

IvanNardi:
> I am in general not a big fan of `ndpi_extra_dissection_possible()`, because it should be possible to achieve the same result with `extra_packets_func()`. It decreases readability and increases confusion.

I agree. We should be able to have something like (from a logical POV)
```
int ndpi_extra_dissection_possible()
{
  return flow->extra_packets_func() != NULL;
}
```

PR #1649, opened 2022-07-07
LDAP: rewrite dissection

IvanNardi:


src/lib/protocols/ldap.c
```diff
@@ -35,67 +35,67 @@ static void ndpi_int_ldap_add_connection(struct ndpi_detection_module_struct *nd
   ndpi_set_detected_protocol(ndpi_struct, flow, NDPI_PROTOCOL_LDAP, NDPI_PROTOCOL_UNKNOWN, NDPI_CONFIDENCE_DPI);
 }
 
+static int asn1_ber_length(const unsigned char *payload, int payload_len, u_int16_t *value_len)
```
utoni: Remembered that I've wrote a similar function for Kerberos some time ago. Why not use that one instead of creating code duplicates?

utoni (changes requested):

src/lib/protocols/ldap.c
IvanNardi: It makes sense...

utoni (approved):
:+1:

PR #1650, opened 2022-07-07
SMTP: add support for X-ANONYMOUSTLS command

IvanNardi:


utoni (approved):
All the `mail_*` dissectors are in need for refactoring - reduce and simplify.

PR #1652, opened 2022-07-08
Add support for Memory sanitizer

IvanNardi:


utoni (commented):
What exactly does `-fsanitize=memory` do?

IvanNardi:
> What exactly does `-fsanitize=memory` do?

It is another kind of sanitizer tool (other than asan and tsan) which should detect uninitialized reads.
It has some limitations but, AFAIK, it works well with nDPI.
See: https://clang.llvm.org/docs/MemorySanitizer.html
It is supported by clang, no idea about gcc

utoni (approved):

PR #1657, opened 2022-07-10
Keep track of how many dissectors calls we made for each flow

IvanNardi:


utoni:
This is very important patch! I was thinking of implementing something similar. Thanks!

IvanNardi:
> This is very important patch! I was thinking of implementing something similar. Thanks!

I am glad I am not the only one who thinks that these stats are useful!
The only drawback is that we will have a lot of diffs in utests results every time we add a new protocol. But I think the pros outweigh the cons.

utoni (approved):

PR #1662, opened 2022-07-15
Fix unitialized use of W[16].

utoni:


IvanNardi (approved):
Not a fun of this change, since this issue seems to be a false positive triggered only by some "uncommon" compilers.
Anyway, this is obviously the quickest and simplest solution.

utoni:
> Not a fun of this change, since this issue seems to be a false positive triggered only by some "uncommon" compilers. Anyway, this is obviously the quickest and simplest solution.

It adds some small performance overhead, but nDPI will manage that.

PR #1667, opened 2022-07-20
Skype: improvements

IvanNardi:
Skype detection over TCP has been completely disable since https://github.com/ntop/nDPI/commit/659f75138c2a95e5823608a545b9a3d3ced223bc (3
years ago!).
Since that logic was too weak anyway, remove it.

Commit https://github.com/ntop/nDPI/commit/ba6a48c9fec64b0e3617f7f7f1f6afeedd8b6437 is completely bogus: we can't set extra dissection
without having set a proper classification.

The idea behind that commit seems to be that we need to look for 2
(consecutives?) packets with the same crc/pattern: try to implement this
logic in a saner way.


utoni (approved):
We did not even had PCAP which triggers that code path.

PR #1669, opened 2022-07-20
TLS: improve reassembler 

IvanNardi:
The first commit contains only cosmetic changes; the real stuff is in the second one.

We might need to contemporary re-order messages from both directions:
use one buffer per direction.


utoni (approved):
We should think about a general reassembling interface. We have multiple reassembling algorithms for different dissectors. What do you think about that? LGTM anyways

IvanNardi:
> We should think about a general reassembling interface. We have multiple reassembling algorithms for different dissectors. What do you think about that? LGTM anyways

Yeah, we might think about a generic (but lightweight) solution...

Issue #1211, opened 2021-06-18
SYN flood attack mitigation strategy for an app using ndpi


subhajit-cdot:
Hi,
I have gone through the thread #139 , where you suggested that the app using ndpi can skip SYN packet send to ndpi. 
" if you see a SYN you do not need to allocate nDPI memory but just when the first payload packets hits your app."

I have few doubts regarding the handling of SYN packets in the app which uses ndpi.

In the ndpi source I can see the below snippet describing that a protocol identification of a tcp flow may go wrong if SYN packet is not seen. Also, the app using ndpi can cause memory or session table overflow  if we send/handle all tcp connection (including enbryonic connections). To mitigate this kind of attack people suggest to use syn proxy, syn cache syn cookies and syn authentication. You also wrote  ([http://luca.ntop.org/CommoditisingDDoSMitigation.pdf](url)) that "For  SYN  flood  attacks,  two  mitigation  methods  have  been implemented  and  compared:  SYN  Cookie  and  SYN  Sanity Check.  Contrary  to  the  SYN  Sanity  Check  that  does  not challenge  traffic,  the  SYN  Cookie  method  needs  to  inject  at least  one  packet  back  for  every  processed  SYN  in  order  to validate  clients. "  I am keeping syn proxy aside because this is a different solution altogether but here I want to know how can SYN sanity check (assuming it similar to syn authentication) detect ip spoofing or similar ddos attacks? What would you suggest for handling SYN flood type of dos/ddos attack in an app using ndpi?

` if((flow->num_processed_pkts == 1) && (ret.master_protocol == NDPI_PROTOCOL_UNKNOWN) &&
     (ret.app_protocol == NDPI_PROTOCOL_UNKNOWN) && flow->packet.tcp && (flow->packet.tcp->syn == 0) &&
     (flow->guessed_protocol_id == 0)) {
    u_int8_t protocol_was_guessed;
`
`    /*`
`      This is a TCP flow`
`      - whose first packet is NOT a SYN`
`      - no protocol has been detected`
 `     We don't see how future packets can match anything`
`      hence we giveup here`
`    */`
 `   ret = ndpi_detection_giveup(ndpi_str, flow, 0, &protocol_was_guessed);`
`  }`

@lucaderi  please comment!

Thanks

utoni:
I agree with @lucaderi in #139. Ignoring SYN packets should be part of your integration and not of nDPI.

[...]

PR #1671, opened 2022-07-22
Add support for flow client/server information

IvanNardi:
In a lot of places in ndPI we use *packet* source/dest info
(address/port/direction) when we are interested in *flow* client/server
info, instead.

Add basic logic to autodetect this kind of information.

nDPI doesn't perform any "flow management" itself but this task is
delegated to the external application. It is then likely that the
application might provide more reliable hints about flow
client/server direction and about the TCP handshake presence: in that case,
these information might be (optionally) passed to the library, disabling
the internal "autodetect" logic.

These new fields have been used in some LRU caches and in the "guessing"
algorithm.
It is quite likely that some other code needs to be updated.

### DIFF
--- a/src/include/ndpi_typedefs.h
+++ b/src/include/ndpi_typedefs.h
@@ -528,6 +528,31 @@ struct ndpi_vxlanhdr {
 } PACK_OFF;
 
 /* ************************************************************ */
+
+/**
+ * The application might inform the library about client/server direction
+ */
+#define NDPI_IN_PKT_DIR_UNKNOWN		0	/**< The application doesn't provide this kind of information */
+#define NDPI_IN_PKT_DIR_C_TO_S		1	/**< Current packet is from client to server */
+#define NDPI_IN_PKT_DIR_S_TO_C		2	/**< Current packet is from server to client */
+
+/**
+ * The application might choose to not pass TCP handshake packets to the library
+ * (for performance reasons), but it might want to inform the library itlsef that these
+ * packets have been captured/seen anyway (to avoid losing classifiation capabilities).
+ */
+#define NDPI_FLOW_BEGINNING_UNKNOWN	0	/**< The application doesn't provide this kind of information */
+#define NDPI_FLOW_BEGINNING_SEEN	1	/**< The application informs the library that the TCP handshake has been seen (even if its packets might not have been passed to the library) */
+#define NDPI_FLOW_BEGINNING_NOT_SEEN	2	/**< The application informs the library that the TCP handshake has not been seen */
+
+/**
+ * Optional information about flow management (per packet)
+ */
+struct ndpi_flow_input_info {
+  unsigned char in_pkt_dir;
+  unsigned char seen_flow_beginning;
+};
+
 /* ******************* ********************* ****************** */
 /* ************************************************************ */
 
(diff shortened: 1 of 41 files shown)


src/include/ndpi_typedefs.h
```diff
@@ -528,6 +528,20 @@ struct ndpi_vxlanhdr {
...
 
 /* ************************************************************ */
+
+#define NDPI_IN_PKT_DIR_UNKNOWN		0
+#define NDPI_IN_PKT_DIR_C_TO_S		1	/* Client -> Server */
+#define NDPI_IN_PKT_DIR_S_TO_C		2	/* Server -> Client */
+
+#define NDPI_FLOW_BEGINNING_UNKNOWN	0
+#define NDPI_FLOW_BEGINNING_SEEN	1
+#define NDPI_FLOW_BEGINNING_NOT_SEEN	2
```
utoni: Can please give some more information what `UNKNOWN`, `SEEN` and `NOT_SEEN` exactly means?
Any kind of doxygen compatible input appreciated.

IvanNardi: Ok, I'll add a description.

The basic idea behind these values is to allow the application to NOT pass SYNs packets to the library, without losing any classification capabilities (see #1211)
If the application wants (or is able) it can inform the library that the 3 way handshake packets have not been passed to nDPI but they have been received, i.e. the flows has been received from its beginning

I don't like the names very much... if you have hany proposal...

src/include/ndpi_typedefs.h
```diff
@@ -1182,7 +1208,7 @@ struct ndpi_flow_struct {
   /* init parameter, internal used to set up timestamp,... */
   u_int16_t guessed_protocol_id, guessed_host_protocol_id, guessed_category, guessed_header_category;
   u_int8_t l4_proto, protocol_id_already_guessed:1, host_already_guessed:1, fail_with_unknown:1,
-    init_finished:1, setup_packet_direction:1, packet_direction:1, check_extra_packets:1, is_ipv6:1;
+    init_finished:1, client_packet_direction:1, packet_direction:1, check_extra_packets:1, is_ipv6:1;
```
utoni, line 1211: What about `initial_packet_direction` and two macros `NDPI_PKTDIR_CLIENT_RECV` and `NDPI_PKTDIR_CLIENT_SEND` or something similar? Could also used for `packet_direction`. What do you think?

src/include/ndpi_typedefs.h
utoni: > I don't like the names very much... if you have hany proposal...

Together with the documentation it make sense, I guess. Thanks.

src/include/ndpi_typedefs.h
IvanNardi: Sorry but I don't get it; could you elaborate please?

`client_packet_direction` is not the direction of the first/initial packet but the direction of the packets from the client to the server.
Two helpers to get the direction of the current packet are already provided: `ndpi_current_pkt_from_client_to_server()` and `ndpi_current_pkt_from_server_to_client()`

utoni: There was a misunderstanding on my connection end. Sorry. Case closed.

utoni (approved):

PR #1675, opened 2022-07-27
Update the protocol bitmask for some protocols

IvanNardi:
Tcp retransmissions should be ignored.

Remove some unused protocol bitmasks.

Update script to download Whatsapp IP list.


src/lib/protocols/aimini.c
```diff
@@ -262,7 +262,7 @@ void init_aimini_dissector(struct ndpi_detection_module_struct *ndpi_struct, u_i
   ndpi_set_bitmask_protocol_detection("Aimini", ndpi_struct, detection_bitmask, *id,
 				      NDPI_PROTOCOL_AIMINI,
 				      ndpi_search_aimini,
-				      NDPI_SELECTION_BITMASK_PROTOCOL_V4_V6_TCP_OR_UDP_WITH_PAYLOAD,
```
utoni, line 265: Do we even need `NDPI_SELECTION_BITMASK_PROTOCOL_*_TCP_WITH_PAYLOAD` anymore?

IvanNardi: Not really; it is better to remove them to avoid "copy-and-paste" errors in new code.
If we need them again, we will re-add in the future.

utoni (approved):
Interesting change especially the reduction of total dissector calls.

PR #1677, opened 2022-07-27
Patricia tree, Ahocarasick automa, LRU cache: add statistics

IvanNardi:
Add (basic) internal stats to the main data structures used by the
library; they might be usefull to check how effective these structures
are.

Add an option to `ndpiReader` to dump them; disabled by default to avoid
too much fuss with the unit tests.

### DIFF
--- a/example/ndpiReader.c
+++ b/example/ndpiReader.c
@@ -123,6 +123,7 @@ extern u_int8_t enable_doh_dot_detection, enable_ja3_plus;
 extern u_int32_t max_num_packets_per_flow, max_packet_payload_dissection, max_num_reported_top_payloads;
 extern u_int16_t min_pattern_len, max_pattern_len;
 extern void ndpi_self_check_host_match(); /* Self check function */
+u_int8_t dump_internal_stats;
 
 struct flow_info {
   struct ndpi_flow_info *flow;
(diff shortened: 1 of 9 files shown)


example/ndpiReader.c
```diff
@@ -798,7 +799,7 @@ static void parseOptions(int argc, char **argv) {
   }
 #endif
 
-  while((opt = getopt_long(argc, argv, "a:b:e:c:C:dDf:g:i:Ij:k:K:S:hHp:pP:l:r:s:tu:v:V:n:Jrp:x:w:zq0123:456:7:89:m:T:U:",
+  while((opt = getopt_long(argc, argv, "a:Ab:e:c:C:dDf:g:i:Ij:k:K:S:hHp:pP:l:r:s:tu:v:V:n:Jrp:x:w:zq0123:456:7:89:m:T:U:",
```
utoni, line 802: Can you add it to the app usage to give more information about that feature to the user?

utoni (commented):
Would it make sense to include such information in the regression tests output?

IvanNardi:
> Would it make sense to include such information in the regression tests output?

I have mixed feelings about that. These information are useful, especially for the regression tests, no doubts about that.
But I don't like big diffs in the unit test results.
In my experience, big diffs are overlooked (or simply ignored).
However, this ship already sailed with the printing of the "number of dissector calls" info...

No strong opinion. What do you prefer?

example/ndpiReader.c
IvanNardi: Right. I'll do

utoni:
> > Would it make sense to include such information in the regression tests output?
> 
> I have mixed feelings about that. These information are useful, especially for the regression tests, no doubts about that. But I don't like big diffs in the unit test results. In my experience, big diffs are overlooked (or simply ignored). However, this ship already sailed with the printing of the "number of dissector calls" info...
> 
> No strong opinion. What do you prefer?

I would prefer some kind of "once-per-tests.sh-run" statistics. That'll need some more implementation effort, but will reduce the diff. The amount of dissector calls could be also part of such stats. What do you think?

IvanNardi:
> > > Would it make sense to include such information in the regression tests output?
> > 
> > 
> > I have mixed feelings about that. These information are useful, especially for the regression tests, no doubts about that. But I don't like big diffs in the unit test results. In my experience, big diffs are overlooked (or simply ignored). However, this ship already sailed with the printing of the "number of dissector calls" info...
> > No strong opinion. What do you prefer?
> 
> I would prefer some kind of "once-per-tests.sh-run" statistics. That'll need some more implementation effort, but will reduce the diff. The amount of dissector calls could be also part of such stats. What do you think?

For now, let's enable the prints. Later, if we face too  many diffs, we will come up with a solution. Could it work?

IvanNardi:
Merged the wrong version. Reverted

PR #1683, opened 2022-07-29
Patricia tree, Ahocarasick automa, LRU cache: add statistics

IvanNardi:
Add (basic) internal stats to the main data structures used by the
library; they might be usefull to check how effective these structures
are.

Add an option to `ndpiReader` to dump them; enabled by default in the
unit tests.
This new option enables/disables dumping of "num dissectors calls"
values, too (see b4cb14ec).


IvanNardi:
See #1677 for a discussion of this change (sorry for the wrong merge/revert)

utoni (approved):
Let's see how this will affect future regression test diff's.

PR #1685, opened 2022-07-29
SMTPS, POPS, IMAPS: fix classification and extra dissection

IvanNardi:
The big change in TLS code is to allow "master" protocols other than
TLS/DTLS, like SMTPS, POPS and IMAPS.
This change will allow, in a future, a proper and complete TLS dissection
for all these protocols with "STARTTLS"-like messages.


utoni (commented):
We are loosing some detection here. Why is that?

IvanNardi:
> We are loosing some detection here. Why is that?

Could you elaborate it, please? In all the unit tests results we have a better classification and/or more metadata, never less. Am I  missing something?

tests/result/imaps.pcap.out
```diff
@@ -1,29 +1,32 @@
-Guessed flow protos:	0
+Guessed flow protos:	1
```
utoni, line 1: Shouldn't there be a `0` ?

tests/result/pops.pcapng.out
```diff
@@ -1,6 +1,6 @@
-Guessed flow protos:	0
+Guessed flow protos:	1
```
utoni, line 1: same here

tests/result/zoom.pcap.out
```diff
@@ -1,6 +1,6 @@
-Guessed flow protos:	4
+Guessed flow protos:	5
```
utoni, line 1: and `4` here

tests/result/imaps.pcap.out
IvanNardi: The flow is too short, now, to have a "not guessed" classification, because the packets with the certificate are not present.
For IMAPS/POPS/SMTPS flows before this change we only needed the Client Hello (basically), now we need the full TLS handshake, like the "standard" TLS flows. But we don't have it in these 3 traces.

Bottom line: the flows ends (i.e. we don't have more packets) while we are still classifying it.

utoni (approved):

PR #1686, opened 2022-07-29
Fix possible null deref in ndpi_utils:2347

koltiradw:
Hi!
Pointer 'flow' that can have only NULL value (ndpi_get_http_method function), is passed as 2nd parameter in call to function ndpi_set_risk where it is dereferenced at ndpi_utils.c:2347.


utoni:
Can you provide more information under what circumstances that NUL deref may happen? Maybe a PCAP?

IvanNardi:
> Can you provide more information under what circumstances that NUL deref may happen? Maybe a PCAP?

+1

PR #1689, opened 2022-07-30
Improved nDPI JSON serialization.

utoni:

### DIFF
--- a/src/lib/ndpi_utils.c
+++ b/src/lib/ndpi_utils.c
@@ -381,7 +381,7 @@ u_int8_t ndpi_is_safe_ssl_cipher(u_int32_t cipher) {
 
 /* ***************************************************** */
 
-const char* ndpi_cipher2str(u_int32_t cipher) {
+const char* ndpi_cipher2str(u_int32_t cipher, char unknown_cipher[8]) {
   switch(cipher) {
   case TLS_NULL_WITH_NULL_NULL:	return("TLS_NULL_WITH_NULL_NULL");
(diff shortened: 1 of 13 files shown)


src/lib/ndpi_utils.c
```diff
@@ -1302,61 +1425,69 @@ int ndpi_dpi2json(struct ndpi_detection_module_struct *ndpi_struct,
...
     if(flow->http.url != NULL) {
-      ndpi_serialize_string_string(serializer,   "url", flow->http.url);
-      ndpi_serialize_string_uint32(serializer,   "code", flow->http.response_status_code);
-      ndpi_serialize_string_string(serializer,   "content_type", flow->http.content_type);
-      ndpi_serialize_string_string(serializer,   "user_agent", flow->http.user_agent);
+      ndpi_risk_enum risk = ndpi_validate_url(flow->http.url);
+      if (risk != NDPI_NO_RISK)
+      {
+        NDPI_SET_BIT(flow->risk, risk);
+      }
```
IvanNardi, line 1433: I know that you simply cut-and-paste this piece of code, but setting a flow risk in the serialization code (and, before your change, in ndpiReader code) seems very wrong; it should be done in the generic code in the library (http code?).
Anyway, feel free to merge this PR anyway. We will look into this issue as a follow-up

IvanNardi (approved):
I don't use the serialization code, but it LGTM (indipendent issue in the comment)

src/lib/ndpi_utils.c
utoni: True. It was wrong before. This shouldn't be part of ndpiReader as well. I've just pasted it there. A follow-up would be nice.

PR #1690, opened 2022-07-30
TLS: avoid useless extra dissection

IvanNardi:
If we have seen Application Data blocks in both directions, it means we
are after the handshake. At that point, extra dissection is useless.

Useful feature with TLS mid-sessions, i.e. sessions without initial
packets.


utoni (approved):

utoni:
Please rebase with `dev`.

Issue #1687, opened 2022-07-29
HTTP sub-classification: best protocol classification


IvanNardi:
The attached session (extracted from the unit tests) contains a SOAP flow over HTTP with a Microsoft hostname.
How this flow should be classified?

1) SOAP.Microsoft: current result. We lost that information that it is an HTTP flow and so we don't dump HTTP metadata (note that we can have SOAP without HTTP...)
2) HTTP.SOAP
3) HTTP.Microsoft

I don't like the current code because we lose HTTP attribute, but I don't have strong opinions.
What do you think? What is the best classification? @utoni @lucaderi 

Clearly, there is the same issue with all the HTTP sub-protocols.

[soap.zip](https://github.com/ntop/nDPI/files/9222503/soap.zip)

utoni:
> The attached session (extracted from the unit tests) contains a SOAP flow over HTTP with a Microsoft hostname. How this flow should be classified?
> 
>     1. SOAP.Microsoft: current result. We lost that information that it is an HTTP flow and so we don't dump HTTP metadata (note that we can have SOAP without HTTP...)
> 
>     2. HTTP.SOAP
> 
>     3. HTTP.Microsoft
> 
> 
> I don't like the current code because we lose HTTP attribute, but I don't have strong opinions. What do you think? What is the best classification? @utoni @lucaderi
> 
> Clearly, there is the same issue with all the HTTP sub-protocols.
> 
> [soap.zip](https://github.com/ntop/nDPI/files/9222503/soap.zip)

I would probably stick to `HTTP.SOAP`, because `Microsoft` should be still available via `guessed_host_protocol_id`. That way we won't loose information IMHO.

lucaderi:
In my opinion Tony is right but in the short term. For a long term solution i think we need to rethink how protocols are reported. Microsoft, Amazon or Google are not protocols and I have added them years ago to make easy for people to implement things like "How much traffic to I send to Google?". So IMHO IP classification can either be a third-level protocol (e.g.  HTTP.SOAP.Microsoft) or another "attribute" of the traffic similar to breed: in essence they are similar but we need to decide if they are part of the protocol (so HTTP.SOAP.Microsoft) or as an external attribute to HTTP.SOAP.

The first solution is better IMHO but it has the problem of layers. Example: DNS.Microsoft is two layer with respect to HTTP.SOAP.Microsoft that is 3. Due to this we don't have a clean way of identifying what protocol items represent,

IvanNardi:
[...]

PR #1696, opened 2022-08-04
HTTP: improve sub-classification

IvanNardi:
Content-matched sub-protocols (DASH, IPP, MPEGDASH...) shouldn't
ovewrite the previous master protocol (if any; usually HTTP).

Furthermore. the HTTP dissector shouldn't update the classification
(in the extra-dissection code path) if a content-matched sub-protocols
has already been found.

This commit should address the first part of the changes described in #1687.


src/lib/protocols/bittorrent.c
```diff
@@ -105,11 +105,14 @@ static void ndpi_add_connection_as_bittorrent(struct ndpi_detection_module_struc
 					      int bt_offset, int check_hash,
 					      ndpi_confidence_t confidence) {
   struct ndpi_packet_struct *packet = &ndpi_struct->packet;
+  u_int16_t master;
+
+  master = flow->detected_protocol_stack[1] ? flow->detected_protocol_stack[1] : flow->detected_protocol_stack[0];
```
utoni: It is probably more error prone if this line would be a API function.

IvanNardi: Ok, I'll create a new wrapper

utoni (approved):

PR #1698, opened 2022-08-05
Further simplification of `ndpi_process_extra_packet()`

IvanNardi:
See 95e16872.

After c0732eda, we can safely remove the protocol list from
`ndpi_process_extra_packet()`.

The field `flow->check_extra_packets` is redundant; remove it.


utoni (approved):
Satisfying..

PR #1699, opened 2022-08-05
Add TiVoConnect dissector. Fixes #1697.

### LINKED ISSUE #1697: Add TiVo DVR Discovery Support
[TivoDVR.pcap.zip](https://github.com/ntop/nDPI/files/9267381/TivoDVR.pcap.zip)

utoni:


tests/result/gnutella.pcap.out
```diff
@@ -1,15 +1,17 @@
-Guessed flow protos:	597
```
utoni: Not sure what triggered that diff..

IvanNardi: Very strange...I will try to take a look during the weekend

IvanNardi: @utoni : did you trigger the test on dev?
Anyway... I am seeing the same issue with the current dev. No idea how that is possible.... (and no error yesterday)

utoni: @IvanNardi 
yeah and it fails now

utoni: definitely strange since this diff is not related to any external library

IvanNardi: I just re-run the tests on the original #1698 and now they fail (yesterday everything was fine, obviously)

IvanNardi: Test now failing also on c0732eda45884de91e0c221e9dd23eeec364bf68, i.e. "HEAD parent"

utoni: It seems somehow related to one of the LRU caches.

IvanNardi: I tried up to ab3a678ad423fcd431037093220a145925f64148: all latest commits are now failing on dev
Dev also fails on my notebook, so I think we can exclude some updates/changes on Github runners

IvanNardi: @utoni , found the issue!! Unbelievable.... PR is coming

example/reader_util.h
```diff
@@ -234,6 +235,12 @@ typedef struct ndpi_flow_info {
       char hostname[48];
       char fqdn[48];
     } softether;
+    struct {
+      char identity_uuid[36];
+      char machine[48];
+      char platform[48];
+      char services[64];
+    } tivoconnect;
```
IvanNardi, line 243: Could we reduce the size of these fields a little bit?
Right now, the size of `ndpi_flow_info->protos` is 168 (on x86_64; because of tls); with this change it will be 196 (because of tivoconnect).
If possible, I would prefer to avoid increasing (again...) the overall size of `struct ndpi_flow_info`

IvanNardi, line 243: It is not critical, but it s probably better to align these sizes to the ones in `struct ndpi_flow_struct`

example/ndpiReader.c
```diff
@@ -523,6 +523,11 @@ static void help(u_int long_help) {
 #endif
 
   if(long_help) {
+    printf("\n\nSize of nDPI Flow structure: %u\n"
+           "Sizeof of nDPI Flow protocol union: %zu\n",
+           ndpi_detection_get_sizeof_ndpi_flow_struct(),
+           sizeof(((struct ndpi_flow_struct *)0)->protos));
+
```
IvanNardi, line 530: Useful. I think is will be more useful to have the size of the entire `struct ndpi_flow_struct`. Or both?

utoni: Probably better to keep both, because of the two `_Static_assert`s in `ndpi_typedefs.h`.

IvanNardi (approved):

PR #1706, opened 2022-08-09
Improved QUIC IETF 0-RTT detection.

utoni:


IvanNardi:
@utoni , if it is ok with you, I would like to carefully review this change. It may take some days, because it is holiday time in Italy...

BTW, the session you attached is from real traffic?

IvanNardi:
Can I share the session in some (semi) public forums? I would like to ask some questions about this trace

utoni:
> @utoni , if it is ok with you, I would like to carefully review this change. It may take some days, because it is holiday time in Italy...

Take all the time you need. Enjoy your work-free time.

> BTW, the session you attached is from real traffic?

Yes.

utoni:
> Can I share the session in some (semi) public forums? I would like to ask some questions about this trace

Yea, it is already public.

IvanNardi:
I proposed another solution to this issue in #1713. The main advantage is that it correctly decode the Initial/CH received after the 0-RTT, allowing sub-classification
What do you think?

utoni:
Your solution seems better. Mine was just a quick fix for classification. Further dissection did not work.

PR #1708, opened 2022-08-13
Update Makefile.in to use @libdir@ and @includedir@

alcir:
Use `@libdir@` instead of *hardcoded* path, and also `@includedir@`


utoni:
Thank you for your efforts.

utoni (approved):

PR #1723, opened 2022-09-05
Fix string copy truncation warning.

utoni:
protocols/ubntac2.c: In function ‘ndpi_search_ubntac2’:
protocols/ubntac2.c:69:4: warning: ‘strncpy’ output may be truncated copying between 0 and 31 bytes from a string of length 255 [-Wstringop-truncation]
   69 |    strncpy(flow->protos.ubntac2.version, (const char *)version, len);
      |    ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~

### DIFF
--- a/src/lib/protocols/ubntac2.c
+++ b/src/lib/protocols/ubntac2.c
@@ -65,8 +65,8 @@ void ndpi_search_ubntac2(struct ndpi_detection_module_struct *ndpi_struct, struc
 	  
 	  version[j] = '\0';
 
-	  len = ndpi_min(sizeof(flow->protos.ubntac2.version)-1, j);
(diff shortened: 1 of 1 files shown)


IvanNardi (commented):
What compiler/environment triggers this warning?

utoni:
gcc-10 on Debian-11.4

utoni:
In general I would like to replace all `strncpy`s with `memcpy`s as `strncpy` is slower most of the time.

IvanNardi:
> In general I would like to replace all `strncpy`s with `memcpy`s as `strncpy` is slower most of the time.

I couldn't agree more....

PR #1725, opened 2022-09-05
Improved/json serialize ndpi block

utoni:
It makes sense to me to add all detection/dissection data to a object with the key "ndpi".

### DIFF
--- a/src/lib/ndpi_utils.c
+++ b/src/lib/ndpi_utils.c
@@ -1198,7 +1198,6 @@ void ndpi_serialize_proto(struct ndpi_detection_module_struct *ndpi_struct,
 {
   char buf[64];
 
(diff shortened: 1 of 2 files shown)


utoni:
@IvanNardi @cardigliano
Any progress on this tiny change?

IvanNardi:
> @IvanNardi @cardigliano Any progress on this tiny change?

I don't use the serialization code, but it LGTM.
I asked @cardigliano simply because it was the last one to touch that code...

cardigliano:
@IvanNardi on the API side this looks fine for me. I do not exclude that this may break something in ntopng, but we will address that in case. Feel free to accept this.

IvanNardi (approved):

PR #1729, opened 2022-09-10
STUN: several improvements

IvanNardi:
Add detection over TCP and fix detection over IPv6.
Rename some variables since Stun dissector is no more "udp-centric".
Stun dissector should always classified the flow as `STUN` or `STUN/Something`.
Don't touch `flow->guessed_host_protocol_id` field, which should be always be related to "ip-classification" only.


utoni (commented):
`mergecap` the `stun*.pcap` files may be a good idea

IvanNardi:
> `mergecap` the `stun*.pcap` files may be a good idea

I merged most of the traces. I kept `stun_signal.pcapng` since it contains a lot of flows correlated among them (basically, it is a single Signal voice call)

src/lib/protocols/stun.c
```diff
@@ -118,43 +118,11 @@ typedef enum {
...
-  31.13.86.5
-*/
-
-static int is_messenger_ip_address(u_int32_t host) {
-  if(host == 0x1F0D5605 /* 31.13.86.5 */)
-    return(1);
-  else
-    return(0);
-}
-
```
utoni, line 151: no idea what the code had to do here

utoni (approved):
good job, a simplification worthy of support

PR #1731, opened 2022-09-11
Fix `ndpi_do_guess()`

IvanNardi:
Avoid a double call of `ndpi_guess_host_protocol_id()`.
Some code paths work for ipv4/6 both

### DIFF
--- a/src/lib/ndpi_main.c
+++ b/src/lib/ndpi_main.c
@@ -6193,42 +6193,15 @@ static int ndpi_do_guess(struct ndpi_detection_module_struct *ndpi_str, struct n
     }
 
     if(user_defined_proto && flow->guessed_protocol_id != NDPI_PROTOCOL_UNKNOWN) {
-      if(packet->iph) {
-	if(flow->guessed_host_protocol_id != NDPI_PROTOCOL_UNKNOWN) {
-	  u_int8_t protocol_was_guessed;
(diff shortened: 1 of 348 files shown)


src/lib/ndpi_main.c
```diff
@@ -6193,42 +6193,34 @@ static int ndpi_do_guess(struct ndpi_detection_module_struct *ndpi_str, struct n
...
-	flow->guessed_host_protocol_id = ndpi_guess_host_protocol_id(ndpi_str, flow);
+      /*
+	We could implement a shortcut here skipping dissectors for
+	protocols we have identified by other means such as with the IP
 
+	However we do NOT stop here and skip invoking the dissectors
+	because we want to dissect the flow (e.g. dissect the TLS)
+	and extract metadata.
+      */
+#if SKIP_INVOKING_THE_DISSECTORS
```
utoni: I did not know about the the macro. Maybe it would be better to let the user decide with an ndpi preference?

IvanNardi: If we need to touch that code, I'd prefer to delete it altogether... Too extreme?

utoni: Maybe you're right. I mean this code part was disabled the whole time and I did not noticed it. So I guess nobody will miss it.

IvanNardi: Agree. Removed in the last iteration

utoni (approved):

PR #1732, opened 2022-09-11
Create proj and sln files for compile of x64 lib under Visual Studio 2019

mmaypo:
I needed to run tests under VS, so added these files to get a windows dll (can easily generate static lib if needed, one change in project settings).  It compiles and successfully finds apps using ndpi_detection_process_packet(...).   caveat: I have not tested thoroughly, and did just enough work to get it running.

add VS2019 project and solution files
code changes required to fix compile errors under VS


utoni:
I am against a VS integration. We had already one and removed it because nobody was able to maintain it and will be outdated very soon resulting in compilation failures. At the latest when somebody adds a new protocol dissector.

But providing some missing Windows API functions seems to be a good idea.

utoni:
What do you think, @IvanNardi?

IvanNardi:
I think the Windows code, the code changes in general and even the .gitignore file are useful, anyway.
I have zero experience with VS, but I do remember when we removed the old integration.
However, even if I don't have strong opinions, I think that we can merge it:
* best case scenario: we will find a way to test VS into Github CI
* worst case scenario: the VS files will become stale (again).  However they might be useful to someone anyway. If we decided to remove them again, removing two files will be trivial (without any impact on the rest of the code)

.gitignore
```diff
@@ -12,72 +38,57 @@ compile
...
 /tests/unit/Makefile
-/example/libndpiReader.a
-/example/ndpiReader.exe
-/src/lib/libndpi.a
-/src/lib/libndpi.so*
-/src/lib/libndpi
-/src/lib/libndpi-*.dll
-/src/lib/libndpi.4
-/tests/dga/dga_evaluate
-/tests/dga/dga_evaluate.exe
```
utoni, line 75: Why remove? This executable is generated when compiling with MingW.

utoni:
>     * best case scenario: we will find a way to test VS into Github CI

Not sure, but https://github.com/microsoft/setup-msbuild might be able to build C/C++ `*.sln` files. Will investigate.

Ok, I'll give this PR a try.

.gitignore
```diff
@@ -12,72 +38,57 @@ compile
...
+/rrdtool/rrd_similarity
 /src/include/Makefile
-/src/include/ndpi_config.h
 /src/include/ndpi_api.h
+/src/include/ndpi_config.h
 /src/include/ndpi_define.h
 /src/lib/.deps/
 /src/lib/Makefile
+/src/lib/libndpi
+/src/lib/libndpi-4.1.0.dll
```
utoni, line 78: Probably better to replace it with `/src/lib/libndpi-*.dll`

utoni:
@mmaypo
I can not compile your solution with VS2019.
How did you handle the missing `*.h` which are generated from `*.h.in` files? Since they are generated by autoconf and I do not know if there is something available with msbuild.

mmaypo:
under WSL on the windows dev box, I ran the linux autoreconf to generate 
the files and the generated files worked unchanged in VS2019

utoni:
Not sure if microsoft/msbuild comes with WSL support.

utoni:
@mmaypo Could elaborate a bit more how you did run autoreconf on your Windows machine?

utoni:
Thank you for you work. Without that, a CI integration would be impossible. Your PR was dragged to #1742 and extended with CI/MSBuild support.

mmaypo:
glad to do my little bit!

PR #1734, opened 2022-09-12
HTTP: remove some wrong code

IvanNardi:
This code seems wrong or in the wrong place, at least:
* "classification by port" and "classification by ip" protocols (i.e "guessed" protocols) should be used to set the protocol stack only after trying all the dissectors, and only by the generic code
* there are no reason (for a dissector) to update the "guessed" information using the protocol stack values: it is usually the other way around (see previous point)


utoni (approved):
removing legendary code LGTM

PR #1736, opened 2022-09-14
Dissectors shouldn't update `flow->guessed_host_protocol_id`

IvanNardi:
The field `flow->guessed_host_protocol_id` is set at the beginning of the flow analysis and it represents the "classification by ip" of the flow itself.

This field should never be changed. Dissectors which want to provide an "hint" about the classification, should update `flow->guessed_protocol_id` instead. Such "hint" is useless if the dissector set the "extra-dissection" data-path.

Rename such field to `guessed_protocol_id_by_ip` to better describe its role.

Preliminary work necessary for #1687

### DIFF
--- a/src/include/ndpi_typedefs.h
+++ b/src/include/ndpi_typedefs.h
@@ -1260,9 +1260,9 @@ struct ndpi_flow_struct {
   u_int16_t detected_protocol_stack[NDPI_PROTOCOL_SIZE];
 
   /* init parameter, internal used to set up timestamp,... */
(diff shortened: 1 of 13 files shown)


utoni:
The similarity of `guessed_host_protocol_id` and `guessed_protocol_id` confused me a lot in the past. There has to be a better name than `guessed_host_protocol_id` like `guessed_protocol_id_by_ip` or something else.

IvanNardi:
I agree. Do you want I rename `guessed_host_protocol_id` into `guessed_protocol_id_by_ip` int this PR?

utoni:
Con piacere. :)

utoni (approved):

PR #1737, opened 2022-09-14
Remove a case of guessed sub-classification

IvanNardi:
This code is triggered only for "unknown" flows with a valid sni/hostname.
Why in that case the guessed classification should be something like `DNS/Subprotocol_depending_on_hostname`? Why DNS as master and not HTTP or TLS or QUIC?

Furthermore, I have not been able to trigger a positive match from that lookup. I strongly think that if we had a valid subprotocol, we would have a valid master in the first place.
In doubt, remove it completely.

As a follow up, we should investigate why some dissectors (the HTTP one, at least) set the sni/hostname field without setting a valid protocol, in the first place.
This behaviour seems quite suspicious, if not plainly buggy.


src/lib/ndpi_main.c
```diff
@@ -5788,17 +5788,6 @@ ndpi_protocol ndpi_detection_giveup(struct ndpi_detection_module_struct *ndpi_st
...
-        ndpi_protocol_match_result ret_match;
-
-        ndpi_match_host_subprotocol(ndpi_str, flow, (char *) flow->host_server_name,
-				    strlen((const char *) flow->host_server_name), &ret_match,
-				    NDPI_PROTOCOL_DNS);
-
-        if(ret_match.protocol_id != NDPI_PROTOCOL_UNKNOWN)
-          guessed_host_protocol_id = ret_match.protocol_id;
-      }
-
```
utoni, line 5801: I am glad you removed that block as it does not make sense to me and should be handled in the dissector (if required).

Although this was added with 6edb7bedd7b5c1eb0060f5d503f9e37a6d58c086 the commit message does not help me much with that specific code block.

// EDIT: It wasn't added, but modified in 6edb7bedd7b5c1eb0060f5d503f9e37a6d58c086.  Sorry.

utoni (approved):

PR #1739, opened 2022-09-16
Add Syncthing dissector.

utoni:
We could dissect more interesting data if we had a protobuf dissector. One of many personal ToDo's.


IvanNardi:
@utoni, could you rebase, please?

IvanNardi (approved):

PR #1741, opened 2022-09-17
Fix warnings on OpenWrt CCs, caused by the memcpy builtin.

utoni:
This is a false positive IMHO, because looking at the calls to `tls13_hkdf_expand_label_context`, I could not find any that leads to `context_length != 0`, but `context_hash == NULL`.

The warning could be related to `-D_FORTIFY_SOURCE=1`, unsure.


IvanNardi (approved):
Definitely a false postive

PR #1742, opened 2022-09-18
Visual Studio 2019 + MSBuild CI support

utoni:
* add CI support via MSBuild


utoni:
@mmaypo @IvanNardi 
What do you think? Visual Studio integration now works also with the CI. The only issue is the missing `autoconf`. That way there is the need for a separate `ndpi_define.h`. I've also moved `NDPI_API_VERSION` to `ndpi_define.h`. It seems more logical to me.

IvanNardi:
First of all, thanks for this work to both of you, @utoni and @mmaypo!
Just three notes:
*) I fear that `src/include/ndpi_define.h` and `windows/src/ndpi_define.h` will be out of sync in the future... But we will live with that...
*)Is it possibile to *test* somehow the compilation results? Even a simple `ndpiReader -v` to be sure that the executable and library are fine... (the entire unit tests suite should be the optimal solution but leave that for another day...)
*) could you add a line to the main README file about Windows + VS?

utoni:
Unfortunately, running ndpiReader and regression tests in the CI is not as easy as expected. I think we can go with the working build process for now. What do you think?

IvanNardi:
> Unfortunately, running ndpiReader and regression tests in the CI is not as easy as expected. I think we can go with the working build process for now. What do you think?

Fine for me. It is anyway a big step forward respect the initial PR!

IvanNardi (approved):

PR #1743, opened 2022-09-18
Remove classification "by-ip" from protocol stack

### LINKED ISSUE #1687: HTTP sub-classification: best protocol classification

IvanNardi:
Basically:
* "classification by-ip" (i.e. `flow->guessed_protocol_id_by_ip` is NEVER returned in the protocol stack (i.e.
`flow->detected_protocol_stack[]`);
* if the application is interested into such information, it can access `ndpi_protocol->protocol_by_ip` itself.

There are mainly 4 points in the code that set the "classification by-ip" in the protocol stack:  the generic `ndpi_set_detected_protocol()`/ `ndpi_detection_giveup()` functions and the HTTP/STUN  dissectors.

In the unit tests output, a print about `ndpi_protocol->protocol_by_ip` has been added for each flow: the huge diff of this commit is mainly due to that.

Strictly speaking, this change is NOT an API/ABI breakage, but there are important differences in the classification results. For examples:
* TLS flows without the initial handshake (or without a matching SNI/certificate) are simply classified as `TLS`;
* similar for HTTP or QUIC flows;
* DNS flows without a matching request domain are simply classified as `DNS`; we don't have `DNS/Google` anymore just because the server is 8.8.8.8 (that was an outrageous behaviour...);
* flows previusoly classified only "by-ip" are now classified as `NDPI_PROTOCOL_UNKNOWN`.

See #1425 for other examples of why adding the "classification by-ip" in the protocol stack is a bad idea.

Please, note that IPV6 is not supported :(  (long standing issue in nDPI) i.e. `ndpi_protocol->protocol_by_ip` wil be always `NDPI_PROTOCOL_UNKNOWN` for IPv6 flows.

Define `NDPI_CONFIDENCE_MATCH_BY_IP` has been removed.

Close #1687


example/ndpiReader.c
```diff
@@ -1401,11 +1403,14 @@ static void printFlow(u_int32_t id, struct ndpi_flow_info *flow, u_int16_t threa
     if(flow->tunnel_type != ndpi_no_tunnel)
       fprintf(out, "%s:", ndpi_tunnel2str(flow->tunnel_type));
 
-    fprintf(out, "%s/%s]",
+    fprintf(out, "%s/%s (IP: %u/%s)]",
```
utoni: I would recommend to use square brackets to reduce confusion.

```suggestion
    fprintf(out, "%s/%s][IP: %u/%s]",
```

src/lib/ndpi_main.c
```diff
@@ -5783,38 +5784,22 @@ ndpi_protocol ndpi_detection_giveup(struct ndpi_detection_module_struct *ndpi_st
...
        is_udp_guessable_protocol(guessed_protocol_id))
       flow->guessed_protocol_id = guessed_protocol_id = NDPI_PROTOCOL_UNKNOWN;
 
-    if((guessed_protocol_id != NDPI_PROTOCOL_UNKNOWN) || (guessed_host_protocol_id != NDPI_PROTOCOL_UNKNOWN)) {
-      ndpi_confidence_t confidence;
-
-      if(guessed_protocol_id != NDPI_PROTOCOL_UNKNOWN)
-	confidence = NDPI_CONFIDENCE_MATCH_BY_PORT;
-      if(guessed_host_protocol_id != NDPI_PROTOCOL_UNKNOWN)
-	confidence = NDPI_CONFIDENCE_MATCH_BY_IP;
```
utoni, line 5808: Am I right, that `NDPI_CONFIDENCE_MATCH_BY_IP` is now unused?
Looking at the regression tests, it seems that even if the flow was matched by IP, it is not mentioned anymore in the test stats.

example/ndpiReader.c

src/lib/ndpi_main.c
IvanNardi: Removed

tests/unit/unit.c
```diff
@@ -286,7 +286,7 @@ int serializeProtoUnitTest(void)
...
+      char const * const expected_json_str = "{\"flow_risk\": {\"6\": {\"risk\":\"Self-signed Cert\",\"severity\":\"High\",\"risk_score\": {\"total\":500,\"client\":450,\"server\":50}},\"7\": {\"risk\":\"Obsolete TLS (v1.1 or older)\",\"severity\":\"High\",\"risk_score\": {\"total\":510,\"client\":455,\"server\":55}},\"8\": {\"risk\":\"Weak TLS Cipher\",\"severity\":\"High\",\"risk_score\": {\"total\":250,\"client\":225,\"server\":25}},\"17\": {\"risk\":\"Malformed Packet\",\"severity\":\"Low\",\"risk_score\": {\"total\":260,\"client\":130,\"server\":130}}},\"confidence\": {\"210\":\"DPI\"},\"proto\":\"TLS.Facebook\",\"proto_id\":\"91.119\",\"encrypted\":1,\"breed\":\"Fun\",\"category_id\":6,\"category\":\"SocialNetwork\",\"float\":340282346638528859811704183484516925440.000000,\"double\":680564693277057719623408366969033850880.000000}";
```
utoni, line 289: `confidence` value of 210? This looks sus.

IvanNardi: Better suggestion? Take a look at `src/include/ndpi_typedefs.h`. The idea is to have "stable" values, safely convertible to `unsigned char`, ordered by "importance" and leaving space for future values everywhere.

utoni: I've missed the changes in `src/include/ndpi_typedefs.h`. Fine with me.

PR #1745, opened 2022-09-19
Improved NATPMP dissection.

utoni:


src/lib/protocols/natpmp.c
```diff
@@ -44,70 +47,144 @@ static void ndpi_int_natpmp_add_connection(struct ndpi_detection_module_struct *
...
+static int ndpi_search_natpmp_extra(struct ndpi_detection_module_struct *ndpi_struct,
+                                    struct ndpi_flow_struct *flow)
+{
+  struct ndpi_packet_struct const * const packet = &ndpi_struct->packet;
+  enum natpmp_type natpmp_type;
+
+  if (natpmp_is_valid(packet, &natpmp_type) == 0)
+  {
+    ndpi_set_risk(ndpi_struct, flow, NDPI_MALFORMED_PACKET, "Invalid NATPMP Header");
+    natpmp_disable_extra_dissection(flow);
```
IvanNardi: Unnecessary, generic code will disable extra dissection if the return value of the callabck is 0

tests/result/iphone.pcap.out
```diff
@@ -88,6 +88,6 @@ JA3 Host Stats:
...
 	48	ICMPV6 [fe80::823:3f17:8298:a29c]:0 -> [ff02::2]:0 [proto: 102/ICMPV6][ClearText][Confidence: DPI][cat: Network/14][2 pkts/132 bytes -> 0 pkts/0 bytes][Goodput ratio: 6/0][4.21 sec][Plen Bins: 100,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0]
-	49	UDP 192.168.2.1:5351 -> 224.0.0.1:5350 [proto: 312/NAT-PMP][ClearText][Confidence: Match by port][cat: Network/14][2 pkts/120 bytes -> 0 pkts/0 bytes][Goodput ratio: 30/0][< 1 sec][Plen Bins: 100,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0]
+	49	UDP 192.168.2.1:5351 -> 224.0.0.1:5350 [proto: 312/NAT-PMP][ClearText][Confidence: Match by port][cat: Network/14][2 pkts/120 bytes -> 0 pkts/0 bytes][Goodput ratio: 30/0][< 1 sec][Result: 0][Internal Port: 0][External Port: 0][External Address: 0.0.0.0][Plen Bins: 100,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0]
```
IvanNardi: This is probably a generic issue (not related to NATPMP): printing metadata for flows with "Match by port" confidence doesn't make sense since these value can't be right (if we had valid metadata, the classification would be "higher".
Like I said, this is probably a generic issues: I simply noticed it now for the first time

Not sure about the best course of action: printing only if != 0? It seems 0 is not a valid value for all of 3

example/reader_util.c
```diff
@@ -1107,6 +1107,14 @@ void process_ndpi_collected_info(struct ndpi_workflow * workflow, struct ndpi_fl
     ndpi_snprintf(flow->softether.fqdn, sizeof(flow->softether.fqdn), "%s",
                   flow->ndpi_flow->protos.softether.fqdn);
   }
+  /* NATPMP */
+  else if(is_ndpi_proto(flow, NDPI_PROTOCOL_NATPMP)) {
+    flow->info_type = INFO_NATPMP;
+    flow->natpmp.result_code = flow->ndpi_flow->protos.natpmp.result_code;
+    flow->natpmp.internal_port = flow->ndpi_flow->protos.natpmp.internal_port;
+    flow->natpmp.external_port = flow->ndpi_flow->protos.natpmp.external_port;
+    inet_ntop(AF_INET, &flow->ndpi_flow->protos.natpmp.external_address.ipv4, &flow->natpmp.ip[0], sizeof(flow->natpmp.ip));
```
IvanNardi, line 1116: inet_***n***top here...

src/lib/protocols/natpmp.c
```diff
@@ -44,70 +47,144 @@ static void ndpi_int_natpmp_add_connection(struct ndpi_detection_module_struct *
...
+      flow->protos.natpmp.internal_port = ntohs(get_u_int16_t(packet->payload, 4));
+      flow->protos.natpmp.external_port = ntohs(get_u_int16_t(packet->payload, 6));
+      if (flow->protos.natpmp.internal_port == 0)
+      {
+        ndpi_set_risk(ndpi_struct, flow, NDPI_MALFORMED_PACKET, "Request Port Mapping: Internal port must not 0");
+      }
+      break;
+    case NATPMP_RESPONSE_ADDRESS:
+      flow->protos.natpmp.result_code = ntohs(get_u_int16_t(packet->payload, 2));
+      flow->protos.natpmp.external_address.ipv4 = ntohl(get_u_int32_t(packet->payload, 8));
```
IvanNardi: ... and ntohl here... is that right? Unfortunately there is no an example with this parameter

utoni: It is wrong. `man 3 inet_ntop` tells me that it expects network byte order.

tests/result/iphone.pcap.out
utoni: An additional check as work-around in ndpiReader seems legit.

IvanNardi (approved):

PR #1746, opened 2022-09-21
Fix CI after nBPF integration

IvanNardi:
Add one CI job testing nBPF

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -13,6 +13,9 @@ AS_IF([test "${with_only_libndpi+set}" = set],[
   EXTRA_TARGETS=""
 ])
 
+PWD=`pwd`
+
+
 AC_ARG_WITH(sanitizer,     AS_HELP_STRING([--with-sanitizer], [Build with support for address, undefined and leak sanitizer]))
 AC_ARG_WITH(thread-sanitizer, AS_HELP_STRING([--with-thread-sanitizer], [Build with support for thread sanitizer]))
 AC_ARG_WITH(memory-sanitizer, AS_HELP_STRING([--with-memory-sanitizer], [Build with support for memory sanitizer]))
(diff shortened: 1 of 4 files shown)


utoni:
There is an annoying warning in the test diff's: `[Unknown]: nDPI compiled without nBPF support: skipping rule`.

IvanNardi:
> There is an annoying warning in the test diff's: `[Unknown]: nDPI compiled without nBPF support: skipping rule`.

There is a `NDPI_LOG_ERR(...` call if there is a custom nBPF rule but no nBPF support.
This log is quite annoying in the CI, but the CI are compiled with the debug logs enabled, so it is kind of expected...
Some solutions:
1) Lower the level of the message so it is not print by-default, even if debug logs are enabled
2) Enable nBPF in all the CI (that might be usefull anyway, but we are "hiding" the issue)
3) ...

What do you think?

utoni:
>     1. Lower the level of the message so it is not print by-default, even if debug logs are enabled
>
>     2. Enable nBPF in all the CI (that might be usefull anyway, but we are "hiding" the issue)
> 

We could switch some CI jobs to nBPF enabled and lower the level of the message.

configure.ac
```diff
@@ -198,6 +198,7 @@ fi
 
 MACHINE=`uname -m`
 
+NBPF_ENABLED=0
 AC_MSG_CHECKING([PF_RING nBPF])
 NBPF_HOME=${PWD}/../PF_RING/userland/nbpf
```
utoni: A nice2have would be to let the user decide, what the path of nbpf should be.

IvanNardi: I add some autotool "magic": it seems to work (the user decide the path; default as before) but could you take a look, please?

utoni: lgtm

utoni (approved):
Good job!

PR #1751, opened 2022-09-22
Sizes of LRU caches are now configurable

IvanNardi:
0 as size value disable the cache.

The diffs in unit tests are due to the fact that some lookups are performed before the first insert: before this change these lookups weren't counted because the cache was not yet initialized, now they are.


example/ndpiReader.c
```diff
@@ -2427,6 +2427,10 @@ static void setupDetection(u_int16_t thread_id, pcap_t * pcap_handle) {
   if(_maliciousSHA1Path)
     ndpi_load_malicious_sha1_file(ndpi_thread_info[thread_id].workflow->ndpi_struct, _maliciousSHA1Path);
 
+  /* Enable/disable/configure LRU caches here */
+  ndpi_set_lru_cache_size(ndpi_thread_info[thread_id].workflow->ndpi_struct,
+			  NDPI_LRUCACHE_BITTORRENT, 32768);
```
utoni, line 2432: This statement is not required and just here for documentation purposes, am I right?

IvanNardi: yes. I kept the original values; this is only an example about where/when during the init phase call these functions

utoni (approved):

PR #1755, opened 2022-09-26
Hangout: avoid useless lookups in the protocol tree

IvanNardi:
We already performed exactly these lookups in the generic code to populate `flow->guessed_protocol_id_by_ip`: use it!

This code probably needs a deeper review, since it is basicaly a simple matching on ip + port.


utoni:
The review will take some time.

utoni (approved):

PR #1756, opened 2022-09-27
TLS/DTLS: we process certificate for UDP flows, too

IvanNardi:
Note that current code access `certificate_processed` state even before setting the protocol classification, so this piece of information can't be saved in `flow->protos` union.

### DIFF
--- a/src/include/ndpi_typedefs.h
+++ b/src/include/ndpi_typedefs.h
@@ -708,7 +708,6 @@ struct ndpi_flow_tcp_struct {
     message_t message[2]; /* Directions */
 
     /* NDPI_PROTOCOL_TLS */
-    u_int8_t certificate_processed:1, fingerprint_set:1, _pad:6;
     u_int8_t app_data_seen[2];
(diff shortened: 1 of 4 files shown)


src/include/ndpi_typedefs.h
```diff
@@ -1373,6 +1372,10 @@ struct ndpi_flow_struct {
     u_int16_t num_processed_pkts;
   } stun;
 
+  struct {
+    u_int8_t certificate_processed:1, _pad:7;
+  } tls_quic; /* Used also by DTLS and POPS/IMAPS/SMTPS/FTPS */
```
utoni, line 1377: Since even QUIC uses a generic TLS handshake, why not 
name it only `tls` ?

IvanNardi: I simply copied the name already used for `flow->protos.tls_quic`.
I could change both of them. What do you think?

utoni: Not necessary. I guess the change-set might be huge.

PR #1759, opened 2022-09-30
Line app support

sharonenoch:
Added Standard Support for LINE app  https://line.me/en/  using ASN and SNI fields.


sharonenoch:
1) Not sure why the 1 Code Smell warning is showing up as the other lines in host_match above it are longer than the lines I had added. Not sure why it's raising the flag only for this line being more than 128 characters.

2) In the 2 different machines which I compiled with make check it skipped the h323-overflow.pcap and the WebattackRCE.pcap as the PCRE_ENABLED flag in do.sh was set to 0. Do I need to install any package to generate the new pcap? I tried it on the below Ubuntu hw I have but it SKIPPED in both.

Any help in resolving the above 2 is appreciated

root@kickseed:~/sharonnDPI/nDPI/tests# lsb_release -a
No LSB modules are available.
Distributor ID: Ubuntu
Description:    Ubuntu 18.10
Release:        18.10
Codename:       cosmic


root@DC:~/ntop_compile/nDPI# lsb_release -a
No LSB modules are available.
Distributor ID: Ubuntu
Description:    Ubuntu 18.04.4 LTS
Release:        18.04
Codename:       bionic

sharonenoch:
Figured out that autogen is not enabling PCRE by default. So had to enable it via ./configure --with-pcre --with-maxminddb
After that the do.sh had the PCRE_ENABLED=1.  After that the h323-overflow.pcap file was generated. 

Not sure what caused the MSBuild to fail as there was no code changes in this commit and previous commit it had ran without any issues.. 

Still the code smell issue is present related to the line more than 128 characters

IvanNardi:
> Not sure what caused the MSBuild to fail as there was no code changes in this commit and previous commit it had ran without any issues..

It was a temporary failure in Github runners. Fixed

> Still the code smell issue is present related to the line more than 128 characters

This happens all the time when we add domains to that list. We ignore them

IvanNardi (approved):

IvanNardi:
Thanks for this useful contribution

PR #1760, opened 2022-10-01
Restore Confidence enum constants to ascending values.

utoni:
* NDPI_CONFIDENCE_MAX is useless otherwise since it is supposed the be the highest value w/o any gaps
 * for client applications, it makes sense to use that the numbers directly w/o any additional mapping


IvanNardi:
I see your point here: `NDPI_CONFIDENCE_MAX` is also used in `ndpiRreeader` code as size in allocating an array...
I do like having stable values here (and in general, a more stable API/values) but we don't even have stable protocol ids :-( so having stable confidence values might be a bit too much, at the moment.

Anyway, if you want to push this change, no issue from me, but please remove the comment at the beginning of the enum

IvanNardi (approved):

PR #1761, opened 2022-10-01
LINE_CALL: add detection of LINE voip calls

IvanNardi:
These flows are classifed as `LINE_CALL`; another option was `RTP/LINE_CALL`. No sure about the best solution...

Extend LINE domains list.

Remove RTP dead code.


src/lib/protocols/line.c
```diff
@@ -0,0 +1,92 @@
...
+    if(flow->l4.udp.line_pkts[packet->packet_direction] == 0) {
+      flow->l4.udp.line_base_cnt[packet->packet_direction] = packet->payload[3];
+      flow->l4.udp.line_pkts[packet->packet_direction] += 1;
+      return;
+    } else {
+      /* It might be a RTP/RTCP packet. Ignore it and keep looking for the
+         LINE packet numbers */
+      /* Basic RTP detection */
+      if((packet->payload[0] >> 6) == 2 && /* Version 2 */
+         packet->payload_packet_len > 1 &&
```
utoni: not necessary, you've checked already for `packet->payload_packet_len > 10`

src/lib/protocols/quic.c
```diff
@@ -1609,7 +1609,7 @@ static int eval_extra_processing(struct ndpi_detection_module_struct *ndpi_struc
   return 0;
 }
 
-static int is_valid_rtp_payload_type(uint8_t type)
+int is_valid_rtp_payload_type(uint8_t type)
```
utoni: I think it would make for sense to move this function to `rtp.c` even it is not called from there. What do you think?

IvanNardi: It makes sense. Done

utoni (approved):

PR #1764, opened 2022-10-04
Fix compilation and sync utests results

IvanNardi:


src/include/ndpi_typedefs.h
```diff
@@ -1572,8 +1572,8 @@ struct ndpi_flow_struct {
 _Static_assert(sizeof(((struct ndpi_flow_struct *)0)->protos) <= 200,
                "Size of the struct member protocols increased to more than 200 bytes, "
                "please check if this change is necessary.");
-_Static_assert(sizeof(struct ndpi_flow_struct) <= 904,
-               "Size of the flow struct increased to more than 904 bytes, "
+_Static_assert(sizeof(struct ndpi_flow_struct) <= 920,
```
utoni, line 1575: 916 should be enough, theoretically...

utoni: forgot about 8 byte padded structs..

utoni (approved):

src/include/ndpi_typedefs.h
utoni: I think I've got it, structs are still 4-byte padded, but you thought ahead due to #1761.

IvanNardi: Not thinking ahead, sorry :)
I simply checked:
```
ivan@ivan-Latitude-E6540:~/svnrepos/nDPI(utets)$ ../pahole/pahole -a src/lib/libndpi.so | grep ndpi_flow_struct -A 460
[...]

	u_int8_t                   ovpn_session_id[8];   /*   878     8 */
	u_int8_t                   ovpn_counter;         /*   886     1 */
	u_int8_t                   tinc_state;           /*   887     1 */
	u_int16_t                  flow_payload_len;     /*   888     2 */

	/* XXX 6 bytes hole, try to pack */

	/* --- cacheline 14 boundary (896 bytes) --- */
	char *                     flow_payload;         /*   896     8 */
	u_int8_t                   priv_data[16];        /*   904    16 */

	/* size: 920, cachelines: 15, members: 77 */
	/* sum members: 891, holes: 5, sum holes: 22 */
	/* sum bitfield members: 53 bits, bit holes: 2, sum bit holes: 3 bits */
	/* paddings: 1, sum paddings: 4 */
	/* last cacheline: 24 bytes */

PR #1770, opened 2022-10-09
ndpiReader: fix help message. There isn't a 'J' option

IvanNardi:


example/ndpiReader.c
```diff
@@ -441,7 +441,7 @@ static void help(u_int long_help) {
          "-i <file|device> "
 #endif
          "[-f <filter>][-s <duration>][-m <duration>][-b <num bin clusters>]\n"
-         "          [-p <protos>][-l <loops> [-q][-d][-J][-h][-H][-D][-e <len>][-E][-t][-v <level>]\n"
```
utoni, line 444: What does `-J` do? AFAIS it is not handled in the switch-case.

IvanNardi: Exactly... it is probably some leftover of a previous commit where some features has been removed

utoni (approved):

PR #1771, opened 2022-10-10
Fix compilation on Visual Studio and with Mingw

IvanNardi:


IvanNardi:
Let's try to have only one implementation of `gettimeofday`.
I know very little of these Win/Mingw defines: CI is now ok, but I don't know if this change makes sense...

src/lib/ndpi_classify.c
```diff
@@ -683,24 +683,59 @@ ndpi_timeval_to_microseconds(pkt_timeval ts)
 /* **************************************** */
 
 #if defined(WIN32) || defined(_MSC_VER)
-int gettimeofday(struct timeval* tp, struct timezone* tzp)
+
```
utoni, line 686: I actually liked the idea of splitting platform specific code into a separate source file.

IvanNardi: I am fine with any solutions fixing the CI: this was only a blind attempt...

IvanNardi: I think that the core issue is that the library should provide (to the applications) a `gettimeofday` implementation for "any" Windows/Mingw environments, **even** if it doesn't use the VS configuration files present in the repository: in that specific scenario, `windows/src/win-gettimeofday.c` is not compiled right now

utoni: Something like a `ndpi_gettimeofday` ?

IvanNardi: No, exactly `gettimeofday`
I **think** that @lucaderi re-added `gettimeofday` in e4dcec560edf392ff7d962d430aa8a9e12e73097 because it wasn't available in his enviroment. If you look at that commit it seems he fixed the compilation with VS.
Having said that, I know very little of Windows environments and I definitely don't use nDPI on Win... :(

utoni: Still not convinced that this is a nice solution, but it fixes the CI issue.

IvanNardi: Another possibilities might be to always compile windows/src/win-gettimeofday.c into the library (i..e also via standard makefile) and get rid of the implementation added in https://github.com/ntop/nDPI/commit/e4dcec560edf392ff7d962d430aa8a9e12e73097
Does it makes sense?

utoni: I would like to know which VS version @lucaderi is using. On my VM is a VS-2019 installed and everything works fine.

PR #1776, opened 2022-10-17
TLS: explicit ignore client certificate

IvanNardi:
TLS classification usually stops after processing *server* certificates (if any). That means, that *client* certificate, if present, is usually ignored.
However in some corner cases (i.e. unidirectional traffic) we might end up processing client certificate and exposing its metadata: the issue is that the application will think that this metadata are about the server and not about the client.
So, for the time being, always ignore client certificate processing.

As a future work, we might find an efficient way to process and export both certificates.


utoni (commented):
Does this also work for unidirectional midstream traffic i.e. if the first packet passed to nDPI came from the server?

IvanNardi:
Now it should work
I added some more unit tests, to help testing this feature

IvanNardi:
> Does this also work for unidirectional midstream traffic i.e. if the first packet passed to nDPI came from the server?

Now it should work. Tests extended

utoni (approved):

PR #1778, opened 2022-10-17
TLS: allow sub-classification via ALPN

IvanNardi:
In some rare cases, it is possible to sub-classify the flow via ALPN matching. This is particularly usefull for asymmetric traffic where the Client Hello doens't have the SNI.

For the time being there is only one rule, about ANYDESK.


utoni (commented):
Interesting. Just out of curiosity: Do you have any pcap available?

IvanNardi:
> Interesting. Just out of curiosity: Do you have any pcap available?

If you are only interested in a CH with such ALPN, you can look at `tests/pcap/anydesk.pcapng`

tests/result/tls_client_certificate_with_missing_server_one.pcapng.out
```diff
@@ -19,13 +19,14 @@ Patricia risk mask:   4/0 (search/found)
...
+TLS	8	2093	1
+AnyDesk	9	3433	1
 
 JA3 Host Stats: 
 		 IP Address                  	 # JA3C     
 	1	 192.168.1.128            	 1      
 	2	 195.181.174.176          	 1      
 
 
-	1	TCP 195.181.174.176:443 <-> 192.168.1.128:48260 [proto: 91/TLS][IP: 0/Unknown][Encrypted][Confidence: DPI][cat: Web/5][3 pkts/1654 bytes <-> 6 pkts/1779 bytes][Goodput ratio: 87/78][0.04 sec][ALPN: anydesk/6.2.0/linux][bytes ratio: -0.036 (Mixed)][IAT c2s/s2c min/avg/max/stddev: 3/0 10/8 17/20 7/9][Pkt Len c2s/s2c min/avg/max/stddev: 66/66 551/296 1514/1160 681/400][Risk: ** Missing SNI TLS Extn **][Risk Score: 50][TLSv1.2][JA3C: 29b5a018fa5992fe23560c16af0dc9fc][JA3S: e58f0b3c1e9eefb8ee4f92aeceee5858][Firefox][Cipher: TLS_ECDHE_ECDSA_WITH_AES_256_GCM_SHA384][Plen Bins: 0,0,0,0,0,0,0,0,0,33,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,33,0,0,0,0,0,0,0,0,0,0,33,0,0]
```
utoni, line 30: I was wondering why there was no `NDPI_TLS_UNCOMMON_ALPN` set.

IvanNardi: ALPN checks are quite... "unusual"...
`tlsCheckUncommonALPN` is called:
*  only if we have an ALPN on the ServerHello (--> never for TLS 1.3, for example), AND
*  the ALPN checked is, usually, the list in the CH and not the real one in the SH

I think that, at very least, if we have the "negotiated ALPN" we should check it and not the "proposed list".
Bottom line: there is some confusion between the proposed and the negotiated values.
Definitely some work for a followup

IvanNardi: BTW, in the specific case, there is no ALPN in the SH, so the check is not performed

utoni: > I think that, at very least, if we have the "negotiated ALPN" we should check it and not the "proposed list". Bottom line: there is some confusion between the proposed and the negotiated values. Definitely some work for a followup

There is a raison d'être to check both since it is still an uncommon behavior.
But anyway, you're right. This should be a follow-up.

utoni (approved):

PR #1780, opened 2022-10-20
Add Munin protocol dissector.

utoni:
* all credits goes to @verzulli


src/lib/protocols/munin.c
```diff
@@ -0,0 +1,94 @@
...
+  }
+
+  if (memcmp(packet->payload, munin_prefix, NDPI_STATICSTRING_LEN(munin_prefix)) != 0)
+  {
+    NDPI_EXCLUDE_PROTO(ndpi_struct, flow);
+    return;
+  }
+
+  ndpi_int_munin_add_connection(ndpi_struct, flow);
+  flow->host_server_name[0] = '\0';
```
IvanNardi: Do you think that is really necessary?

src/lib/protocols/munin.c
```diff
@@ -0,0 +1,94 @@
...
+
+  if (packet->payload[packet->payload_packet_len - 1] != '\n')
+  {
+    ndpi_set_risk(ndpi_struct, flow, NDPI_MALFORMED_PACKET, "Missing Munin Hostname");
+    return;
+  }
+
+  size_t host_len = packet->payload_packet_len - NDPI_STATICSTRING_LEN(munin_prefix) - 1;
+  if (host_len > 0)
```
IvanNardi, line 71: It seems wrong... if packet->payload_packet_len == NDPI_STATICSTRING_LEN(munin_prefix), the `if` condition is true...
Having `host_len` as a **signed** integer?

utoni: Under those circumstances it should already bail out at `packet->payload[packet->payload_packet_len - 1] != '\n'` or am I wrong?

src/lib/protocols/munin.c
utoni: not really

src/lib/protocols/munin.c
IvanNardi: no, you are right. Thanks

IvanNardi (approved):

PR #1784, opened 2022-10-22
TLS: improve handling of ALPN(s)

IvanNardi:
Tell "Advertised" ALPN list from "Negotiated" ALPN; the former is extracted from the CH, the latter from the SH.

Add some entries to the known ALPN list.

Fix printing of "TLS Supported Versions" field.


utoni:
Thanks, @IvanNardi. Exactly what I wanted.

utoni (approved):

PR #1785, opened 2022-10-23
Fix classification "by-port" of QUIC flows

IvanNardi:
nDPI is able to properly classify QUIC flows only if it elaborates the very first packets of the flow.

The protocol list in `is_udp_guessable_protocol()` is basically a list of protocols which can be detected from *any* packets in the flow. Rename such function to `is_udp_not_guessable_protocol()`: the name is still quite cryptic, but at least not plainly wrong


utoni:
`is_udp_guessable_protocol()` is ambigious, but I do not know any better naming

IvanNardi:
> `is_udp_guessable_protocol()` is ambigious, but I do not know any better naming

Unfortunately, me neither...

utoni (approved):
`is_single_packet_protocol` is probably not an option...

IvanNardi:
> `is_single_packet_protocol` is probably not an option...

No strong opinions...
At very least, I would update  `is_udp_guessable_protocol()` to `is_udp_not_guessable_protocol()` since it returns 1 if the protocol is NOT guessable...

Issue #1781, opened 2022-10-21
Optimal memory allocation with DPDK for nDPI


DimaWittmann:
Dear nDPI team,
Thank you for your work and time.

I was doing a performance test of our application, it's based on DPDK with nDPI integrated.
I noticed that malloc is a bottleneck, essentially it's expected. So I decided to use memory pools to replace rte_malloc for nDPI.
My question is about optimal object sizes for the pools. I tried to have 4 memory pools: 64B, 512B, 4KB, and 32KB and it worked out pretty well. I wondered if you had a similar experience and have some information about optimal sizes.
Another question is how many objects are allocated per traffic flow, to estimate the needed size of the pools.
Please share your thoughts and assumption.

Thank you very much.

utoni:
This is interesting. I was also thinking about mempools for `nDPId`s implementation. But I was too scarred about a possible memory waste. Can you tell something about the usage of the mempool?
~~How do you handle memory fragmentation?~~ Edit: I presume that your mempool chunks have a fixed size, so fragmentation is not an issue.

I guess it is useful to have a pool size where the ndpi flow struct fits perfectly in and some smaller sizes e.g. for string allocation.

IvanNardi:
> I noticed that malloc is a bottleneck, essentially it's expected.

I am very surprised by that.
In my experience, once you pre-allocate `struct ndpi_flow_struct` structures, memory allocation is no more an issue.
I mean, there is still a lot of room for improvements, but memory is definitely not the main performance bottleneck, to me.

Could you share some info on your environment? Traffic rate, hardware... Are you using plain `ndpiReader` with DPDK support?

DimaWittmann:
The setup is following:
15 cores are dedicated to packet processing, essentially they do some basic packet parsing, and then a call to _ndpi_detection_process_packet_
Four 10G ports.
The flow table is about 8+ million entries.

Initially, I did not provide malloc to nDPI, so libc malloc was used, and I did not notice that malloc was the bottleneck. The problem was that nDPI allocated much memory besides ` struct ndpi_flow_struct,` about 1GB per 1 million entries. The lion part of the memory in our platform is allocated for DPDK as huge pages, so I provided _rte_malloc_ to nDPI so that huge pages will be used. 
With more than 10G of traffic, I see significant performance degradation, particularly with `rte_malloc`. Profiling data shows that rte_malloc-related functions take about 60%-80% of the processing time under load.

[...]

PR #1789, opened 2022-10-26
Add basic profiling of memory allocations on data-path

IvanNardi:
The goal is to have an idea of the memory allocation sizes performed in
the **library data-path**, i.e. excluding init/deinit phases and all
the allocations made by the application itself.
In other words, how much memory is needed per-flow, by nDPI, other than
`struct ndpi_flow_struct`?
    
It works only on single-thread configurations.
    
It is not enabled by default (in the unit tests) since different
configuration options (example: `--enable-pcre`) lead to different
results.
    
See: #1781

### DIFF
--- a/example/ndpiReader.c
+++ b/example/ndpiReader.c
@@ -127,6 +127,11 @@ extern u_int16_t min_pattern_len, max_pattern_len;
 extern void ndpi_self_check_host_match(); /* Self check function */
 u_int8_t dump_internal_stats;
 
+struct ndpi_bin malloc_bins;
+int enable_malloc_bins = 0;
+int max_malloc_bins = 14;
+int malloc_size_stats = 0;
+
 struct flow_info {
   struct ndpi_flow_info *flow;
   u_int16_t thread_id;
@@ -506,6 +511,7 @@ static void help(u_int long_help) {
          "  -I                        | Ignore VLAN id for flow hash calculation\n"
          "  -z                        | Enable JA3+\n"
          "  -A                        | Dump internal statistics (LRU caches / Patricia trees / Ahocarasick automas / ...\n"
+         "  -M                        | Memory allocation stats on data-path (only by the library). It works only on single-thread configuration\n"
          ,
          human_readeable_string_len,
          min_pattern_len, max_pattern_len, max_num_packets_per_flow, max_packet_payload_dissection,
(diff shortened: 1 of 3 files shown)


example/ndpiReader.c
```diff
@@ -1110,6 +1124,9 @@ static void parseOptions(int argc, char **argv) {
       for(thread_id = 1; thread_id < num_threads; thread_id++)
         _pcap_file[thread_id] = _pcap_file[0];
     }
+
+    if(num_threads > 1)
```
utoni: Shouldn't that be:

```suggestion
    if(num_threads == 1)
```

or did I miss smth?

example/ndpiReader.c
```diff
@@ -884,6 +890,9 @@ static void parseOptions(int argc, char **argv) {
 
     case 'n':
       num_threads = atoi(optarg);
+      if(num_threads > 1)
+        enable_malloc_bins = 0;
```
utoni: I think a warning to notify the user that memory profiling was deactivated may be helpful. I would go a step further and abort the program to make sure that this parameter configuration does not work.

example/ndpiReader.c
utoni: ~~got it..~~

example/ndpiReader.c

example/ndpiReader.c
IvanNardi: Fixed. Thanks!

example/ndpiReader.c
```diff
@@ -884,6 +890,11 @@ static void parseOptions(int argc, char **argv) {
 
     case 'n':
       num_threads = atoi(optarg);
+      if(num_threads > 1 && enable_malloc_bins == 1)
```
utoni: Is this check really needed? You are already checking for `if(num_threads > 1 && enable_malloc_bins == 1)` after the switch-case.

example/ndpiReader.c
```diff
@@ -974,6 +985,16 @@ static void parseOptions(int argc, char **argv) {
       }
       break;
 
+    case 'M':
+      enable_malloc_bins = 1;
+      if(num_threads > 1)
```
utoni: same here

example/ndpiReader.c
IvanNardi: You are right!

example/ndpiReader.c
IvanNardi: Removed

utoni (approved):

PR #1800, opened 2022-11-13
Add a new fuzzer for QUIC

IvanNardi:
QUIC packets are encrypted/obfuscated; that means that we need to decrypt them before parsing the real (TLS) message. Fuzzing is not effective here, since a random buffer is hardly a valid encrypted QUIC packet.

Add a new fuzzer, testing *decrypted* QUIC packets.

Add a basic corpus.

Fix a few bugs already found by this fuzzer.


utoni (commented):
Is it possible to generate those bin corpus files instead of checking them in to the git repository?

IvanNardi:
> Is it possible to generate those bin corpus files instead of checking them in to the git repository?

It is possible, but it is beside the point. Corpus data is made from initial seed (these few files) and it is constantly merged with results of previous fuzzing iterations. You might end up having thousands of files in your corpus, all of them with meaningless names...
For example:
https://boringssl.googlesource.com/boringssl/+/refs/heads/master/fuzz/
https://github.com/curl/curl-fuzzer/tree/master/corpora/curl_fuzzer

I am going to use this new, trivial fuzzer to try to understand the best way to handle corpus files.

utoni (approved):
makes sense, thanks

PR #1808, opened 2022-11-22
Add protocol disabling feature

IvanNardi:
The application may enable only some protocols.
Disabling a protocol means:
*) don't register/use the protocol dissector code (if any) 
*) disable classification by-port for such a protocol
*) disable string matchings for domains/certificates involving this protocol
*) disable subprotocol registration (if any)

This feature can be tested with `ndpiReader -B list_of_protocols_to_disable`.

Custom protocols are always enabled.

Technically speaking, this commit doesn't introduce any API/ABI
incompatibility. However, calling `ndpi_set_protocol_detection_bitmask2()`
is now mandatory, just after having called `ndpi_init_detection_module()`.

Most of the diffs (and all the diffs in `/src/lib/protocols/`) are due to the removing of some function parameters.

Fix the low level macro `NDPI_LOG`. This issue hasn't been detected sooner simply because almost all the code uses only the helpers `NDPI_LOG_*`


IvanNardi:
Not sure if the new API (add the optional bitmask directly to `ndpi_init_detection_module()`) is the best one.
The core issue is that we need the protocol list in a lot of functions (load custom protocols, categories...) that are called before `ndpi_finalize_initialization()`: providing such list directly to `ndpi_init_detection_module()` is the simpliest way to be sure that we have it when we need it.
Another options would be to mandate the use of `ndpi_set_protocol_detection_bitmask2()` always just after `ndpi_init_detection_module()`, but having two functions that you need to call one after the other is no so different that having only one function...

IvanNardi:
@aouinizied , could you take a look at the python errors, please?
I don't know anything about python bindings, so it probably is something very trivial...

vel21ripn:
> Even if trivial, this commit introduces an important API change: ndpi_set_protocol_detection_bitmask2() has been removed, and the protocol bitmask may be passed to ndpi_init_detection_module().

This is bad news. Now, after loading the module, it will not be possible to disable protocols?
The ndpi_set_protocol_detection_bitmask2() function did not register disabled protocol handlers. I recently fixed this bug in commit 4896dabb6b8d3469f87fa8c746ee070b12429024 (March 27, 2022).
We already have the ndpi_enabled_callbacks_init() function which only registers enabled protocol handlers.

IvanNardi:
@vel21ripn , thanks for your feedback!
Just to be sure; are you saying that "disabling a protocol" already works on current code?
That is not my experience. Trivial example:
```
ivan@ivan-Latitude-E6540:~/svnrepos/nDPI(dev)$ git diff
diff --git a/example/ndpiReader.c b/example/ndpiReader.c
index e82e089d..053896f5 100644
--- a/example/ndpiReader.c
+++ b/example/ndpiReader.c
@@ -2396,6 +2396,7 @@ static void setupDetection(u_int16_t thread_id, pcap_t * pcap_handle) {
 
   // enable all protocols
   NDPI_BITMASK_SET_ALL(all);
+  NDPI_BITMASK_DEL(all, NDPI_PROTOCOL_DNS);
   ndpi_set_protocol_detection_bitmask2(ndpi_thread_info[thread_id].workflow->ndpi_struct, &all);
 
   // clear memory for results
ivan@ivan-Latitude-E6540:~/svnrepos/nDPI(dev)$ ./example/ndpiReader -t -i ./tests/pcap/dns_long_domainname.pcap -v2
[...]
	1	UDP 192.168.1.168:65311 <-> 8.8.8.8:53 [proto: 5/DNS][IP: 126/Google][ClearText][Confidence: DPI][cat: Network/14][1 pkts/103 bytes <-> 1 pkts/159 bytes][Goodput ratio: 59/73][0.02 sec][Hostname/SNI: gmr02c.16.0.fhkfhsdkfhsk.tunnel.example.com][::][Risk: ** Error Code **][Risk Score: 10][Risk Info: DNS Error Code NXDOMAIN][PLAIN TEXT (fhkfhsdkfhsk)][Plen Bins: 0,50,0,50,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0]

```
i.e. even if I tried to disable DNS, the flow is classified as such.
Am I doing something wrong?

vel21ripn:
You are right, ndpi_set_protocol_detection_bitmask2() not work.
And calling ndpi_callback_init() again does not create problems?
There is protection against re-call 
````
if(ndpi_str->callback_buffer) return 0;
````
Must be change to
````
if(ndpi_str->callback_buffer)
   ndpi_free(ndpi_str->callback_buffer);
```

IvanNardi:
`ndpi_callback_init()` is only a part of the problem...
We need the final protocols bitmask BEFORE calling
* `ndpi_callback_init()` and `ndpi_enabled_callbacks_init()`: to register only the needed protocol dissectors
* `ndpi_init_protocol_defaults()`: for correct "by port" classification, correct "sub-protocols" classification and correct certificates/domains matching. Note that in current code `ndpi_init_protocol_defaults()` is directly called in `ndpi_init_detection_module()`

I see only 2 solutions:
1) the one showed in this PR
2) keep `ndpi_set_protocol_detection_bitmask2()` and move `ndpi_init_protocol_defaults()` in it. That means that calling `ndpi_set_protocol_detection_bitmask2()` is probably mandatory.

@vel21ripn , do you prefer option 2?

vel21ripn:
The option 2 is a more flexible solution.

IvanNardi:
@vel21ripn, what do you think of the latest version?

IvanNardi:
> @vel21ripn, what do you think of the latest version?

@vel21ripn , kindly pinging...

vel21ripn:
Apply this PR. If I find unsolvable problems, I will write about it.

IvanNardi:
> Apply this PR. If I find unsolvable problems, I will write about it.

Thanks!

IvanNardi:
@utoni , any thoughts/doubts/preferences?

utoni (approved):
Nope, LGTM so far.
Thanks. This will perfectly integrate into nDPId.

PR #1811, opened 2022-11-26
DTLS: handle (certificate) fragments

IvanNardi:
Keep using the existing function to handle reassembling buffer: rename it from `ndpi_search_tls_tcp_memory` to
`ndpi_search_tls_memory` and make it "transport" agnostic


utoni:
I would love to have a generic reassmble interface, but #1556 is still WiP.

IvanNardi:
> I would love to have a generic reassmble interface, but #1556 is still WiP.

@utoni , do you think we can merge this PR as-is, for the time being?

utoni (approved):
LGTM

IvanNardi:
> LGTM

Thanks!

PR #1816, opened 2022-12-02
Extend CodeQL configuration and fix two warnings

IvanNardi:
```
Call to `memset` may be deleted
Comparison of narrow type with wide type in loop condition
```

See https://www.usenix.org/system/files/conference/usenixsecurity17/sec17-yang.pdf for the new portable function `secure_memzero`


utoni (approved):
Awesome!

PR #1817, opened 2022-12-03
CI: GitHub is updating Ubuntu runners

IvanNardi:
GitHub is moving `ubuntu-latest` to `ubuntu-22.04`: update our dependencies.
See: https://github.blog/changelog/2022-11-09-github-actions-ubuntu-latest-workflows-will-use-ubuntu-22-04/
    
This is the reason of the recent random failures in CI.

Update "newest" tested gcc to gcc-12.

Fix a memory error introduced in 557bbcfc5a5165c9eb43bbdd78435796239cd3c9


.github/workflows/build.yml
```diff
@@ -186,31 +188,31 @@ jobs:
         msan: [""]
         nBPF: [""]
         include:
-          - compiler: "gcc-7"
-            os: ubuntu-latest
```
utoni, line 190: Is it really necessary to change `ubuntu-latest` to `ubuntu-20.04` and why (or why not) ?

IvanNardi: gcc-7 is no more available in (default ppa of) `ubuntu-22.04` (`ubuntu-latest` is a "moving" target and it was `ubuntu-20.04` ans it is going to be `ubuntu-22.04`)
If we want to keep testing gcc-7|12 and clang-7|14 we need to force the OS version

utoni (approved):

PR #1830, opened 2022-12-14
fuzz: add fuzzer testing nDPI (initial) configurations

IvanNardi:
The goal of this fuzzer is to test init and deinit of the library, with
different configurations. In details:
* random memory allocation failures, even during init phase
* random `ndpi_init_prefs` parameter of `ndpi_init_detection_module()`
* random LRU caches sizes
* random bitmask of enabled protocols
* random parameters of `ndpi_set_detection_preferences()`
* random initialization of opportunistic TLS
* random load/don't load of configuration files
    
This new fuzzer is a C++ file, because it uses `FuzzedDataProvider`
class (see
https://github.com/google/fuzzing/blob/master/docs/split-inputs.md).
Note that the (existing) fuzzers need to be linked with C++ compiler
anyway, so this new fuzzer doesn't add any new requirements.

### DIFF
--- a/fuzz/fuzz_common_code.c
+++ b/fuzz/fuzz_common_code.c
@@ -1,6 +1,32 @@
 
 #include "fuzz_common_code.h"
 
+
+#ifdef ENABLE_MEM_ALLOC_FAILURES
+
+static int mem_alloc_state = 0;
+
+__attribute__((no_sanitize("integer")))
+static int fastrand ()
+{
+  if(!mem_alloc_state) return 1; /* No failures */
+  mem_alloc_state = (214013 * mem_alloc_state + 2531011);
+  return (mem_alloc_state >> 16) & 0x7FFF;
+}
+
+void *malloc_wrapper(size_t size) {
+  return (fastrand () % 16) ? malloc (size) : NULL;
+}
+void free_wrapper(void *freeable) {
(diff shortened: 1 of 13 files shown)


fuzz/fuzz_common_code.c
```diff
@@ -1,13 +1,39 @@
 
 #include "fuzz_common_code.h"
 
+
+#ifdef ENABLE_MEM_ALLOC_FAILURES
+
+int mem_alloc_state = 0;
```
utoni: static? =}

IvanNardi: No. It is used by the fuzzer themselves
```
ivan@ivan-Latitude-E6540:~/svnrepos/nDPI(fuzz-config)$ git grep mem_alloc_state
[..]
fuzz/fuzz_config.cpp:  mem_alloc_state = size;
fuzz/fuzz_ndpi_reader.c:  mem_alloc_state = Size;
```
Do you prefer some kind of helper like `set_mem_alloc_state()`?

utoni: Probably a "cleaner" method of localizing global variables.

IvanNardi: Done!

utoni (approved):

PR #1831, opened 2022-12-15
postgres: improve detection

IvanNardi:
Remove some dead code (found via coverage report)


src/lib/protocols/ajp.c
```diff
@@ -110,12 +110,6 @@ static void ndpi_check_ajp(struct ndpi_detection_module_struct *ndpi_struct,
 void ndpi_search_ajp(struct ndpi_detection_module_struct *ndpi_struct,
  struct ndpi_flow_struct *flow)
 {
-  // Break after 20 packets.
-  if(flow->packet_counter > 20) {
```
utoni, line 114: Are you sure that this is 0xdeadc0de or just not triggered by any pcap file?
I can imagine that the side effects of other dissectors may it look like this code is dead, but is it the case in a "real" environment?

IvanNardi: All the code paths in `ndpi_check_ajp()` leads to `set_ajp_detected()` or to `NDPI_EXCLUDE_PROTO()`
In other words, `ndpi_search_ajp()` is called only for the first packet of the flow.
So I do think that this "if block" is never triggered

utoni (approved):
Interesting. At least one 0xdeadc0de line was written by me. =)

PR #1832, opened 2022-12-15
Remove some old protocols

IvanNardi:
These protocols:
* have been addeded in the OpenDPI era
* have never been updated since then
* we don't have any pcap examples [*]

If (and it is a big if...) some of these protocols are still somehow used and if someone is still interested in them, we can probably re-add them starting from scratch (because the current detection rules are probably outdated)

Protocols removed: DIRECT_DOWNLOAD_LINK, APPLEJUICE, DIRECTCONNECT,
OPENFT, FASTTRACK, SHOUTCAST, THUNDER, AYIYA, STEALTHNET, FIESTA,
FLORENSIA, AIMINI, SOPCAST

PPSTREAM dissector works (...) only on UDP.
    
[*]: we do have an AIMINI test pcap but it was some trivial http
traffic detected only by hostname matching, on domains no more
available...


utoni:
I remember that I did some testing for aimini last year and the detection ratio was pretty good.

IvanNardi:
> I remember that I did some testing for aimini last year and the detection ratio was pretty good.

How can I test aimini stuff? Some app? Domains `aimini.com` and `aimini.net` are not reachable (for me, at least)...
Anyway, if you prefer we could keep it, and remove the other ones

utoni (approved):
I guess Aimini is not that important and can be re-added later together with a pcap. So, you're good to go.

PR #1833, opened 2022-12-16
Fix infinite loop when a custom rule has port 65535

### LINKED ISSUE #1829: When the port of the custom tcp/udp protocol is set to 65535, there is a problem with the ndpi interface
## Describe the bug
When the port of the custom tcp/udp protocol is set to 65535，loop can‘t be beak!
I found in the function addDefaultPort that when port = 65535, the for loop cannot be beak，I use ndpi 4.0.0。

IvanNardi:
Close #1829


utoni (approved):
Whoops, I missed that one during code review..

PR #1845, opened 2023-01-05
Reduce the size of some traces used as unit test

IvanNardi:
No traces and no flows has been removed; only long sessions has been reduced, keeping only their first packets.

This is quite important in fuzzing systems, since these pcaps are used as initial seed. There is no value in fuzzing long sessions, because only the very first packets are really used/processed by nDPI.

Before:
```
du -h tests/pcap/
200M	tests/pcap/
```
After:
```
du -h tests/pcap/
98M	tests/pcap/
```


utoni (approved):

utoni:
It makes sense to reduce the size of the PCAPs even if it does not affect the `git clone` speed (except if `--depth` is used) but it shrinkes the tarball size which is good I guess.

PR #1851, opened 2023-01-14
Fix classification "by-port"

IvanNardi:
Classification "by-port" should be the last possible effort, *after* having test all the LRU caches.

Remove some dead code from `ndpi_detection_giveup()`: `flow->guessed_protocol_id` is never set to any of those voip protocols and at that point in this function we never have both a master *and* a application protocols.
Coverage reports (both from unit tests and from fuzzing) confirms that was dead code.


utoni:
> Classification "by-port" should be the last possible effort, _after_ having test all the LRU caches.
> 
> Remove some dead code from `ndpi_detection_giveup()`: `flow->guessed_protocol_id` is never set to any of those voip protocols and at that point in this function we never have both a master _and_ a application protocols. Coverage reports (both from unit tests and from fuzzing) confirms that was dead code.

I am unsure if fuzzing is able to trigger **all** code paths. It may depend on the delivered fuzz corpus which depends on the pcap based unit tests.

IvanNardi:
> > Classification "by-port" should be the last possible effort, _after_ having test all the LRU caches.
> > Remove some dead code from `ndpi_detection_giveup()`: `flow->guessed_protocol_id` is never set to any of those voip protocols and at that point in this function we never have both a master _and_ a application protocols. Coverage reports (both from unit tests and from fuzzing) confirms that was dead code.
> 
> I am unsure if fuzzing is able to trigger **all** code paths. It may depend on the delivered fuzz corpus which depends on the pcap based unit tests.

You are right. However, the meaning of that phrase was that the manual code analysis was *confirmed* by the code coverage. See below

utoni (approved):

PR #1852, opened 2023-01-14
Add some fuzzers to test algorithms and data structures

IvanNardi:
Fix some issues found with these new fuzzers


utoni (approved):
An additional fuzzer for the serialization interface might be useful.
But this is already good to go.

IvanNardi:
> An additional fuzzer for the serialization interface might be useful. But this is already good to go.

Yeah, I tried that and some issues (that I am not able to fix...) arose... I might push that partial work, to share the bugs...

PR #1855, opened 2023-01-17
LRU caches: add a generic (optional and configurable) expiration logic

IvanNardi:
Two caches already implemented a similar mechanism: make it generic.

### DIFF
--- a/src/include/ndpi_api.h
+++ b/src/include/ndpi_api.h
@@ -993,11 +993,11 @@ extern "C" {
   void ndpi_set_debug_bitmask(struct ndpi_detection_module_struct *ndpi_mod, NDPI_PROTOCOL_BITMASK debug_bitmask);
 
   /* LRU cache */
-  struct ndpi_lru_cache* ndpi_lru_cache_init(u_int32_t num_entries);
+  struct ndpi_lru_cache* ndpi_lru_cache_init(u_int32_t num_entries, u_int32_t ttl);
   void ndpi_lru_free_cache(struct ndpi_lru_cache *c);
   u_int8_t ndpi_lru_find_cache(struct ndpi_lru_cache *c, u_int32_t key,
-			       u_int16_t *value, u_int8_t clean_key_when_found);
-  void ndpi_lru_add_to_cache(struct ndpi_lru_cache *c, u_int32_t key, u_int16_t value);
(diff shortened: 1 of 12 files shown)


src/include/ndpi_api.h
```diff
@@ -993,11 +993,11 @@ extern "C" {
...
 
   /* LRU cache */
-  struct ndpi_lru_cache* ndpi_lru_cache_init(u_int32_t num_entries);
+  struct ndpi_lru_cache* ndpi_lru_cache_init(u_int32_t num_entries, u_int32_t ttl);
   void ndpi_lru_free_cache(struct ndpi_lru_cache *c);
   u_int8_t ndpi_lru_find_cache(struct ndpi_lru_cache *c, u_int32_t key,
-			       u_int16_t *value, u_int8_t clean_key_when_found);
-  void ndpi_lru_add_to_cache(struct ndpi_lru_cache *c, u_int32_t key, u_int16_t value);
+			       u_int16_t *value, u_int8_t clean_key_when_found, u_int32_t now_sec);
+  void ndpi_lru_add_to_cache(struct ndpi_lru_cache *c, u_int32_t key, u_int16_t value, u_int32_t now_sec);
```
utoni, line 1003:
```suggestion
  void ndpi_lru_add_to_cache(struct ndpi_lru_cache *c, u_int32_t key, u_int16_t value, struct ndpi_flow *flow);
```

AFAIS, the flow part is always required for the time calculation and that way you can skip all those `time_in_ms / 1000` statements.

IvanNardi: I thought about that, but I preferred the current interface; this way the LRU data structure might be used "stand alone", independently from Deep-Packet-Inspection stuff. The other data structures and algorithms provided from nDPI (patricia, ahocorasick...) have this same property, AFAIK

Anyway, I don't have strong feelings: if you prefer I'll update the interface

utoni: Another way could be a helper function/macro that does the ms2s calculation. Might look cleaner (or maybe not). :D

IvanNardi: Ok, I'll give it a try with a helper/macro!
Worst case, we will revert to the current code

utoni (approved):
Looks way better now. =)

IvanNardi:
> Looks way better now. =)

I agree :)
Thanks for taking to time to review these PRs

PR #1860, opened 2023-01-18
Fuzz: fix compilation with nBPF

### LINKED ISSUE #1859: nDPI-dev fuzzers cause linker error with PF_RING-dev
## Describe the bug
nDPI (current `dev` branch) is not able to build the fuzzers if PF_RING (current `dev` branch) is enabled.

### Expected behavior
There should not be any linker errors during the build process.
### Obtained behavior
The linker is missing some symbols:
image

## nDPI Environment:
* OS name: Debian
* OS version: stable
* Architecture: amd64
* nDPI version or commit hash: `dev`.
* nDPI compilation flags used: `--with-sanitizer --enable-fuzztargets CC=clang-11 CXX=clang++-11 --enable-option-checking=fatal --enable-debug-messages --with-pcre --with-maxminddb --enable-tls-sigs`.
* [config.log](https://github.com/ntop/nDPI/files/10439490/config.log)

Nope.

## Additional context
If this turns out to be a real bug, maybe we should add/modify a CI job to build against PF_RING.
Or modify `configure.ac` that it does not automatically tries to link against `../PF_RING/nbpf/...`.

IvanNardi:
Close #1859

### DIFF
--- a/fuzz/Makefile.am
+++ b/fuzz/Makefile.am
@@ -6,8 +6,8 @@ bin_PROGRAMS += fuzz_ds_patricia fuzz_ds_ahocorasick
 
 fuzz_process_packet_SOURCES = fuzz_process_packet.c fuzz_common_code.c
 fuzz_process_packet_CFLAGS = @NDPI_CFLAGS@ $(CXXFLAGS)
-fuzz_process_packet_LDADD = ../src/lib/libndpi.a
-fuzz_process_packet_LDFLAGS = $(ADDITIONAL_LIBS) $(LIBS)
+fuzz_process_packet_LDADD = ../src/lib/libndpi.a $(ADDITIONAL_LIBS)
+fuzz_process_packet_LDFLAGS = $(LIBS)
 if HAS_FUZZLDFLAGS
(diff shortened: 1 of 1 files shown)


IvanNardi:
No sure why, but clang + nBPF + fuzzing didn't like the order of the static libraries...

utoni (approved):
LGTM. Should we add or modify at least one CI job to build with PF_RING/nBPF ?

IvanNardi:
> LGTM. Should we add or modify at least one CI job to build with PF_RING/nBPF ?

We already have a job with nBPF, but without fuzzing (this bug was in the "fuzzing" Makefile, not in the generic ones)
Fuzzing jobs in CI are deeply integrated with oss-fuzz and at the moment they don't use any external libraries (pcre, nBPF, maxminddb)[*]
Not sure if we want to add them now... preferences?



[*] `LDFLAGS="-lpcap" ./autogen.sh --enable-fuzztargets --with-only-libndpi`

utoni:
Maybe not required. I think that this is only an issue when new fuzzers are added to the repository (which won't happen that often).

utoni (approved):

PR #1861, opened 2023-01-18
Minor changes in `ndpi_detection_giveup()`

IvanNardi:
Check the mining cache at the end of the function, like all the others LRU caches.
Rewrite the STUN checks: same identical semantic but in a cleaner way, hopefully


utoni (approved):
Looks waaay better now. I can't believe that the code is now readable. =)

PR #1862, opened 2023-01-18
Trying to fuzz more serialization code

IvanNardi:


IvanNardi:
@utoni, like i said yesterday, I am facing same issues trying to fuzz more serialization code.

First of all, does this code make any sense? [I don't use the serialization stuff and I don't know the API]

If so, if you try to run this (modified) fuzzer (with an empty corpus) you end up with a timeout or a memory error in a few seconds.
Note:
* the error is NOT reproducibile using the final artifact
* if, instead of using `ndpi_reset_serializer()`, you call `ndpi_init_serializer()` at *very* iterations, the issue seems to disappear

That let me think that there is some issues in `ndpi_reset_serializer()`

Any ideas?

utoni (commented):
This is good!
I am not sure if it makes sense to fuzz `ndpi_flow2json` as well.
AFAIR the function signature was **huge**.

utoni:
> @utoni, like i said yesterday, I am facing same issues trying to fuzz more serialization code.
> 
> First of all, does this code make any sense? [I don't use the serialization stuff and I don't know the API]
> 
> If so, if you try to run this (modified) fuzzer (with an empty corpus) you end up with a timeout or a memory error in a few seconds. Note:
> 
>     * the error is NOT reproducibile using the final artifact
> 
>     * if, instead of using `ndpi_reset_serializer()`, you call `ndpi_init_serializer()` at _very_ iterations, the issue seems to disappear
> 
> 
> That let me think that there is some issues in `ndpi_reset_serializer()`
> 
> Any ideas?

I've never used `ndpi_deserialize_clone_all` at all. Maybe there is an issue there?

IvanNardi:
> This is good! I am not sure if it makes sense to fuzz `ndpi_flow2json` as well. AFAIR the function signature was **huge**.

I think we already test it in this exact fuzzer.

IvanNardi:
> I've never used `ndpi_deserialize_clone_all` at all. Maybe there is an issue there?

Honestly, no idea. It might be...

utoni:
I'll try this out ASAP. I hope that I am able to reproduce this issue on my machine.

IvanNardi:
@utoni, I was using the API in the wrong way.
Please ignore this code.
(I added an example of how to use `ndpi_deserialize_clone_all` in `tests/unit/unit.c` in #1876)
Sorry for the noise

PR #1867, opened 2023-01-19
fuzz: fix memory allocation failure logic

IvanNardi:
We *do* want to have some allocation errors.
Fix some related bugs
Fix: 29be01ef


utoni (commented):
`ENABLE_MEM_ALLOC_FAILURES` needs to be enabled manually, right? I do not understand the reason behind that.
Isn't it better to have the malloc NULL wrapper/injector enabled all the time?

IvanNardi:
> `ENABLE_MEM_ALLOC_FAILURES` needs to be enabled manually, right? I do not understand the reason behind that. Isn't it better to have the malloc NULL wrapper/injector enabled all the time?

I prefer to have the possibility to enable/disable the injection of allocation failures.
Anyway, let me try to simplify the code, avoiding spreading that define everywhere...

IvanNardi:
> > `ENABLE_MEM_ALLOC_FAILURES` needs to be enabled manually, right? I do not understand the reason behind that. Isn't it better to have the malloc NULL wrapper/injector enabled all the time?
> 
> I prefer to have the possibility to enable/disable the injection of allocation failures. Anyway, let me try to simplify the code, avoiding spreading that define everywhere...

I removed the define; every fuzzers now decide if to enable or not allocation failures via API
Exception: `fuzz_ndpi_reader.c` since the same file is used to compile two different fuzzers (with and without allocation failures)

utoni (approved):
Good job!

PR #1869, opened 2023-01-20
Some small changes

IvanNardi:
All dissector callbacks should not be exported by the library; make static some other local functions.
The callback logic in `ndpiReader` has never been used. With internal libgcrypt, `gcry_control()` should always return no errors.
We can check `categories` length at compilation time.


src/lib/third_party/include/gcrypt_light.h
```diff
@@ -18,7 +18,7 @@
 #define GPG_ERR_ANY -2
 #define GPG_ERR_INV_ARG -3
 #define GCRYCTL_INITIALIZATION_FINISHED_P 1,0
-#define GCRYCTL_INITIALIZATION_FINISHED 2
+#define GCRYCTL_INITIALIZATION_FINISHED 1
```
utoni: Why change it?

IvanNardi: `gcry_control()` and the defines `GCRYCTL_INITIALIZATION_FINISHED*` are useless when using the internal crypto code, but they are needed to be compatible with external libgcrypt.
With internal crypto, `gcry_control()`  must return true.
I reworked the code to make it clearer

utoni (approved):

PR #1870, opened 2023-01-21
Add some fuzzers to test other data structures 

IvanNardi:
Start using a dictionary for fuzzing (see: https://llvm.org/docs/LibFuzzer.html#dictionaries).
Remove some dead code.
Fuzzing with debug enabled is not usually a great idea (from performance POV). Keep the code since it might be useful while debugging.


src/lib/ndpi_main.c
```diff
@@ -3469,16 +3469,12 @@ u_int16_t ndpi_guess_protocol_id(struct ndpi_detection_module_struct *ndpi_str,
     case NDPI_IPSEC_PROTOCOL_ESP:
     case NDPI_IPSEC_PROTOCOL_AH:
       return(NDPI_PROTOCOL_IPSEC);
-      break;
```
utoni, line 3472: Why those `breaks;`s?

IvanNardi: Sorry,  are you asking why I removed them? They are dead code (just after a `return`) and no compilers (at least, the ones in CI) complains about needing such `break`

utoni: Sorry, I overlooked that those breaks are actually removed...

utoni (approved):

Issue #1873, opened 2023-01-24
Failure to detect OpenVPN traffic when using `tls-crypt`


hfz1337:
## Describe the bug
The OpenVPN dissector assumes that all packets contain an HMAC.

### Obtained behavior
When using OpenVPN in `tls-crypt` mode, control packets won't have an HMAC, like shown in the following picture:

![OpenVPN layer](https://ctrlv.link/shots/2023/01/24/hte7.png)

It would be great to add support for this mode using something like:

```diff
 #define P_SHA1_HMAC_SIZE 20
 #define P_HMAC_128 16                            // (RSA-)MD5, (RSA-)MD4, ..others
 #define P_HMAC_160 20                            // (RSA-|DSA-)SHA(1), ..others, SHA1 is openvpn default
+#define P_HMAC_NONE 0                            // No HMAC
 #define P_HARD_RESET_PACKET_ID_OFFSET(hmac_size) (9 + hmac_size)
-#define P_PACKET_ID_ARRAY_LEN_OFFSET(hmac_size)  (P_HARD_RESET_PACKET_ID_OFFSET(hmac_size) + 8)
+#define P_PACKET_ID_ARRAY_LEN_OFFSET(hmac_size)  (P_HARD_RESET_PACKET_ID_OFFSET(hmac_size) + (hmac_size == P_HMAC_NONE ? 0 : 8))
 #define P_HARD_RESET_CLIENT_MAX_COUNT  5
 
 static 
@@ -75,6 +76,9 @@ int8_t check_pkid_and_detect_hmac_size(const u_int8_t * payload) {
   
   if(get_packet_id(payload, P_HMAC_128) == 1)    
     return P_HMAC_128;
+
+  if((get_packet_id(payload, P_HMAC_NONE) & 0xffffff) == 0)
+    return P_HMAC_NONE;
   
   return(-1);
 }
@@ -90,7 +94,7 @@ void ndpi_search_openvpn(struct ndpi_detection_module_struct* ndpi_struct,
   int8_t failed = 0;
   /* No u_ */int16_t ovpn_payload_len = packet->payload_packet_len;
   
-  if(ovpn_payload_len >= 40) {
+  if(ovpn_payload_len >= 16) {
     // skip openvpn TCP transport packet size
     if(packet->tcp != NULL)
       ovpn_payload += 2, ovpn_payload_len -= 2;;
[...]

PR #1874, opened 2023-01-24
OPENVPN: add detection of flows without HMAC

IvanNardi:
Partial fix of #1873


IvanNardi:
TODO: the part about missing Remote-Session-Id (see https://github.com/ntop/nDPI/issues/1873)

utoni:
I think we can merge this even w/o the remote sid fix as it's still an improvement.
What do you think, @IvanNardi ?

IvanNardi:
I completely forgot about that... let me check...

PR #1876, opened 2023-01-25
fuzz: add a new fuzzer to test serialization/deserialization code

IvanNardi:


utoni (approved):
I am still not a fan of checking in all those binary files. But I get the reason behind that.

PR #1878, opened 2023-01-26
ndpiReader: fix packet dissection (CAPWAP and TSO)

IvanNardi:
Fix decapsulation of CAPWAP; we are interested only in "real" user data tunneled via CAPWAP.
When Tcp Segmentation Offload is enabled in the NIC, the received packet might have 0 as "ip length" in the IPv4 header
(see
https://osqa-ask.wireshark.org/questions/16279/why-are-the-bytes-00-00-but-wireshark-shows-an-ip-total-length-of-2016/)

The effect of these two bugs was that some packets were discarded.

Be sure that flows order is deterministic


utoni (commented):
`capwap_data.pcapng` fails in the CI. But on do not get the issue. There is a diff, but it looks quite similar to the original.

IvanNardi:
Merging, so it is included in the next release

PR #1879, opened 2023-01-27
Further reduction of the size of some traces used as unit test

IvanNardi:
See a944514d. No flow/classification/metadata have been removed.


IvanNardi:
Merging, so it is included in the next release

PR #1882, opened 2023-02-05
Add Yandex services detection

0xA50C1A1:
Yandex services are quite popular in Eastern Europe, so I think it would be good to have their detection support in nDPI.


IvanNardi:
Everything seems fine. Just two notes:
1) could you please:
  * merge all these traces into one (called `yandex.pcpang`) and push only this one? You can try using `mergecap`
  * while doing the previous point, please keep only the first ~20/25 pkts for sessions (i.e. just a little after the TLS handshake ends)   The idea is to have small(er) pcap in the repository, with only the "interesting" stuff (for a simple TLS matching , that means only TCP handshake + TLS handshake)

2) Update the ASN list:
 * AS44534, AS207207, AS202611, AS13238 --> NDPI_PROTOCOL_YANDEX
 * AS210656, AS200350 --> NDPI_PROTOCOL_YANDEX_CLOUD

0xA50C1A1:
@IvanNardi Thanks for the review, I've fixed it.

IvanNardi (approved):

IvanNardi:
Thanks!

PR #1885, opened 2023-02-12
Fix export/serialization of `flow->risk`

IvanNardi:
Any risks bigger than 31 was never exported/serialized; found looking at the fuzz coverage report.


IvanNardi:
I am quite surprised that this bug has not been discovered sooner...

utoni (approved):
Good catch!

PR #1888, opened 2023-02-13
fuzz: extend fuzz coverage

IvanNardi:


IvanNardi:
Merging. With this last batch, the bulk of the fuzzing work should have been done

PR #1891, opened 2023-02-16
ndpiReader: print how many packets (per flow) were needed to perform full DPI

IvanNardi:
Average values are already printed, but this change should ease to identify regressions/improvements.


utoni (approved):
A feature I've wanted to see.
I would also like to have a counter that prints how many dissector calls per protocol were required to set the detection or exclude a protocol.

PR #1892, opened 2023-02-16
Add a new flow risk about literal IP addresses used as SNI

IvanNardi:
RFC 6066 3: "Literal IPv4 and IPv6 addresses are not permitted in "HostName"."

Don't set this risk if we have a valid sub-classification (example: via certificate)

Since a similar risk already exists for HTTP hostnames, reuse it, with a
more generic name.


utoni (approved):

lucaderi:
@IvanNardi @utoni What about making NDPI_HTTP_NUMERIC_IP_HOST more general?

IvanNardi:
> @IvanNardi @utoni What about making NDPI_HTTP_NUMERIC_IP_HOST more general?

Fine for me. Something like `NDPI_NUMERIC_IP_HOST`? @lucaderi

utoni:
> > @IvanNardi @utoni What about making NDPI_HTTP_NUMERIC_IP_HOST more general?
> 
> Fine for me. Something like `NDPI_NUMERIC_IP_HOST`? @lucaderi

Makes sense.

PR #1896, opened 2023-03-02
fuzz: simplify fuzzers dependencies in CIFuzz

IvanNardi:
CIFuzz (based on oss-fuzz) is the GitHub action/CI job that runs fuzz
targets on pull requests. It only runs the fuzzers affected by a pull
request/commit. Otherwise it will divide up the allotted fuzzing time
among all fuzzers in the project.
Since:
* we have more than 20 fuzzers and most of them use the custom memory
allocation functions (to force allocation failures) even if they are not
strictly about DPI stuff;
* we need to keep fuzzing time relatively small (to avoid waiting for the CI
results for a long time)

it is important that fuzzers dependencies (which are based on *files*
changed by the single commit/PR) are as small as possible.

Bottom line: move all the low-level allocation callbacks to a dedicated
file; this way most of the fuzzers don't depend anymore on `ndpi_main.c`
file (which is touched by every commit/PR).

The goal is to have only the "most important" fuzzers running during (most
of) the CI.

### DIFF
--- a/src/lib/ndpi_main.c
+++ b/src/lib/ndpi_main.c
@@ -124,9 +124,6 @@ u_int ndpi_verbose_dga_detection = 0;
 static void *(*_ndpi_flow_malloc)(size_t size);
 static void (*_ndpi_flow_free)(void *ptr);
 
-static void *(*_ndpi_malloc)(size_t size);
-static void (*_ndpi_free)(void *ptr);
-
 /* ****************************************** */
 
 static ndpi_risk_info ndpi_known_risks[] = {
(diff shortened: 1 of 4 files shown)


utoni (commented):
I do not know what the intention of those flow memory wrappers were in the first place. But in my imagination, it could be useful e.g. for fixed size memory pools.

I am not sure if this feature is used by developers. But I think it has it's justification.

utoni (approved):

IvanNardi:
> I do not know what the intention of those flow memory wrappers were in the first place. But in my imagination, it could be useful e.g. for fixed size memory pools.
> 
> I am not sure if this feature is used by developers. But I think it has it's justification.

Ok, I'll push a new version restoring the flow wrappers

PR #1903, opened 2023-03-17
Rework CI jobs to try reducing CI duration

IvanNardi:
CI duration is quite long: the longest jobs is the "Performance" one.
Try to reduce the overall duration: that job (and some others) will not
be triggered for each PR/commit anymore, but asynchronously, once a day
(this scheduling seems right since the frequency of the PR/commits in
the project).
It should be possibly to trigger them manually, via GUI, anyway.

Remove two identical jobs; we already tests ASAN with 4 different
compilers.

After 9eff075419cbb0fe2758d7f8505359e3130e6df6 it is safe to reduce fuzzing time.

Bottom line: try to have as upper-time of CI tests the duration of the
fuzzing jobs

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -10,40 +10,6 @@ on:
   release:
     types: [created]
 jobs:
-  coverage:
-    name: Coverage (ubuntu-latest)
-    runs-on: ubuntu-latest
-    env:
-      CFLAGS: -Werror
-    steps:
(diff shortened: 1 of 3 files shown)


IvanNardi:
@utoni , what do you think?

utoni:
It's sufficient to run this once per day. I guess even once per week or before a planned release.

utoni (approved):

PR #1904, opened 2023-03-17
Add a new protocol id for generic advertisement/analytics/tracking stuff

IvanNardi:


IvanNardi:
This id is conceptually similar to the existing `NDPI_PROTOCOL_CYBERSECURITY`

utoni (approved):

PR #1909, opened 2023-03-21
Make Bittorrent LRU cache IPv6 aware.

IvanNardi:
It was the only remaining LRU cache without IPv6 support. See 81e1ea545ca465cda064e7cc80333fe7f0ef2aff


src/lib/ndpi_main.c
```diff
@@ -7768,9 +7765,20 @@ ndpi_protocol ndpi_guess_undetected_protocol(struct ndpi_detection_module_struct
       }
     }
 
-    if(ndpi_search_into_bittorrent_cache(ndpi_str, NULL /* flow */,
-					 htonl(shost), htons(sport),
-					 htonl(dhost), htons(dport))) {
+    /* Workaround: create a fake flow struct since ndpi_search_into_bittorrent_cache is ipv4/6 agnostic,
+       while this function is intrinsecally ipv4 only.
+       This way we don't need to change the (public) ndpi_guess_undetected_protocol() prototype */
```
utoni: Why not changing the public api? We're doing that so often. Why not this time? The result would be cleaner code. :)

IvanNardi: I will talk with @lucaderi about that!

utoni (approved):

PR #1900, opened 2023-03-14
Add support for vxlan decapsulation (#1441)

mmaatuq:
* decapsulate vxlan tunnel. 
* update vxlan.pcap.out with the new outp after detunneling capability.

PR #1911, opened 2023-03-22
remove redefinition to vxlanhdr struct in vxlan dissector

mmaatuq:
* vxlanhdr struct was redefined in vxlan dissector
* in this [PR](https://github.com/ntop/nDPI/pull/1900) @utoni  asked me to remove a redefinition of vxlanhdr struct.

### DIFF
--- a/src/include/ndpi_typedefs.h
+++ b/src/include/ndpi_typedefs.h
@@ -532,7 +532,8 @@ PACK_ON
 struct ndpi_vxlanhdr {
   u_int16_t flags;
   u_int16_t groupPolicy;
-  u_int32_t vni;
+  u_int8_t vni[3];
+  u_int8_t reserved;
 } PACK_OFF;
 
 /* ************************************************************ */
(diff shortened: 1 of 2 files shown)


IvanNardi:
Instead of accessing the data byte-by-byte, I think it would be cleaner if you use the structure already defined in `src/include/ndpi_typedefs.h` (updating it, if you think so)

mmaatuq:
I thought we could redefine the struct as per rfc-7348 
```c
struct ndpi_vxlanhdr{
    u_int8_t flags;
    u_int8_t reserved1[3];
    uint_8_t vni[3];
    uint_8_t reserved2;
};
```
after looking at some references, and how they defined rfc-7348 vxlanhdr 
[linux kernel](https://github.com/torvalds/linux/blob/master/include/net/vxlan.h) 
 ```c
struct vxlanhdr 
    __be32 vx_flags;
    __be32 vx_vni;
};
```

 [wireshark](https://gitlab.com/wireshark/wireshark/-/blob/master/epan/dissectors/packet-v52.c) 
       there is not direct definition for vxlanhdr but from the protocol dissector 
```c
struct vxlanhdr {
    u_int16_t flags;
    u_int16_t group_policy;
    u_int8_t vni[3];
    uint_8_t reserved;
};
```
The last definition is more convenient for nDPI (classification)

IvanNardi (approved):

utoni (approved):

IvanNardi:
@mmaatuq, thanks!

PR #1922, opened 2023-03-29
Ookla: rework detection

IvanNardi:
The logic of the LRU cache has been changed: once we know an ip has connected to an Ookla server, all the following (unknown) flows (for a short time interval) from the same ip to the port 8080 are treated as Ookla ones.

Most of the changes in this commit are about introducing the concept of "aggressive detection". In some cases, to properly detect a protocol we might use some statistical/behavior logic that, from one side, let us to identify the protocol more often but, from the other side, might lead to some false positives.
To allow the user/application to easily detect when such logic has been triggered, the new confidence value `NDPI_CONFIDENCE_DPI_AGGRESSIVE` has been added.
It is always possible to disable/configure this kind of logic via the API.

Detection of Ookla flows using plain TLS over port 8080 is the first example of aggressive detection in nDPI.

Tested with:
* Android 9.0 with app 4.8.3
* Ubuntu 20.04 with Firefox 110
* Win 10 with app 1.15 and 1.16
* Win 10 with Chrome 108, Edge 108 and Firefox 106


example/ndpiReader.c
```diff
@@ -2472,6 +2472,8 @@ static void setupDetection(u_int16_t thread_id, pcap_t * pcap_handle) {
 			  NDPI_LRUCACHE_BITTORRENT, 32768);
   /* Enable/disable LRU caches TTL here */
 
+  /* Set aggressiviness here */
```
utoni, line 2510: I would suggest to provide a command line option instead.

IvanNardi: Done. Added a generic `-Z proto_name:aggressiveness_value` to `ndpiReader`

utoni (approved):
Great job!

PR #1927, opened 2023-04-02
Global context and global/shared LRU caches

IvanNardi:


IvanNardi:
I will push an updated version

PR #1931, opened 2023-04-03
Test multiple `ndpiReader` configurations

IvanNardi:
Extend internal unit tests to handle multiple configurations. As some examples, add tests about:
* disabling some protocols
* disabling Ookla aggressiveness

Every configurations data is stored in a dedicated directory under
 `tests\cfgs`


IvanNardi:
@utoni, is there a better way to handle all these relative paths?

utoni (approved):
This might be coming handy.

utoni:
> @utoni, is there a better way to handle all these relative paths?

dunno any better way

PR #1947, opened 2023-04-17
TLS: ignore padding bytes in ACK packets.

vel21ripn:
Some short ACK packets have additional null padding bytes at the end of the packet. The TLS parser should ignore them.

### DIFF
--- a/src/lib/protocols/tls.c
+++ b/src/lib/protocols/tls.c
@@ -960,6 +960,22 @@ static void ndpi_looks_like_tls(struct ndpi_detection_module_struct *ndpi_struct
     flow->guessed_protocol_id = __get_master(ndpi_struct, flow);
 }
 
+static int tcp_ack_padding(struct ndpi_packet_struct *packet) {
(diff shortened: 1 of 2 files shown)


src/lib/protocols/tls.c
```diff
@@ -977,7 +993,8 @@ static int ndpi_search_tls_tcp(struct ndpi_detection_module_struct *ndpi_struct,
      generic "extra function" code doesn't honour protocol bitmask.
      TODO: handle that in ndpi_main.c for all the protocols */
   if(packet->payload_packet_len == 0 ||
-     packet->tcp_retransmission) {
+     packet->tcp_retransmission ||
+     tcp_ack_padding(packet)) {
```
utoni, line 997: Does the TCP ACK padding only occur for TLS?

vel21ripn: In the process of being clarified.

vel21ripn: Statistics showed that padding does not depend on the application protocol.

PR #1956, opened 2023-04-28
Improve detection of crawler/bot traffic

IvanNardi:


utoni (approved):
Good job! I would also like to a have such a list for bad/malicious HTTP URLs that are known for malware content or exploits.

PR #1959, opened 2023-04-29
HTTP: improve extraction of metadata and of flow risks

IvanNardi:


IvanNardi:
HTTP dissector is probably the most complex one, for historical reasons...
It is not perfect yet (there are some issues with flows starting with a response), but with this change a lot more metadata and flow risks are extracted.
In the unit tests results, the diffs show more information everywhere (with only one exception; see below) 
Next step: more clean separation among:
* protocol identification (request vs reply)
* lines parsing
* sub-classification

tests/cfgs/default/result/KakaoTalk_talk.pcap.out
```diff
@@ -47,7 +47,7 @@ JA3 Host Stats:
...
 	9	TCP 10.24.82.188:51021 <-> 103.246.57.251:8080 [proto: 131/HTTP_Proxy][IP: 0/Unknown][ClearText][Confidence: Match by port][DPI packets: 11][cat: Web/5][6 pkts/543 bytes <-> 5 pkts/945 bytes][Goodput ratio: 25/64][24.77 sec][bytes ratio: -0.270 (Download)][IAT c2s/s2c min/avg/max/stddev: 77/47 4920/8061 17431/17434 6679/7163][Pkt Len c2s/s2c min/avg/max/stddev: 68/68 90/189 130/504 24/164][Plen Bins: 16,51,0,16,0,0,0,0,0,0,0,0,0,16,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0]
```
IvanNardi, line 51: This should be the only test where we have less information than before. However the sequence number of the HTTP reply is clearly "wrong" and that packet is now correctly identified as retransmission (and Wireshark agrees). Bottom line: previous code extracted more info only by chance

src/lib/protocols/http.c
```diff
@@ -401,53 +405,127 @@ static void setHttpUserAgent(struct ndpi_detection_module_struct *ndpi_struct,
...
+    ndpi_protocol_match_result ret_match;
+    char *ptr, *origin, *origin_hostname;
+
+    /* Origin syntax:
+        Origin: null
+        Origin: <scheme>://<hostname>
+        Origin: <scheme>://<hostname>:<port>
+    Try extracting hostname */
+
+    origin = ndpi_malloc(packet->http_origin.len + 1);
```
utoni: Is it really required to use dynamic memory here?

src/lib/protocols/http.c
```diff
@@ -886,54 +899,67 @@ static void check_content_type_and_change_protocol(struct ndpi_detection_module_
...
-          OCSP:  application/ocsp-request, application/ocsp-response
-       */
-      if(packet->content_line.len > 17 &&
-         strncmp((const char *)packet->content_line.ptr, "application/ocsp-", 17) == 0) {
-        NDPI_LOG_DBG2(ndpi_struct, "Found OCSP\n");
-        ndpi_set_detected_protocol(ndpi_struct, flow, NDPI_PROTOCOL_OCSP, NDPI_PROTOCOL_HTTP, NDPI_CONFIDENCE_DPI);
+      double_col = strchr((char*)flow->host_server_name, ':');
+      if(double_col) double_col[0] = '\0';
+      if(ndpi_struct->packet.iph
+         && (sscanf(flow->host_server_name, "%d.%d.%d.%d", &a, &b, &c, &d) == 4)) {
```
utoni, line 931: This format string does also accept invalid IPv4 addresses. Maybe add a IPv4 address validation method and set a risk if the address is invalid?

src/lib/protocols/http.c
```diff
@@ -1138,6 +1164,47 @@ static void ndpi_check_http_header(struct ndpi_detection_module_struct *ndpi_str
...
+    /* https://en.wikipedia.org/wiki/List_of_HTTP_status_codes */
+    if((flow->http.response_status_code < 100) || (flow->http.response_status_code > 509))
+      flow->http.response_status_code = 0; /* Out of range */
+
+    if(flow->http.response_status_code >= 400) {
+      snprintf(ec, sizeof(ec), "HTTP Error Code %u", flow->http.response_status_code);
+      ndpi_set_risk(ndpi_struct, flow, NDPI_ERROR_CODE_DETECTED, ec);
+    }
+
+    if(flow->http.response_status_code >= 400) {
```
utoni: Can be merged with the if-statement above.

utoni (changes requested):

src/lib/protocols/http.c
IvanNardi: No, it isn't...

src/lib/protocols/http.c
IvanNardi: We extract the address only to compare it with the address in the IP header to set the NDPI_HTTP_SUSPICIOUS_HEADER risk. An invalid address will trigger this risk even more so

src/lib/protocols/http.c
IvanNardi: I'll do

src/lib/protocols/http.c
utoni: Not sure if it is worth to replace it with a stack allocation. If I remember correctly allocating a variable sized buffer on the stack is a GNu extension, so MSVC may complain here.

IvanNardi: In the new version I simply used the original buffer, without any temporary/intermediate buffer

utoni (approved):
LGTM. Awesome, perfect solution!

PR #1965, opened 2023-05-05
Remove special handling of some TCP flows without SYN

IvanNardi:
This piece of code has multiple problems:
* nDPI is able to detect some TCP protocols even with mid-flows (i.e. without the initial packets of the session); TLS is the most significative example
* since e6b332aa4a1399e33df68998cf8351bccaee3fc4 it is perfectly valid to not pass the TCP Handshake packets to nDPI
* in any case, we shouldn't call `ndpi_detection_giveup()`. That function is usually called by the application and we end up calling it twice in some cases.

The simple solution is to completely remove that code: process these kinds of flows like everyone else.

Note that the application can always avoid to pass to nDPI any TCP flows without the initial handshake; the flow managemnt is always up to the application.

Looking at the CI results, some rare flows are now processed significantly longer. As a follow-up we could look into that.


utoni (approved):
Strange. Some flows require 20~30 more packets till classification.

IvanNardi:
> Strange. Some flows require 20~30 more packets till classification.

Yeah, I noticed (and reported it in the commit message).
Basically those flows pass from 1 pkt to the upper limit 33 (see `NDPI_DEFAULT_MAX_NUM_PKTS_PER_FLOW_TO_DISSECT` ;we process one more pkt than expected...)
Where you see 20, it is basically because the flow is made by only 20 pkts...

Patches will be pushed soon to reduce those numbers

PR #1966, opened 2023-05-06
HTTP: rework state machine

IvanNardi:
The goal if to correlate the right request-response pair, exporting
metadata from only one transaction (for example, the right url & return
state pair)

As a nice side effect, the code should be much cleaner, but that is a
matter of taste.
    
Two differences respect to the previous code:
* as it happens in the CI, if in the flow there are only one response
(before) and one request (after), only the metadata of the response are
saved/exported
* for performance reasons, we don't call `ndpi_parse_packet_line_info()`
anymore for ALL packets triggering the HTTP dissector, but only for the
packets that we already know belong to an HTTP flow. This is the reason
for the changes in RTSP/SOAP/... code


tests/cfgs/default/result/1kxun.pcap.out
```diff
@@ -83,7 +83,7 @@ JA3 Host Stats:
...
+	37	TCP 192.168.2.126:45422 <-> 161.117.13.29:80 [proto: 7.295/HTTP.1kxun][IP: 0/Unknown][ClearText][Confidence: DPI][DPI packets: 5][cat: Streaming/17][3 pkts/2139 bytes <-> 4 pkts/6060 bytes][Goodput ratio: 91/96][4.31 sec][Hostname/SNI: mangaweb.1kxun.mobi][bytes ratio: -0.478 (Download)][IAT c2s/s2c min/avg/max/stddev: 221/224 2062/1374 3902/3898 1841/1787][Pkt Len c2s/s2c min/avg/max/stddev: 502/1413 713/1515 819/1720 149/124][URL: mangaweb.1kxun.mobi/images/detail_revision/go_homepage.png][StatusCode: 200][Content-Type: image/png][Server: openresty/1.13.6.1][User-Agent: Mozilla/5.0 (Linux; Android 11; sdk_gphone_x86 Build/RSR1.201013.001; wv) AppleWebKit/537.36 (KHTML, like Gecko) Version/4.0 Chrome/83.0.4103.106 Mobile Safari/537.36][PLAIN TEXT (GET /js/application.min.j)][Plen Bins: 0,0,0,0,0,0,0,0,0,0,0,0,0,14,0,0,0,0,0,0,0,0,0,28,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,28,0,0,14,0,14]
```
utoni, line 86: Why did the URL change? This looks odd.

tests/cfgs/default/result/fuzz-2006-09-29-28586.pcap.out
```diff
@@ -30,15 +30,15 @@ Cloudflare	1	854	1
...
+	2	TCP 172.20.3.5:2606 <-> 172.20.3.13:80 [proto: 7/HTTP][IP: 0/Unknown][ClearText][Confidence: DPI][DPI packets: 12][cat: Web/5][8 pkts/2287 bytes <-> 5 pkts/2963 bytes][Goodput ratio: 80/91][11.18 sec][Hostname/SNI: 172.20.3.13][bytes ratio: -0.129 (Mixed)][IAT c2s/s2c min/avg/max/stddev: 0/0 58/58 177/172 83/81][Pkt Len c2s/s2c min/avg/max/stddev: 60/54 286/593 1514/1514 478/662][URL: 172.20.3.13/servlets/mms?message-id=189301][Risk: ** HTTP/TLS/QUIC Numeric Hostname/SNI **][Risk Score: 10][Risk Info: Found host 172.20.3.13][PLAIN TEXT (GET /servlets/mms)][Plen Bins: 0,0,0,0,0,0,0,0,0,0,0,25,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,25,0,0,0,0,0,0,50,0,0]
```
utoni: There is no UA set, but the risk is gone. Is that the expected behavior?

tests/cfgs/default/result/fuzz-2006-09-29-28586.pcap.out
```diff
@@ -47,7 +47,7 @@ AmazonAWS	1	477	1
...
-	19	TCP 172.20.3.5:80 -> 172.57.3.13:53132 [proto: 7/HTTP][IP: 0/Unknown][ClearText][Confidence: Match by port][DPI packets: 1][cat: Web/5][1 pkts/195 bytes -> 0 pkts/0 bytes][Goodput ratio: 72/0][< 1 sec][Server: Microsoft-IIS/4.0][Risk: ** Unidirectional Traffic **][Risk Score: 10][Risk Info: No client to server traffic][PLAIN TEXT (HTTo/1.1 202 Accepted)][Plen Bins: 0,0,0,0,100,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0]
+	19	TCP 172.20.3.5:80 -> 172.57.3.13:53132 [proto: 7/HTTP][IP: 0/Unknown][ClearText][Confidence: Match by port][DPI packets: 1][cat: Web/5][1 pkts/195 bytes -> 0 pkts/0 bytes][Goodput ratio: 72/0][< 1 sec][Risk: ** Unidirectional Traffic **][Risk Score: 10][Risk Info: No client to server traffic][PLAIN TEXT (HTTo/1.1 202 Accepted)][Plen Bins: 0,0,0,0,100,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0]
```
utoni, line 50: Possible information loss?

tests/cfgs/default/result/1kxun.pcap.out
IvanNardi: The main goal of this PR is to export the "right" metadata for the first request-response pair.
Usually in a HTTP flow we have something like REQ1-RESP1-REQ2-RESP2-... and we export REQ1-RESP1 metadata. Easy.

Sometime, especially in the local traces or in the CI, we have RESP1-REQ2-RESP2 or (like in that case) REQ1-REQ2-RESP2 (because of missing pkts, for example).
Before this change, the metadata were of REQ1-RESP2, now they are of REQ2-RESP2 (correctly): so the url changed.
It is quite obvious that the old information were probably wrong since in the url there was a js file, but the reply content type was png...

tests/cfgs/default/result/fuzz-2006-09-29-28586.pcap.out
IvanNardi: It is expected from me, but let's discuss about that!
The idea is to set the UA risk only if we *found* a response without an UA; in other words: if there is no response (like in that case), we don't know anything about the ua and so we don't set the risk (and, in general, the metadata/risks exported are only about the request)
Is it reasonable? What do you think?

tests/cfgs/default/result/fuzz-2006-09-29-28586.pcap.out
IvanNardi: it depends on how you look at it :)
This flow is NOT classified as HTTP (the classification is only by port, so pretty much meaningless...) so I think that we should not set any HTTP metadata for flows that are not HTTP.

The core reason is that we now call `ndpi_parse_packet_line_info()` (and set the metadata that we find "parsing the lines") only for HTTP flows, while before we called it pretty much for every TCP flows, even if we knew that they were not HTTP.
I think that the new behavior is better, but I don't have strong opinions on that

tests/cfgs/default/result/fuzz-2006-09-29-28586.pcap.out
utoni: But isn't the UA part of a Request, not a Response?

IvanNardi: Never reply if you haven't drunk enough coffee and always re-read what you have just wrote:  lesson learned :-D
Let me take another look at the issue...

utoni (approved):
Good job, @IvanNardi <3

IvanNardi:
> Good job, @IvanNardi <3

Thanks to you for the multiple reviews!

PR #1969, opened 2023-05-09
All protocols should be excluded sooner or later

IvanNardi:
For a lot of protocols, reduce the number of packets after which the protocols dissector gives up.
The values are quite arbitary, tring to not impact on classification


utoni (approved):
This has hopefully no impact on the classification for real-time traffic. :)

IvanNardi:
I have been running that code for a while now, without any specific issues.
But I must admit that most of these protocols are not really so used/common on the networks I have access to

PR #1973, opened 2023-05-11
proplem build version 2

rezasafaei70:
config.status: error: cannot find input file: `Makefile.in'


IvanNardi:
It is @lucaderi call, but I am pretty sure that this kind of change will be never committed to a so old stable branch (even with all these conflicts fixed)
What is your real problem? Why can't you use latest nDPI version?

IvanNardi:
I don't know why the last 2 PRs just merged ended here, too....

utoni:
> I don't know why the last 2 PRs just merged ended here, too....

The source branch is set to `dev`.

utoni:
I am not sure if this PR should merged. There is a reason why developers use versioning. IMHO, backporting everything besides bugfixes does not seem useful, because you can just upgrade to a newer version as @IvanNardi mentioned. Anyway, `2.4` is quite old and should not used anymore.

utoni:
I am closing this PR now to safe CI resources. Please re-open if necessary.

PR #1974, opened 2023-05-11
Added printf/fprintf replacement for some internal modules.

utoni:
* logging is instead redirected to `ndpi_debug_printf`

### DIFF
--- a/src/lib/ndpi_main.c
+++ b/src/lib/ndpi_main.c
@@ -26,6 +26,10 @@
 #include <sys/types.h>
 #include <dirent.h>
 
+#ifdef __APPLE__
+#include <netinet/ip.h>
+#endif
+
 #define NDPI_CURRENT_PROTO NDPI_PROTOCOL_UNKNOWN
 
 #include "ndpi_config.h"
@@ -36,7 +40,7 @@
 #ifdef USE_HOST_LIBGCRYPT
 #include <gcrypt.h>
 #else
-#include <gcrypt_light.h>
+#include "gcrypt_light.h"
 #endif
 
 #include <time.h>
@@ -122,10 +126,7 @@
 #include "nbpf.h"
 #endif
 
-/* #define DGA_DEBUG 1 */
 /* #define MATCH_DEBUG 1 */
-
-u_int ndpi_verbose_dga_detection = 0;
 
 /* ****************************************** */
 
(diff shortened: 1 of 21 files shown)


src/lib/ndpi_main.c
```diff
@@ -8611,26 +8620,26 @@ void ndpi_generate_options(u_int opt) {
...
 
-void ndpi_dump_risks_score() {
+void ndpi_dump_risks_score(FILE *risk_out) {
   u_int i;
 
-  printf("%3s %-48s %-8s %s %-8s %-8s\n",
+  fprintf(risk_out, "%3s %-48s %-8s %s %-8s %-8s\n",
 	 "Id", "Risk", "Severity", "Score", "CliScore", "SrvScore");
 
   for(i = 1; i < NDPI_MAX_RISK; i++) {
```
IvanNardi, line 8819: A missing `printf->fprintf(risk_out ` below

utoni: Oh, forgot to change it.

I just saw that `ndpi_check_dga_name` uses lots of `printf()`s. But this function is mainly invoked by the dissectors. Do you think it makes sense to replace it with either `NDPI_LOG_DBG` or `NDPI_LOG_DBG2` ?

utoni: or use `#ifdef NDPI_ENABLE_DEBUG_MESSAGES` for all `printf()`s as done in other translation units

IvanNardi: It is a little bit tricky because we want to be still able to enable/disable these logs via command line parameter in `tests/dga/dga_evaluate.c`...

utoni: I'll try to use debug printf callback and see if it works as expected.
I think that we are wasting performance doing I/O stuff in dissectors especially for TLS/QUIC  which are widely used.

// EDIT: This does of course only apply if `ndpi_verbose_dga_detection` is enabled.
But I think it is still better to move to the debug printf callback.

utoni:
I've decided to use a similar approach as syslog does. Do do that, I've introduced a logging facility which allows the logging callback to distinguish between different nDPI components.

What's your opinion on that, @IvanNardi ?

The long term goal should be to get rid of all `*printf()`s even those that are currently replaced and use only the logging callback together with logging facilities.

IvanNardi (commented):
Two notes:
* we can probably remove the define `DGA_DEBUG` from `ndpi_main.c` and convert the printf in `ndpi_ends_with` with a standard `NDPI_LOG_DBG2` (not a `NDPI_LOG_DGA` because it is used also by HTTP code)
* there are still some `if(ndpi_verbose_dga_detection)` that can probably be removed

Other than that, it seems ok to me. Adding the facility might be a little overwhelming right now, but it serves its scope and it might be useful in future.

IvanNardi:
@utoni, any updates on this PR?

utoni:
LGTM 👍

utoni:
I kind of forgot about the PR and the remaining ToDo.

src/lib/ndpi_main.c
```diff
@@ -8788,26 +8794,26 @@ void ndpi_generate_options(u_int opt) {
...
```
IvanNardi, line 8819: Still a missing "fprintf(risk_out" just below...

utoni: sooorry

IvanNardi (approved):
Thanks!

Issue #1957, opened 2023-04-28
Question about using "guessed_protocol_id"


vel21ripn:
Consider a fairly typical option: a tcp connection to port 443.
(This is a very frequent situation for the Telegram protocol)

In this case "guessed_protocol" would be "TLS". Based on the ip address was determined "guessed_protocol_id_by_ip=Telegram".

On the first packet with payload, we determine that it is not "TLS" and it is excluded from further checks.
If the parsing of the tcp stream data failed, then the final result will be "confidence: match by ip" and "proto: TLS.Telegram", which is actually not correct, since "TLS" is missing in this case.

As a solution, you can offer this option:
do not result in "guessed_protocol" if it was excluded (flow->excluded_protocol_bitmask bitmap).

And another question: is the ndpi_detection_giveup() function in the ndpi_main.c file
````
  if((flow->guessed_protocol_id_by_ip != NDPI_PROTOCOL_UNKNOWN)
     && ((ret.app_protocol == NDPI_PROTOCOL_UNKNOWN) || (ret.master_protocol == NDPI_PROTOCOL_UNKNOWN))) {

    if(ret.app_protocol == NDPI_PROTOCOL_UNKNOWN)
      ndpi_int_change_protocol(ndpi_str, flow,
                               flow->guessed_protocol_id_by_ip, ret.master_protocol,
                               NDPI_CONFIDENCE_MATCH_BY_IP);
    else
      /* master_protocol == NDPI_PROTOCOL_UNKNOWN) */
      ndpi_int_change_protocol(ndpi_str, flow,
                               flow->guessed_protocol_id_by_ip, ret.app_protocol,
                               NDPI_CONFIDENCE_DPI_PARTIAL);
````
Why is NDPI_CONFIDENCE_DPI_PARTIAL used instead of NDPI_CONFIDENCE_MATCH_BY_IP?

vel21ripn:
@IvanNardi  Please look at the traffic examples from the directory "utils/tcp_check_seq/samples_guessed_protocol" in the repository "vel21ripn/nDPI" branch "tcp_ack_padding". There are full debug output dumps (-V 4).

Note that the TLS protocol is excluded after the first packet containing data has been processed.

It is possible that there is a typo in ndpi_detection_giveup(). NDPI_CONFIDENCE_DPI_PARTIAL is specified instead of NDPI_CONFIDENCE_MATCH_BY_IP.

IvanNardi:
[...]

PR #1981, opened 2023-05-16
Fix classification-by-ip in `ndpi_detection_giveup`

IvanNardi:
Return the "classification-by-ip" as protocol results only if no other results are available.
In particular, never return something like
"protocol_by_port/protocol_by_ip" (i.e. `NTP/Apple`, `BitTorrent/GoogleCloud`, `Zoom/AWS`) because this kind of classification is quite confusing, if not plainly wrong.

Notes:
* the information about "classification-by-ip" is always available, so no information is lost with this change;
* in the unit tests, the previous classifications with confidence `NDPI_CONFIDENCE_DPI_PARTIAL` were wrong, as noted in #1957


lucaderi:
Thank you

PR #1986, opened 2023-05-20
Improved helper scripts to be more restrictive.

utoni:


utils/whatsapp_ip_addresses_download.sh
IvanNardi, line 1: could you check the WA change?
It seems that fetching this list now fails, while before it was ok, IMO

utoni: On my local machine, the old URL was failing. I thought that Facebook may give different results depending on the UA/Referer, but that seems not to be the case here.

IvanNardi: Strange...

utoni: Reverted to the old URL/Regex.

IvanNardi (approved):

PR #1988, opened 2023-05-24
Fixed invalid use of ndpi_free(). Sorry, my fault.

utoni:


src/lib/third_party/src/gcrypt/aesni.c
```diff
@@ -97,7 +97,7 @@ int mbedtls_aesni_has_support( unsigned int what )
 	  break; /* We giveup */
       }
 
-      ndpi_free(line);
+      free(line);
```
IvanNardi: What about adding a comment like "Don't use ndpi_free here!" ? Might it be useful?

IvanNardi (approved):

PR #1993, opened 2023-05-27
Remove some useless checks

IvanNardi:
In the main dissector callbacks the flow protocols are (almost) always unknown. Only two exceptions:
* extra dissection data path
* HTTP sub-protocols

### DIFF
--- a/src/lib/protocols/ftp_control.c
+++ b/src/lib/protocols/ftp_control.c
@@ -589,80 +589,76 @@ static void ndpi_check_ftp_control(struct ndpi_detection_module_struct *ndpi_str
 				   struct ndpi_flow_struct *flow) {
   struct ndpi_packet_struct *packet = &ndpi_struct->packet;
   u_int32_t payload_len = packet->payload_packet_len;
-
-  /* Check connection over TCP */
-  if(packet->tcp) {
-    u_int16_t twentyfive = htons(25);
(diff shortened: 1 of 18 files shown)


src/lib/protocols/ftp_control.c
```diff
@@ -589,80 +589,76 @@ static void ndpi_check_ftp_control(struct ndpi_detection_module_struct *ndpi_str
...
-	At first check, if this is for sure a response packet (in another direction.
-	If not, do nothing now and return. 
+      /* 
+	Encode the direction of the packet in the stage, so we will know when we need
+	to look for the response packet. 
       */
-      if((flow->ftp_control_stage - packet->packet_direction) == 1) {
-	return;
-      }
+      flow->ftp_control_stage = packet->packet_direction + 1;
```
utoni, line 619: Does this line alter the behavior, because you now have two stages: 1 and 2. Or am I wrong?

IvanNardi: I think this is an artifact due to how the diffs are displayed.
If you hide the whitespaces changes in the Github UI you get that that code didn't change (as expected)
Screenshot from 2023-05-28 12-27-10

utoni: You're right. Github's `diff` is not .. optimal in terms on displaying such ..

utoni (approved):

PR #1995, opened 2023-05-29
Fix some memory errors triggered by allocation failures

IvanNardi:
Some low hanging fruits found using nallocfuzz.
See: https://github.com/catenacyber/nallocfuzz
See: https://github.com/google/oss-fuzz/pull/9902

Most of these errors are quite trivial to fix; the only exception is the stuff in the uthash.
If the insertion fails (because of an allocation failure), we need to avoid some memory leaks. But the only way to check if the `HASH_ADD_*` failed, is to perform a new lookup: a bit costly, but we don't use that code in any critical data-path.

### DIFF
--- a/example/reader_util.c
+++ b/example/reader_util.c
@@ -116,14 +116,14 @@ u_int16_t max_pattern_len = 8;
 
 /* *********************************************************** */
 
-void ndpi_analyze_payload(struct ndpi_flow_info *flow,
-			  u_int8_t src_to_dst_direction,
-			  u_int8_t *payload,
-			  u_int16_t payload_len,
-			  u_int32_t packet_id) {
-  struct payload_stats *ret;
-  struct flow_id_stats *f;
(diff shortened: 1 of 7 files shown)


example/reader_util.c
```diff
@@ -121,9 +121,9 @@ void ndpi_analyze_payload(struct ndpi_flow_info *flow,
 			  u_int8_t *payload,
 			  u_int16_t payload_len,
 			  u_int32_t packet_id) {
```
utoni: Maybe print a warning or error if allocation fails, because there is no return value that marks a failure.

utoni (changes requested):

example/reader_util.c
IvanNardi: It makes sense. I'll do

IvanNardi: Done in latest revision

utoni (approved):
Thanks!

PR #2009, opened 2023-06-12
Fix/missing typedefs and strncasecmp replacement

utoni:


IvanNardi:
@utoni, just out of curiosity, could you share the platforms where you found this warning?

utoni:
It happens when I run https://github.com/utoni/nDPId/blob/main/scripts/get-and-build-libndpi.sh locally and in a GitLab instance.

IvanNardi (approved):

PR #2016, opened 2023-06-20
printed xor_mapped_address

carloleo:
the output has been checked by running `./ndpiReader -i whatsapp_stun_orig_call.pcapng`. 
PCAP file [here](https://github.com/ntop/nDPI/files/1784544/whatsapp_stun_orig_call.zip).

output

### DIFF
--- a/src/lib/protocols/stun.c
+++ b/src/lib/protocols/stun.c
@@ -35,8 +35,6 @@
 
 #define STUN_HDR_LEN   20 /* STUN message header length, Classic-STUN (RFC 3489) and STUN (RFC 8489) both */
 
-void print_xor_mapped_address(const uint8_t* payload,u_int16_t payload_packet_len);
-
 /* ************************************************************ */
 
 u_int32_t get_stun_lru_key(struct ndpi_flow_struct *flow, u_int8_t rev) {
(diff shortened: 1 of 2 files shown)


IvanNardi (changes requested):
@carloleo thank for your code. I agree that extracting this attribute might be useful. However there are two major issues with your PR.

1) Attributes (in this case, xor_mapped_address) are never directly printed by the library. We need to save them in `struct ndpi_flow_struct` and then: a) print them in `ndpiReader` code; b) add them to the json export. You can take a look at any HTTP/TLS metadata for some examples

2) The code parsing the STUN attributes list is already there in `stun.c` around line 323. No need to reinvent it.

carloleo:
Hi @IvanNardi 

thank you for your comment.

I had seen where the attributes are parsed, but I opted for a function because of what I commented at line 337 (`stun.c`).

lucaderi:
@carloleo As @IvanNardi said, we cannot merge your code as it is a duplicate of the existing one, making nDPI unmaintainable. Please update the code as instructed.

src/lib/protocols/stun.c
```diff
@@ -332,8 +334,8 @@ static ndpi_int_stun_t ndpi_int_check_stun(struct ndpi_detection_module_struct *
         case 0x4002:
           /* These are the only messages apparently whatsapp voice can use */
           *app_proto = NDPI_PROTOCOL_WHATSAPP_CALL;
-          return(NDPI_IS_STUN);
-
+          //return(NDPI_IS_STUN); //necessary because the allocate success response packets (from the PCAP file) don't come through TODO: figure out why
+          break;
```
IvanNardi: This is likely an unrelated issue that we can address later: right now we end the dissection when we have a valid sub-proto (whatsapp Call in this case); we might need to keep parsing more pkts/attributes looking for more metadata.

Anyway, for the time being, to test your code simply use a pcap with only one Allocate Response pkt

IvanNardi:
@carloleo
1) could you take a look at the build failure in the CI, please? [some warnings to fix]
2) could you add ipv6 support, too?

carloleo:
@IvanNardi 

I added the attributes to `struct ndpi_flow_struct`, but this violated the constraint on the size.
image

IvanNardi:
> @IvanNardi
> 
> I added the attributes to `struct ndpi_flow_struct`, but this violated the constraint on the size. image

Let's try updating the "952" constant to 960....

IvanNardi:
[6 months later...] STUN code is quite different now... if you are still interested in a similar feature, please open a new PR, please

PR #2018, opened 2023-06-21
STUN: add dissection of DTLS handshake

IvanNardi:


src/lib/protocols/stun.c
```diff
@@ -221,6 +223,44 @@ static ndpi_int_stun_t ndpi_int_check_stun(struct ndpi_detection_module_struct *
     return(NDPI_IS_NOT_STUN);
   }
 
+  /* If we're here it's because this does not look like STUN anymore
+     as this was a flow that started as STUN and turned into something
+     else. Let's investigate what is that about */
+  if(flow->stun.num_pkts > 0 &&
+     payload_length > 12 &&
+     payload[0] == 0x16) {
+    /* Let's check if this is DTLS used by some socials */
```
utoni: Is it not possible to use an already existing function from the TLS dissector to detect DTLS payload?

IvanNardi: Good idea. Done in the last revision

utoni (approved):

PR #2021, opened 2023-06-21
RTP: rework code

IvanNardi:
Try avoiding false positives: look for 3 RTP packets before classifing the flow as such.

Add a generic function `is_rtp_or_rtcp()` to identify RTP/RTCP packets also in other dissectors (see 3608ab01b commit message for an example)


src/lib/protocols/rtp.c
```diff
@@ -165,25 +174,83 @@ static u_int8_t isZoom(struct ndpi_flow_struct *flow,
   return(0);
 }
 
+int is_rtp_or_rtcp(struct ndpi_detection_module_struct *ndpi_struct,
```
utoni, line 183: It makes sense to use an `enum` as retval for a better documentation.

src/lib/protocols/rtp.c
```diff
@@ -165,25 +174,83 @@ static u_int8_t isZoom(struct ndpi_flow_struct *flow,
...
-     || (d_port == 5355 /* LLMNR_PORT */)
-     || (d_port == 5353 /* MDNS_PORT */)
-     ) {
+  if(d_port == 5355 || /* LLMNR_PORT */
+     d_port == 5353    /* MDNS_PORT */) {
     NDPI_EXCLUDE_PROTO(ndpi_struct, flow);
     return;
   }
 
+  /* TODO: should we move zoom stuff in a new, separated dissector? */
```
utoni, line 259: Agree.

src/lib/protocols/rtp.c
IvanNardi: I agree. Done

src/lib/protocols/rtp.c
IvanNardi: I will try something in a follow-up

utoni (approved):

PR #2036, opened 2023-07-07
Add Service Location Protocol dissector.

utoni:


utoni:
I would like to implement version 1 together with the dissection of multiple `<URL Entry>`s.
But without any pcap, this might not be a good idea.

utoni:
Besides the missing documentation, what do you think @IvanNardi ?

IvanNardi:
> I would like to implement version 1 together with the dissection of multiple `<URL Entry>`s. But without any pcap, this might not be a good idea.

I don't know if in this trace there is one or multiple url:  [srv_v1_anon.zip](https://github.com/ntop/nDPI/files/11980810/srv_v1_anon.zip)

utoni:
> > I would like to implement version 1 together with the dissection of multiple `<URL Entry>`s. But without any pcap, this might not be a good idea.
> 
> I don't know if in this trace there is one or multiple url: [srv_v1_anon.zip](https://github.com/ntop/nDPI/files/11980810/srv_v1_anon.zip)

It does not contain multiple URL entries, but at least version 1. That'll help for now. Thanks! :)

src/lib/ndpi_main.c
```diff
@@ -2102,6 +2102,10 @@ static void ndpi_init_protocol_defaults(struct ndpi_detection_module_struct *ndp
 			  "Thrift", NDPI_PROTOCOL_CATEGORY_RPC,
 			  ndpi_build_default_ports(ports_a, 0, 0, 0, 0, 0) /* TCP */,
 			  ndpi_build_default_ports(ports_b, 0, 0, 0, 0, 0) /* UDP */);
+  ndpi_set_proto_defaults(ndpi_str, 1 /* cleartext */, 0 /* nw proto */, NDPI_PROTOCOL_ACCEPTABLE, NDPI_PROTOCOL_SERVICE_LOCATION,
+              "Service Location Protocol", NDPI_PROTOCOL_CATEGORY_RPC,
```
IvanNardi: We usually don't allow spaces in protocol names

src/lib/ndpi_main.c
```diff
@@ -2102,6 +2102,10 @@ static void ndpi_init_protocol_defaults(struct ndpi_detection_module_struct *ndp
...
+              ndpi_build_default_ports(ports_a, 427, 0, 0, 0, 0) /* TCP */,
+              ndpi_build_default_ports(ports_b, 427, 0, 0, 0, 0) /* UDP */);
```
IvanNardi: Could you use the same indentation of the previous protocols?

src/lib/ndpi_main.c
```diff
@@ -2102,6 +2102,10 @@ static void ndpi_init_protocol_defaults(struct ndpi_detection_module_struct *ndp
 			  "Thrift", NDPI_PROTOCOL_CATEGORY_RPC,
 			  ndpi_build_default_ports(ports_a, 0, 0, 0, 0, 0) /* TCP */,
 			  ndpi_build_default_ports(ports_b, 0, 0, 0, 0, 0) /* UDP */);
+  ndpi_set_proto_defaults(ndpi_str, 1 /* cleartext */, 0 /* nw proto */, NDPI_PROTOCOL_ACCEPTABLE, NDPI_PROTOCOL_SERVICE_LOCATION,
+              "Service Location Protocol", NDPI_PROTOCOL_CATEGORY_RPC,
+              ndpi_build_default_ports(ports_a, 427, 0, 0, 0, 0) /* TCP */,
```
IvanNardi: A default port for TCP, but the dissector is registered only on UDP traffic. Is that correct?

IvanNardi (commented):
Missing update in `doc/protocols.rst`

src/lib/ndpi_main.c
utoni: Fixed. According to the RFC both L4 protocols can be used e.g. if the packet is too big to fit in a datagram, TCP is used instead.

utoni:
It will take some more time as I need to forge a PCAP that contains multiple URL entries.

src/lib/protocols/slp.c
```diff
@@ -0,0 +1,286 @@
...
+}
+
+static void ndpi_search_slp(struct ndpi_detection_module_struct *ndpi_struct,
+                            struct ndpi_flow_struct *flow)
+{
+  struct ndpi_packet_struct const * const packet = &ndpi_struct->packet;
+
+  NDPI_LOG_DBG(ndpi_struct, "search Service Location Protocol\n");
+
+  if (packet->payload_packet_len < 1) {
```
IvanNardi: Is this check really necessary? We should always have `packet->payload_packet_len >= 1`, because of `NDPI_SELECTION_BITMASK_PROTOCOL_V4_V6_TCP_OR_UDP_WITH_PAYLOAD_WITHOUT_RETRANSMISSION`

utoni: True. Will remove.

utoni:
Ready for merge soon.

ToDo:

- ~~v1 dissection~~
- ~~URL entry dissection should obey `num_auth` for offset calculation~~

IvanNardi (approved):

PR #2038, opened 2023-07-07
Fixed heap-overflow if compiled with `--enable-tls-sigs`.

utoni:
It may be useful to enable `--enable-tls-sigs` in CiFuzz.


IvanNardi:
Do you want to add `--enable-tls-sigs` to `tests/ossfuzz.sh` right now?

IvanNardi (approved):

PR #2040, opened 2023-07-10
fuzz: extend fuzzing coverage

IvanNardi:
Some notes:
* libinjection: according to https://github.com/libinjection/libinjection/issues/44, it seems NULL characters are valid in the input string;
* RTP: `rtp_get_stream_type()` is called only for RTP packets; if you want to tell RTP from RTCP you should use `is_rtp_or_rtcp()`;
* TLS: unnecessary check; we already make the same check just above, at the beginning of the `while` loop


utoni (approved):
kudos

PR #2041, opened 2023-07-11
Included Gambling website data from the Polish `hazard.mf.gov.pl` list

snicket2100:
The list contains over 30k gambling website hostnames as of today.


snicket2100:
Hmm, it looks like there is no `xmllint` in the build environment. I wanted to avoid using `sed` for parsing the xml file, but I can as well do it if it turns out to be lesser evil...

utoni:
Can you provide any source of the collected gambling sites?

utoni (changes requested):
more information required

snicket2100:
It's a Polish national gambling website blocklist, the API is documented here https://hazard.mf.gov.pl/Content/files/Rejestr_Stron%20Hazardowych_Specyfikacja_WeWy_v1-2.pdf, unfortunately it's in Polish :/ But the list is quite extensive (over 30k entries) and also updated almost on daily basis so I thought it may be helpful.

IvanNardi:
> Hmm, it looks like there is no `xmllint` in the build environment. I wanted to avoid using `sed` for parsing the xml file, but I can as well do it if it turns out to be lesser evil...

@snicket2100 , you can simply add "libxml2-utils" at line 56 of  `.github/workflows/build.yml`

utoni (approved):

IvanNardi (approved):

IvanNardi:
Thanks for the contribution

PR #2050, opened 2023-07-16
tests: restore some old paths as symbolic links

IvanNardi:


utoni (approved):
This may not work on Windows, but running tests on Windows isn't supported anyway. So we're good to go.

PR #2053, opened 2023-07-19
TLS: add basic, basic, detection of Encrypted ClientHello

IvanNardi:


utoni (approved):
Ty for fixing that. I've encountered lot's of sus exts.

PR #2058, opened 2023-07-24
Add an heuristic to detect fully encrypted flows

IvanNardi:
A fully encrypted session is a flow where every bytes of the payload is encrypted in an attempt to “look like nothing”. The heuristic needs only the very first packet of the flow. See: https://www.usenix.org/system/files/sec23fall-prepub-234-wu-mingshi.pdf

A basic, but generic, inplementation of the popcpunt alg has been added


utoni:
Interesting. Did you try to implement Ex4/Ex5?

IvanNardi:
> Interesting. Did you try to implement Ex4/Ex5?

They are already implemented... You can see Ex4 comment just below Ex3.
I implemented a stronger version of Ex5: instead of looking for HTTP/TLS only, we exclude a flow matching ANY nDPI protocols, i.e. this heuristic is triggered only for unknown traffic, which seems the right approach to me (I don't want the risk set for a Whataspp or Telegram flow, I already know what they are...). This is the reason the new risk is set in `ndpi_protocol ndpi_detection_giveup`

PR #2065, opened 2023-07-29
DNS: extract geolocation information, if available

IvanNardi:
The option NSID (RFC5001) is used by Google DNS to report the airport code of the metro where the DNS query is handled.

This option is quite rare, but the added overhead in DNS code is pretty much zero for "normal" DNS traffic


lucaderi:
It looks good to me

utoni (approved):
Same here.

PR #2066, opened 2023-07-31
Add generic functions for configuration

IvanNardi:
TODO

### DIFF
--- a/example/ndpiReader.c
+++ b/example/ndpiReader.c
@@ -967,28 +967,33 @@ static int parse_two_unsigned_integer(char *param, u_int32_t *num1, u_int32_t *n
 
 static int parse_three_strings(char *param, char **s1, char **s2, char **s3)
 {
-  char *saveptr, *tmp_str, *s1_str, *s2_str, *s3_str;
+  char *saveptr, *tmp_str, *s1_str, *s2_str = NULL, *s3_str;
 
   tmp_str = ndpi_strdup(param);
   if(tmp_str) {
-    s1_str = strtok_r(tmp_str, ",", &saveptr);
-    if(s1_str) {
-      s2_str = strtok_r(NULL, ",", &saveptr);
-      if(s2_str) {
-        s3_str = strtok_r(NULL, ",", &saveptr);
-        if(s3_str) {
-          *s1 = ndpi_strdup(s1_str);
-          *s2 = ndpi_strdup(s2_str);
-          *s3 = ndpi_strdup(s3_str);
-          ndpi_free(tmp_str);
-          if(!s1 || !s2 || !s3) {
-            ndpi_free(s1);
-            ndpi_free(s2);
-            ndpi_free(s3);
-            return -1;
-          }
-          return 0;
+    if(param[0] == ',') { /* First parameter might be missing */
+      s1_str = NULL;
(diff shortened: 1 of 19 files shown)


IvanNardi:
This code is only a proof-of-concept to illustrate a new way to provide (more) configuration options in nDPI. The important stuff is at the end of `ndpi_main.c` file: the `ndpi_set_config()` function and the `cfg_params` data structure

The idea is to have a simple way to configure (most of) nDPI:
* only one function to set any configuration parameters (in the present or on in the future)
* try to keep this function prototype as agnostic as possible.

* we don't break API/ABI compatibility of the library; even changing the parameter type (from integer to a list of integer, for example) doesn't break the compatibility

Basically `ndpi_set_config()` is an extended version of `ndpi_set_detection_preferences()` [if you prefer, we can keep this name...]

[There is a LOT of work to do to have some kind of API/ABI compatibility in nDPI (if we ever want to do that...), and the configuration might be a good (and simple) starting point...]

As example of how to extend the configuration, I added the ability to enable/disable the extraction of the sha1 fingerprint of the TLS certificates.

This logic is not new: lots of libraries already uses it

If this idea is accepted, we might remove `ndpi_set_detection_preferences()`, `enum  ndpi_prefs`, `ndpi_set_protocol_aggressiveness()`, `ndpi_set_opportunistic_tls()`....

utoni:
> [There is a LOT of work to do to have some kind of API/ABI compatibility in nDPI (if we ever want to do that...), and the configuration might be a good (and simple) starting point...]

I think there are legitimate reasons to break API and ABI compatibility in some rare case, because there is always be an ongoing development progress and some changes might require to do so.
But as we do not have any API/ABI check in the CI, I think it is not always easy to verify that (I mean in a automated manner).

utoni:
> Basically `ndpi_set_config()` is an extended version of `ndpi_set_detection_preferences()` [if you prefer, we can keep this name...]

But why not e xtending `ndpi_set_detection_preferences()` and break the API/ABI for once?

IvanNardi:
> > [There is a LOT of work to do to have some kind of API/ABI compatibility in nDPI (if we ever want to do that...), and the configuration might be a good (and simple) starting point...]
> 
> I think there are legitimate reasons to break API and ABI compatibility in some rare case, because there is always be an ongoing development progress and some changes might require to do so. But as we do not have any API/ABI check in the CI, I think it is not always easy to verify that (I mean in a automated manner).

Sw compatibility across releases is not the topic itself of this PR: I cited it only because it is a topic where we, historically, never paid attention and where we could likely do better.
If we are able to create a configuration API which, as a side effect, guarantees some kind of API/ABI compatibility, I think it would be nice. I stress out "as side effect" part, it is not the primary goal

IvanNardi:
> > Basically `ndpi_set_config()` is an extended version of `ndpi_set_detection_preferences()` [if you prefer, we can keep this name...]
> 
> But why not e xtending `ndpi_set_detection_preferences()` and break the API/ABI for once?

As I said, the naming is open to discussion: we definitely could keep the old `ndpi_set_detection_preferences` name...

IvanNardi:
A simpler and updated version is available #2190

IvanNardi:
Superseded by #2249

PR #2074, opened 2023-08-20
Improve TFTP detection with options

thomasjwinter:
TFTP read/write requests can have optional strings after the filename and mode such as block size and transfer size to speed the transfer. There is also the Option Acknowledgement opcode that contains the option strings that were accepted.

nDPI fails to recognize tftp transactions using options as tftp.

### DIFF
--- a/src/include/ndpi_typedefs.h
+++ b/src/include/ndpi_typedefs.h
@@ -921,6 +921,10 @@ struct ndpi_flow_udp_struct {
   /* NDPI_PROTOCOL_LINE_CALL */
   u_int8_t line_pkts[2];
   u_int8_t line_base_cnt[2];
+
+  /* NDPI_PROTOCOL_TFTP */
+  u_int16_t tftp_data_num;
+  u_int16_t tftp_ack_num;
 };
 
 /* ************************************************** */
--- a/src/lib/protocols/tftp.c
+++ b/src/lib/protocols/tftp.c
@@ -65,43 +65,69 @@ static void ndpi_search_tftp(struct ndpi_detection_module_struct *ndpi_struct,
         }
 
         {
-          char const * const possible_modes[] = { "netascii", "octet", "mail" };
-          uint8_t mode_found = 0, mode_idx;
+          uint8_t mode_found = 0;
           size_t mode_len;
+          int i;
+          size_t filename_len = 0;
+          const char *string;
+          size_t len = 0;
+          bool first = true;
 
-          for(mode_idx = 0; mode_idx < NDPI_ARRAY_LENGTH(possible_modes); ++mode_idx)
+          /* Skip 2 byte opcode. */
+          for (i = 2; i < packet->payload_packet_len; i++)
           {
-            mode_len = strlen(possible_modes[mode_idx]);
-
-            if (packet->payload_packet_len < mode_len + 1 /* mode is a nul terminated string */)
+            /* Search through the payload until we find a NULL terminated string. */
+            if (packet->payload[i] != '\0')
             {
+              len++;
               continue;
             }
-            if (strncasecmp((char const *)&packet->payload[packet->payload_packet_len - 1 - mode_len],
-                            possible_modes[mode_idx], mode_len) == 0)
+            string = (const char *)&packet->payload[i - len];
+
+            /* Filename should be immediately after opcode followed by the mode. */
+            if (first)
             {
-              mode_found = 1;
-              break;
+              filename_len = len;
+              len = 0;
+              first = false;
+              continue;
             }
+
+            char const * const possible_modes[] = { "netascii", "octet", "mail" };
+            uint8_t mode_idx;
+
+            /* Check the string in the payload against the possible TFTP modes. */
+            for(mode_idx = 0; mode_idx < NDPI_ARRAY_LENGTH(possible_modes); ++mode_idx)
+              {
+                mode_len = strlen(possible_modes[mode_idx]);
+                /* Both are now null terminated */
+                if (len != mode_len)
+                {
+                  continue;
+                }
+                if (strncasecmp(string, possible_modes[mode_idx], mode_len) == 0)
+                {
+                  mode_found = 1;
+                  break;
+                }
+              }
+
+            /* Second string searched must've been the mode, break out before any following options. */
+            break;
           }
 
-          if (mode_found == 0)
+          /* Exclude the flow as TFPT if there was no filename and mode in the first two strings. */
+          if (filename_len == 0 || ndpi_is_printable_buffer(&packet->payload[2], filename_len) == 0 ||
+              mode_found == 0)
           {
             NDPI_EXCLUDE_PROTO(ndpi_struct, flow);
             return;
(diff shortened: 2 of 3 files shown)


src/lib/protocols/tftp.c
```diff
@@ -65,24 +65,55 @@ static void ndpi_search_tftp(struct ndpi_detection_module_struct *ndpi_struct,
...
         {
-          char const * const possible_modes[] = { "netascii", "octet", "mail" };
-          uint8_t mode_found = 0, mode_idx;
+          uint8_t mode_found = 0;
           size_t mode_len;
-
-          for(mode_idx = 0; mode_idx < NDPI_ARRAY_LENGTH(possible_modes); ++mode_idx)
+          int i;
+          size_t filename_len = 0;
+          char *string;
```
utoni:
```suggestion
          const char *string;
```

utoni:
Please take a look at the CI: https://github.com/ntop/nDPI/actions/runs/5933443642/job/16095465366?pr=2074

Seems we are loosing one TFTP flow detection.

thomasjwinter:
It looks like the tftp.pcap was changed in commit 1a01e8dc6 where it added two invalid read requests that don't abide by the rfc. You can open it in wireshark and see that it interprets them strangely. The second read request has 11 zero bytes following the opcode then the mode which is interpreted as an option. I believe this should have been interpreted as a malformed packet with zero length filename so this patch should be more correct.

utoni:
Ok, please update the test results with `NDPI_FORCE_UPDATING_UTESTS_RESULTS=1 NDPI_DISABLE_FUZZY=1 ./tests/do.sh` and upload the changes.

thomasjwinter:
Rerunning the tests and inspecting closer I see the cause.
My new code was excluding the flow as no mode was found due the new NULL terminated string checking. Previous code was only looking at the end of the payload to get the mode then doing the filename risk check.

Attempting to move the filename risk check above the mode_found exclude check results in the stun_classic.pcap failing as it happens to have the 0x0001 opcode bytes in the right place but the following bytes aren't printable characters.

What do you suggest I should do?
Could remove the malformed packet risk and exclude tftp if the filename or mode is not found.
Should we be excluding if not standard TFTP UDP port 69?

utoni:
I think you can safely remove the malformed packet risk. I remember setting this but the TFTP detection patterns are rather sparse. So I guess it is better to have a "stable" detection instead of a risk set.

IvanNardi:
> I think you can safely remove the malformed packet risk. I remember setting this but the TFTP detection patterns are rather sparse. So I guess it is better to have a "stable" detection instead of a risk set.

+1

tests/cfgs/default/result/tftp.pcap.out
```diff
@@ -25,7 +26,7 @@ TFTP	107	31296	7
...
-	4	UDP 172.28.4.53:54626 -> 172.16.5.170:69 [proto: 96/TFTP][IP: 0/Unknown][ClearText][Confidence: DPI][DPI packets: 1][cat: DataTransfer/4][1 pkts/61 bytes -> 0 pkts/0 bytes][Goodput ratio: 31/0][< 1 sec][Risk: ** Malformed Packet **** Unidirectional Traffic **][Risk Score: 20][Risk Info: No server to client traffic / Invalid TFTP RR/WR header: Source/Destination file missing][Plen Bins: 100,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0]
+	4	UDP 172.28.4.53:54626 -> 172.16.5.170:69 [proto: 96/TFTP][IP: 0/Unknown][ClearText][Confidence: Match by port][DPI packets: 1][cat: DataTransfer/4][1 pkts/61 bytes -> 0 pkts/0 bytes][Goodput ratio: 31/0][< 1 sec][Risk: ** Unidirectional Traffic **][Risk Score: 10][Risk Info: No server to client traffic][Plen Bins: 100,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0]
```
utoni, line 29: Shouldn't that be `[Confidence: DPI]` ?

tests/cfgs/default/result/tftp.pcap.out
```diff
@@ -25,7 +26,7 @@ TFTP	107	31296	7
...
-	7	UDP 172.28.4.53:54632 -> 172.16.5.170:69 [proto: 96/TFTP][IP: 0/Unknown][ClearText][Confidence: DPI][DPI packets: 1][cat: DataTransfer/4][1 pkts/51 bytes -> 0 pkts/0 bytes][Goodput ratio: 17/0][< 1 sec][Risk: ** Malformed Packet **** Unidirectional Traffic **][Risk Score: 20][Risk Info: No server to client traffic / Invalid TFTP RR/WR header: Source/Destination file missing][Plen Bins: 100,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0]
+	7	UDP 172.28.4.53:54632 -> 172.16.5.170:69 [proto: 96/TFTP][IP: 0/Unknown][ClearText][Confidence: Match by port][DPI packets: 1][cat: DataTransfer/4][1 pkts/51 bytes -> 0 pkts/0 bytes][Goodput ratio: 17/0][< 1 sec][Risk: ** Unidirectional Traffic **][Risk Score: 10][Risk Info: No server to client traffic][Plen Bins: 100,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0]
```
utoni, line 32: same here

thomasjwinter:
As discussed previously, those are the 2 packets with incorrect formats so are now being excluded.

src/include/ndpi_typedefs.h
```diff
@@ -1551,6 +1551,8 @@ struct ndpi_flow_struct {
 
     struct {
       char filename[128];
+      u_int16_t data_num;
+      u_int16_t ack_num;
```
IvanNardi: This is wrong. These two variables need to be defined in `struct ndpi_flow_udp_struct` (like `tftp_stage`).
Reason: since `ndpi_flow_struct->protos` is an union, we can't write to it until we have a final classification, i.e. until we call `ndpi_int_tftp_add_connection()`.

Could you rebase and update the code, please?

IvanNardi (changes requested):
Thanks for the contribution. Could you try to fix the code, as reported below, please?

IvanNardi (approved):

IvanNardi:
Thanks for the fixes!

PR #2076, opened 2023-08-21
dns: make subprotocol inspection optional

thomasjwinter:
Normally DNS will inspect the hostname to try to match a subprotocol. It may be desireable to have all DNS packets match as DNS instead.

Keep the existing behaviour default so that a new configure option is needed to turn off DNS subprotocol inspection.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -84,6 +84,16 @@ AS_IF([test "x${enable_code_coverage}" = "xyes"],[
 
 AS_IF([test "${enable_npcap+set}" = set],[DISABLE_NPCAP=1],[DISABLE_NPCAP=0])
 
+AC_MSG_CHECKING([whether to do DNS subprotocol inspection])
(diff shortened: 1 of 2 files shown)


utoni:
Shouldn't be the `master_protocol` always set to DNS, whereas the application protocol is set to the application if the hostname matched?

IvanNardi:
I like the idea, but it should be a runtime configuration knob (via a new `ndpi_prefs`) and not a compilation option.
Could you also add a new test configuration, please? You could take a look at `tests/cfgs/disable_stun_monitoring/` (or any others...) for example and use a simple dns trace.

thomasjwinter:
> Shouldn't be the `master_protocol` always set to DNS, whereas the application protocol is set to the application if the hostname matched?

This is correct. I've now changed to use master_protocol instead of app_protocol for the DNS protocols. Thanks.

Although maybe this might still be wanted to save processing on the subprotocol inspection when you don't need it?

PR #2084, opened 2023-09-12
Add `ndpi_domain_classify_finalize()` function

IvanNardi:
The "domain classify" data structure is immutable, because it uses "bitmap64".
Allow to finalize it before starting to process packets (i.e. before calling `ndpi_domain_classify_contains()`) to avoid, in the data-path, all the memory allocations due to compression.

Calling `ndpi_domain_classify_finalize()` is optional.


lucaderi:
Thank you: it is a good addition to the existing API

PR #2088, opened 2023-09-29
Added HAProxy protocol.

utoni:
Fixed wrong ` tests/do.sh.in` "failed pcaps" logging.

IMHO most of the HAProxy payload is HTTP / TLS so it would make sense to use the subprotocol semantics here. However, the TLS dissector is not ready to get called as subprotocol dissector unfortunately.


IvanNardi (commented):
Could you add an entry to `doc/protocols.rst`, please?

IvanNardi (approved):

PR #2094, opened 2023-10-03
QUIC: export QUIC version as metadata

IvanNardi:


utoni (approved):
A valuable addition to other projects that use the JSON serializer! :)

PR #2106, opened 2023-10-18
Add Remote Management Control Protocol (RMCP).

utoni:
If the detection patterns are not sufficient i.e. cause false-positives, we may also dissect `IPMI` / `ASF`.


utoni:
@lucaderi @IvanNardi CLA signed

IvanNardi:
> @lucaderi @IvanNardi CLA signed

Me too

PR #2113, opened 2023-10-25
ipv6: add support for ipv6 addresses lists

IvanNardi:


example/ndpiReader.c
```diff
@@ -3901,6 +3901,9 @@ static void printResults(u_int64_t processing_time_usec, u_int64_t setup_time_us
       printf("\tPatricia protocols:   %llu/%llu (search/found)\n",
 	     (long long unsigned int)cumulative_stats.patricia_stats[NDPI_PTREE_PROTOCOLS].n_search,
 	     (long long unsigned int)cumulative_stats.patricia_stats[NDPI_PTREE_PROTOCOLS].n_found);
+      printf("\tPatricia protocols 6: %llu/%llu (search/found)\n",
```
utoni: What about something like:

```suggestion
      printf("\tPatricia protocols IPv6: %llu/%llu (search/found)\n",
```

src/lib/ndpi_main.c
```diff
@@ -2475,6 +2479,52 @@ u_int16_t ndpi_network_port_ptree_match(struct ndpi_detection_module_struct *ndp
...
+					 struct in6_addr *pin,
+					 u_int16_t port /* network byte order */)
+{
+  ndpi_prefix_t prefix;
+  ndpi_patricia_node_t *node;
+
+  if(!ndpi_str || !ndpi_str->protocols_ptree6)
+    return(NDPI_PROTOCOL_UNKNOWN);
+
+  /* TODO: check on "private" addresses? */
```
utoni, line 2492: Makes sense for me. There could be also risk set, if that happens.

IvanNardi: More than an anomaly, it is an optimization. Like in the ipv4 case, if there are not link local/"private" addresses in the tree (and, by default, there are not) we can avoid the lookup if the flow addressees are of these types

example/ndpiReader.c

utoni (approved):

PR #2121, opened 2023-10-27
Add malicious sites from polish cert.

utoni:
* care about parsing errors for `*.list`


example/ndpiReader.c
```diff
@@ -2680,8 +2680,16 @@ static void setupDetection(u_int16_t thread_id, pcap_t * pcap_handle) {
       exit(-1);
   }
 
-  if(_categoriesDirPath)
-    ndpi_load_categories_dir(ndpi_thread_info[thread_id].workflow->ndpi_struct, _categoriesDirPath);
+  /* Disabled while fuzzing to speed-up init. */
+  if(_categoriesDirPath) {
+#ifndef FUZZING_BUILD_MODE_UNSAFE_FOR_PRODUCTION
```
IvanNardi: I understand the reason, but I am not happy about it: we might end up having unit tests with different results, depending on fuzzing being enable or not...
Not sure about other alternatives, though... :(

At very least, I would add something like that (completly untested!):
```
--- a/tests/ossfuzz.sh
+++ b/tests/ossfuzz.sh
@@ -64,4 +64,5 @@ cp fuzz/bd_param.txt $OUT/
 cp fuzz/splt_param.txt $OUT/
 cp fuzz/random_list.list $OUT/
 mkdir -p $OUT/lists
-cp lists/*.list $OUT/lists
+#Ignore a huge list to speed up init time
+find lists/*.list ! -name 100_malware.list -exec cp -t $OUT/lists/ {} +
```
 to avoid slowing down oss-fuzz (we call directly `ndpi_load_categories_dir` from some fuzzers)

IvanNardi (commented):
While the PR is fine per se (and the error handling is great) I have a "philosophical" doubt:
do we want to start adding lists with ~150K entries to the library?
I know that these lists are somehow optional and any users might skip them...
I had similar doubts with the gambling list, when it increased from the original some K entries to 30k entries...

Just some thoughts... @lucaderi, what do you think?

IvanNardi:
> While the PR is fine per se (and the error handling is great) I have a "philosophical" doubt: do we want to start adding lists with ~150K entries to the library? I know that these lists are somehow optional and any users might skip them... I had similar doubts with the gambling list, when it increased from the original some K entries to 30k entries...

After some thoughts, I am fine with the list.
I am still not sure about how to handle the performance issue, though

utoni:
A "manual-lists-download" script would solve checking-in large files to git (and reduce init-time if no list was downloaded), but then the user needs to download those and needs to know about such script.

IvanNardi:
Ok, let keep it as you proposed... Could you please change `tests/ossfuzz.sh` and add a simple trace with a session matching one of these domains, please?

IvanNardi (approved):
Thanks!

PR #2129, opened 2023-10-30
Remove `struct ndpi_packet_struct` from the public API

IvanNardi:
Right now, the only instance of `struct ndpi_packet_struct` is embedded into `struct ndpi_detection_module_struct`. Since the latter is a private structure (because of `NDPI_LIB_COMPILATION` ) there is no way for the application to get a pointer to `ndpi_struct->packet`.

Bottom line: the application can't use any API functions having `struct ndpi_packet_struct *` as parameter. Remove them all (since they are completly unused and unusable).

There are no public helper functions to initialize/populate/deinit a `struct ndpi_packet_struct` object, so the application can't neither create its own instance of this object.

Protect `struct ndpi_packet_struct` via the same define `NDPI_LIB_COMPILATION`.

### DIFF
--- a/src/include/ndpi_main.h
+++ b/src/include/ndpi_main.h
@@ -107,11 +107,6 @@ extern "C" {
   extern void ndpi_set_proto_subprotocols(struct ndpi_detection_module_struct *ndpi_mod,
 				      int protoId, ...);
 
-  extern int ndpi_packet_src_ip_eql(const struct ndpi_packet_struct *packet, const ndpi_ip_addr_t * ip);
-  extern int ndpi_packet_dst_ip_eql(const struct ndpi_packet_struct *packet, const ndpi_ip_addr_t * ip);
-  extern void ndpi_packet_src_ip_get(const struct ndpi_packet_struct *packet, ndpi_ip_addr_t * ip);
-  extern void ndpi_packet_dst_ip_get(const struct ndpi_packet_struct *packet, ndpi_ip_addr_t * ip);
-
   extern int ndpi_parse_ip_string(const char *ip_str, ndpi_ip_addr_t *parsed_ip);
   extern char *ndpi_get_ip_string(const ndpi_ip_addr_t * ip, char *buf, u_int buf_len);
   extern u_int8_t ndpi_is_ipv6(const ndpi_ip_addr_t *ip);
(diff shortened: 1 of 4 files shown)


utoni:
I think this change is fine for now. But if we ever want to support other developers writing custom dissectors e.g. as part of their own `libnDPI` integration, we might need to re-add that struct.

Another issue is that our "private" structs aren't real private structs, because they're residing inside a public header file dependent on a `#ifdef` guard. A solution I would prefer is creating a private header file below `src/lib` and using opaque pointers in public header files.

src/include/ndpi_main.h
```diff
@@ -107,11 +107,6 @@ extern "C" {
   extern void ndpi_set_proto_subprotocols(struct ndpi_detection_module_struct *ndpi_mod,
 				      int protoId, ...);
 
-  extern int ndpi_packet_src_ip_eql(const struct ndpi_packet_struct *packet, const ndpi_ip_addr_t * ip);
```
utoni, line 110: Those functions could still be useful.
What about:

```C
extern int ndpi_packet_src_ip_eql(const ndpi_ip_addr_t * a, const ndpi_ip_addr_t * b);
```

IvanNardi: I think the entire `ndpi_ip_addr_t` is completely bogus...  because we are not able to tell ipv4 from ipv6 only from such object, if we use prefixes and not addresses only
The function
```
u_int8_t ndpi_is_ipv6(const ndpi_ip_addr_t *ip) {
  return(ip->ipv6.u6_addr.u6_addr32[1] != 0 || ip->ipv6.u6_addr.u6_addr32[2] != 0 ||
	 ip->ipv6.u6_addr.u6_addr32[3] != 0);
}
```
is wrong in these conditions. (trivial example: fec0::/8 will be classified as ipv4...)
The only usage of `ndpi_ip_addr_t` in the library is via `ndpi_ptree_insert()` with does handles prefixes so it is potentially bogus too. Fortunately:
1) `ndpi_ptree_insert()` is never used by the library
2) it is used by ntopng, but with full address, i.e. without any issues
```
ivan@ivan-Latitude-E6540:~/svnrepos/ntopng(dev)$ git grep ndpi_ptree_insert
src/flow_checks/UnexpectedServer.cpp:            ndpi_ptree_insert(whitelist_ptree, &a, 32, naddr);
src/flow_checks/UnexpectedServer.cpp:            ndpi_ptree_insert(whitelist_ptree, &a, 128, naddr);

```

Bottom line: I think we can avoid adding (back) something like that right now
```
extern int ndpi_packet_src_ip_eql(const ndpi_ip_addr_t * a, const ndpi_ip_addr_t * b);
```
and in the future we need to rework all the `ndpi_ip_addr_t` stuff (remove it completely? Handle only addresses and not prefixes?)

What do you think?

IvanNardi:
> Another issue is that our "private" structs aren't real private structs, because they're residing inside a public header file dependent on a `#ifdef` guard. A solution I would prefer is creating a private header file below `src/lib` and using opaque pointers in public header files.

That is exactly the long term goal!!
[and maybe, maybe some ```-fvisibility=hidden``` and ```__attribute__((visibility("default")))``` magic; but definitely different headers for public and private API]

src/include/ndpi_main.h
utoni: In that case handling only addresses seems like a cleaner approach to me.
Thanks for clarification!

utoni (approved):

PR #2135, opened 2023-11-02
fuzz: improve coverage and remove dead code

IvanNardi:
We are not able to remove custom rules: remove the empty stubs (which originate from the original OpenDPI code).

`ndpi_guess_protocol_id()` is only called on the first packet of the flow, so the bitmask `flow->excluded_protocol_bitmask` is always empty, since we didn't call any dissectors yet.

Move another hash function to the dedicated source file.


utoni (approved):
Not sure what the idea of `removeDefaultPort` was, but it does not sound good.

PR #2137, opened 2023-11-07
Have a clear distinction between public and private/internal API

IvanNardi:
1) Public API/headers in `src/include/` [as it has always been]
2) Private API/headers in `src/lib/`
    
Try to keep the "ndpi_" prefix only for the public functions


IvanNardi:
For the time being, there is only one private header, `ndpi_private.h` with both some typedefs/structures and some function declarations. If we need/want, we might have multiple headers...

The most important private structures are `struct ndpi_packet_struct` and `struct ndpi_detection_module_struct`, which are already "private" (via `NDPI_LIB_COMPILATION`)

windows/nDPI.vcxproj
```diff
@@ -91,7 +91,7 @@
   <ItemDefinitionGroup Condition="'$(Configuration)|$(Platform)'=='Debug-ndpiReader|x64'">
     <ClCompile>
       <Optimization>Disabled</Optimization>
-      <AdditionalIncludeDirectories>$(ProjectDir)src\;$(ProjectDir)src\lib\;$(ProjectDir)..\src\lib\protocols\;$(ProjectDir)..\src\include\;$(ProjectDir)..\;$(ProjectDir)..\src\lib\third_party\include\;$(ProjectDir)..\windows\WpdPack\Include\;%(AdditionalIncludeDirectories)</AdditionalIncludeDirectories>
+      <AdditionalIncludeDirectories>$(ProjectDir)src\;$(ProjectDir)src\lib\;$(ProjectDir)..\src\lib\;$(ProjectDir)..\src\lib\protocols\;$(ProjectDir)..\src\include\;$(ProjectDir)..\;$(ProjectDir)..\src\lib\third_party\include\;$(ProjectDir)..\windows\WpdPack\Include\;%(AdditionalIncludeDirectories)</AdditionalIncludeDirectories>
```
utoni: Am I right that `$(ProjectDir)..\src\lib\protocols\` is only there because of `src/lib/protocols/btlib.h` ?

If so, is it possible to get rid of that header file?

IvanNardi: It is quite likely... I am going to remove it....

IvanNardi:
We could/should move a lot more stuff into `ndpi_private.h`, especially from `src/include/ndpi_main.h`....

utoni (approved):
Good job and a good starting point for further "privatization" of structs and functions. :)

PR #2138, opened 2023-11-07
Fixed invalid unidirectional risk set even if packets from both directions were processed.

utoni:
I've observed a strange behavior in `nDPId` after updating to the latest `nDPI` commit.
Literally **all** UDP packets had the `Unidirectional Traffic` risk set.

The issue was that if the flow was detected e.g. after the first packet, the risk is never unset even if there were packets coming from both directions.
It is necessary to call `ndpi_connection_tracking()` if the flow was detected after the first packet.

### DIFF
--- a/src/lib/ndpi_main.c
+++ b/src/lib/ndpi_main.c
@@ -7781,6 +7781,8 @@ static ndpi_protocol ndpi_internal_detection_process_packet(struct ndpi_detectio
     if(ndpi_init_packet(ndpi_str, flow, current_time_ms, packet_data, packetlen, input_info) != 0)
       return(ret);
 
+    ndpi_connection_tracking(ndpi_str, flow);
+
     goto ret_protocols;
   }
 


src/lib/ndpi_main.c
```diff
@@ -7781,6 +7781,8 @@ static ndpi_protocol ndpi_internal_detection_process_packet(struct ndpi_detectio
     if(ndpi_init_packet(ndpi_str, flow, current_time_ms, packet_data, packetlen, input_info) != 0)
       return(ret);
 
+    ndpi_connection_tracking(ndpi_str, flow);
+
```
IvanNardi, line 7785: I am a little bit confused... I am quite confident that this if branch (``` if(flow->detected_protocol_stack[0] != NDPI_PROTOCOL_UNKNOWN)```)is NEVER triggered; if we have a classification, we triggered the extra dissection data path early....
Did you see any differences with this patch?
Could you share some traces:
* triggering the original issue
* triggering this specific code path, please?

utoni: [uni.pcap.gz](https://github.com/ntop/nDPI/files/13284664/uni.pcap.gz)

triggers that issue (at least with `nDPId`)

utoni: Was there a change regarding this Risk after 1f693c3f5a5dcd9d69dffb610b9a81bd33f95382 ?

utoni: With this patch, everything works as expected.

IvanNardi: > [uni.pcap.gz](https://github.com/ntop/nDPI/files/13284664/uni.pcap.gz)
> 
> triggers that issue (at least with `nDPId`)

With this trace `ndpi_detection_process_packet()` should be called only once (with the first packet) by the application (and `ndpiReader` calls it once). Is `nDPId` calling it more? Why?

[I am puzzled that this PR has any impact at all, because that code should never be triggered; now I am going to look at the risk issue itself...]

IvanNardi: `NDPI_UNIDIRECTIONAL_TRAFFIC` is quite "unusual", because it might be the job of the application to unset it, if necessary. This is quite clear from a comment in `src/include/ndpi_typedefs.h`:
```
  NDPI_UNIDIRECTIONAL_TRAFFIC, /* NOTE: as nDPI can detect a protocol with one packet, make sure
				  your app will clear this risk if future packets (not sent to nDPI)
				  are received in the opposite direction */
```
With `ndpiReader` and your trace:
*) first packet: ```ndpi_detection_process_packet()``` is called, the risk is set, and we have a classification
*) following packets: nDPI is no more called (there is a classification and no extra dissection) but the application itself unsets the risk

```
ivan@ivan-Latitude-E6540:~/svnrepos/nDPI(dev)$ git grep UNIDIRECTIONAL -- example/
example/reader_util.c:      flow->risk &= ~(1ULL << NDPI_UNIDIRECTIONAL_TRAFFIC); /* Clear bit */
```

So, i think that should be `nDPId` to remove the risk in that scenario. However, i have no idea why you bump into this issue only now...

utoni: Ok, I did not know about that comment regarding unsettling that specific risk.

But my doubt is still there. Does it really make sense to set this risk after the first packet lead to a successful classification?

IMHO nDPI just assumes that there won't be any more packets coming from the other direction. This looks like an intentional false positive risk to me. :D

utoni:
I fixed it within `nDPId`, not optimal but it works. Closing this PR.

PR #2145, opened 2023-11-12
protocols/ethereum: enhance node discovery protocol detection.

mmaatuq:

Describe changes:
As per [Node discovery protocol v4](https://github.com/ethereum/devp2p/blob/master/discv4.md)
*  packet-type == 0x0 is not among the known types.
* allow more packet types e.g.  ENRRequest, ENRResponse.

### DIFF
--- a/src/lib/protocols/ethereum.c
+++ b/src/lib/protocols/ethereum.c
@@ -25,6 +25,14 @@
 #include "ndpi_api.h"
 #include "ndpi_private.h"
 
+enum ether_disc_packet_type {
+  DISC_PING = 0x01,
+  DISC_PONG = 0x02,
+  DISC_FINDNODE = 0x03,
+  DISC_NEIGHBOURS = 0x04,
+  DISC_ENRREQUEST = 0x05,
+  DISC_ENRRESPONSE = 0x06
+};
 
 /* ************************************************************************** */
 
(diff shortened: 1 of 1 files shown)


src/lib/protocols/ethereum.c
```diff
@@ -50,34 +58,50 @@ static void ndpi_ether_cache_connection(struct ndpi_detection_module_struct *ndp
...
-    if((packet->iph) && ((ntohl(packet->iph->daddr) & 0xFF000000 /* 255.0.0.0 */) == 0xFF000000))
-      ;
-    else if(packet->iphv6 && ntohl(packet->iphv6->ip6_dst.u6_addr.u6_addr32[0]) == 0xFF020000)
-      ;
-    else {
+  if((packet->iph) && ((ntohl(packet->iph->daddr) & 0xFF000000 /* 255.0.0.0 */) == 0xFF000000))
+    ;
+  else if(packet->iphv6 && ntohl(packet->iphv6->ip6_dst.u6_addr.u6_addr32[0]) == 0xFF020000 /*ff02::*/)
+    ;
+  else {
```
utoni: I did not see this in the previous code review, but I think it might improve readability using something like:
```C
if ((packet->iph && ntohl(packet->iph->daddr) & 0xFF000000 /* 255.0.0.0 */) != 0xFF000000)
    || (packet->iphv6 && ntohl(packet->iphv6->ip6_dst.u6_addr.u6_addr32[0]) != 0xFF020000 /* ff02:: */))
{
 // ...
}
```

src/lib/protocols/ethereum.c
```diff
@@ -50,34 +58,47 @@ static void ndpi_ether_cache_connection(struct ndpi_detection_module_struct *ndp
...
-     && (packet->payload_packet_len < 1280)
-     && ((source == 30303) || (dest == 30303))
-     && (packet->payload[97] <= 0x04 /* NODES */)
-     ) {
-    if((packet->iph) && ((ntohl(packet->iph->daddr) & 0xFF000000 /* 255.0.0.0 */) == 0xFF000000))
-      ;
-    else if(packet->iphv6 && ntohl(packet->iphv6->ip6_dst.u6_addr.u6_addr32[0]) == 0xFF020000)
-      ;
-    else {
+  if((packet->iph && ((ntohl(packet->iph->daddr) & 0xFF000000) != 0xFF000000 /* 255.0.0.0 */))
```
IvanNardi, line 94: I know that this is old code, but does anyone know why we skip the netmask 255.0.0.0/8? [the ipv6 part makes sense]

mmaatuq: - not sure about this, but I think  ipv4s start with 255. is kind of a special purpose for broadcast.

IvanNardi (approved):

src/lib/protocols/ethereum.c
mmaatuq: - as per [rfc-5735](https://datatracker.ietf.org/doc/html/rfc5735) range 240.0.0.0/4 is Reserved for Future Use, 
which includes 255.0.0.0/8, maybe it's more convenient to skip the entire range. 
-  do we need a new selection bitmaks e.g not_broadast, not_multicast

IvanNardi: ok, we might address that on a separate patch

PR #2153, opened 2023-11-17
STUN: improve demultiplexing of DTLS packets

IvanNardi:
Keep demultiplexing STUN/RTP/RTCP packets after DTLS ones.

We might end up processing the session a little longer, because we will process the STUN/RTP/RTCP packets after the DTLS handshake.


IvanNardi:
@utoni, could you take a look at the automake stuff, please?

utoni:
naturalmente :)

configure.ac
```diff
@@ -3,7 +3,7 @@ AC_INIT([libndpi],[4.9.0])
 AC_CONFIG_AUX_DIR([.])
 AC_CONFIG_MACRO_DIR([m4])
 
-AM_INIT_AUTOMAKE([foreign subdir-objects])
+AM_INIT_AUTOMAKE([foreign subdir-objects 1.9 tar-pax])
```
utoni: Is there a certain reason to use `tar-pax` ?

According to https://www.gnu.org/software/automake/manual/html_node/List-of-Automake-options.html:

> tar-pax selects the new pax interchange format defined by POSIX 1003.1-2001. It does not limit the length of file names. However, this format is very young and should probably be restricted to packages that target only very modern platforms. As of 2018, this format is supported by the native tar command only on GNU, FreeBSD, and OpenBSD systems; it is not supported by the native tar command on NetBSD, AIX, HP-UX, or Solaris. There are moves to change the pax format in an upward-compatible way, so this option may refer to a more recent version in the future.

utoni: Ok, got it. Maybe worth to check if the mentioned platform support `tar-pax` by now.

IvanNardi: Ok. Let be extra cautious here... I reverted the automake change and shortened some filenames... it is the safer solution

utoni (approved):

PR #2157, opened 2023-11-18
Keep separating public and private API

IvanNardi:
See: b08c787fe


src/include/ndpi_api.h
```diff
@@ -2164,6 +2152,29 @@ extern "C" {
    */
   void *ndpi_get_user_data(struct ndpi_detection_module_struct *ndpi_str);
 
+  /* ******************************* */
+
+  /* Can't call libc functions from kernel space, define some stub instead */
```
utoni, line 2157: Just out of curiosity: Running nDPI in kernel space is officially supported?

utoni (approved):
Good job!

src/include/ndpi_api.h
IvanNardi: That line has been simply copied around...
I don't think that nDPI is "officially" supported in kernel-space, but the original OpenDPI was.
And, more important, there is at least one active fork https://github.com/vel21ripn/nDPI (by @vel21ripn) which, AFAIK, works in kernel-space. Since that fork is maintained and quite in sync with upstream, I think that there are no huge issues in upstream code preventing nDPI from working in kernel-mode. But I never investigate it...

utoni: Maybe we could add a simple kernel module to `examples/` and the CI so we know at least if nDPI is able to build in such environments.

PR #2159, opened 2023-11-21
Get rid of Apache Cassandra false positives

0xA50C1A1:

Describe changes:

While writing and testing my HART-IP and some other industrial protocols dissectors, I encountered these protocols being falsely detected as Apache Cassandra. When I started to figure out wtf was going on, I discovered that the Apache Cassandra dissector checks only the first 9 bytes of the payload... apparently, such a check was clearly not enough, cuz I also noticed there were kludges for the H.323 and SOCKS protocols.

I took the liberty of rewriting this dissector from scratch to definitely get rid of all false positives. Everything works fine, but there is one drawback - now the Apache Cassandra protocol is detected only by the 'STARTUP' message, which the client always sends once after the start of the session, so there may be a problems with midstream flows.


0xA50C1A1:
Added a payload size check and now the fuzzing test is successful.  Only build test utils fails at "(2) Downloading Gambling Sites (Poland)" now, but I can’t do anything about it.

0xA50C1A1:
@IvanNardi can you review this, please?

src/lib/protocols/cassandra.c
```diff
@@ -18,125 +21,69 @@
...
+       !ndpi_validate_cassandra_request(packet->payload[0])))
+  {
+    NDPI_EXCLUDE_PROTO(ndpi_struct, flow);
+    return;
+  }
+  
+  if (flow->packet_counter > 12) {
+    NDPI_EXCLUDE_PROTO(ndpi_struct, flow);
+    return;
+  }
```
IvanNardi, line 72: ```4.1.1. STARTUP: This must be the first message of the connection, except for OPTIONS that can be sent before to find out the options supported by the server.```

Since in the worst case, we are looking for the second packets sent by the client I think that we might be able to have a better/faster check. Something like:
```if(flow->packet_direction_counter[packet->packet_direction] > 2 { /*exclude */```

IvanNardi (changes requested):
With this specific protocol, I am fine if we are not able to detect mid-flows anymore (it is quite common for TCP protocols)


Please restore the QUIC trace that you deleted

0xA50C1A1:
> Please restore the QUIC trace that you deleted

Damn, I did it by accident.

src/lib/protocols/cassandra.c
0xA50C1A1: Thanks for the review, it really works better with this check.

utoni:
> With this specific protocol, I am fine if we are not able to detect mid-flows anymore (it is quite common for TCP protocols)

Totally agree. Better to get rid of false-positives then lacking midstream detection.

IvanNardi (approved):

IvanNardi:
@0xA50C1A1, could you rebase, please?

0xA50C1A1:
> @0xA50C1A1, could you rebase, please?

Yea, sure.

IvanNardi:
Thank you

PR #2164, opened 2023-11-23
Reduce H.323 over TCP false positives

0xA50C1A1:

Describe changes:

I made this check more stringent: since H.323 uses the Q.931 protocol, it makes sense to check for a Q.931 Protocol Discriminator (0x08) and call reference value length (maximum 3 bytes by default). Well, it would be good to add message type and IE type checks, but I guess that would be a little bit redundant.


0xA50C1A1:
@IvanNardi Could you please review this?

IvanNardi (approved):
Next time, could you avoid touching the indentation of unrelated code, please? It would simplify review...
Anyway, useful work!

IvanNardi:
Thanks!

PR #2190, opened 2023-12-01
Add generic functions for library configuration [advanced version]

IvanNardi:
This is a new way to provide (more) configuration options in nDPI.

The idea is to have a simple way to configure (most of) nDPI: only one function (`ndpi_set_config()`) to set any configuration parameters (in the present or on in the future) and this function prototype is as agnostic as possible.

 * we don't break API/ABI compatibility of the library; even changing the parameter type (from integer to a list of integer, for example) doesn't break the compatibility.

Two examples of how to extend the configuration are provided:
 * the ability to enable/disable the extraction of the sha1 fingerprint of the TLS certificates.
 * the upper limit on the number of packets per flow that will be subject to inspection


IvanNardi:
A simpler and updated version of #2066. See previous discussion there

IvanNardi:
**API changes**

Almost all nDPI configuration is made via `ndpi_set_config()` and lots of configuration functions have been removed.
Important examples of API changes:

* Enumeration `ndpi_init_prefs` has been removed: `struct ndpi_detection_module_struct *ndpi_init_detection_module(ndpi_init_prefs prefs)` --> `struct ndpi_detection_module_struct *ndpi_init_detection_module(void)`. \
Example: `ndpi_s = ndpi_init_detection_module(ndpi_track_flow_payload);` --> `ndpi_s = ndpi_init_detection_module(); ndpi_set_config(ndpi_s, NULL, "flow.track_payload.enable", "1");`

* All protocols are enabled by default: `ndpi_set_protocol_detection_bitmask2()` has been removed

* `ndpi_finalize_initialization` must be called at the end of the configuration/init phase and its (new) return value must be checked

* Enabling/disabling guessing algorithm in `ndpi_detection_giveup()` is done in general configuration/init phase and no more as specific parameter of this function
TODO

PR #2196, opened 2023-12-05
Add Monero protocol classification.

utoni:
Not sure if it makes sense to unify `monero.c` detection patterns with the one in `mining.c`. Seems like the JSON protocol for ZCash/Monero look pretty similar.


IvanNardi:
> Not sure if it makes sense to unify `monero.c` detection patterns with the one in `mining.c`. Seems like the JSON protocol for ZCash/Monero look pretty similar.

@mmaatuq, any thoughts?

mmaatuq:
> > Not sure if it makes sense to unify `monero.c` detection patterns with the one in `mining.c`. Seems like the JSON protocol for ZCash/Monero look pretty similar.
> 
> @mmaatuq, any thoughts?

yes, it's safe to do that, it was in my pipeline as most of the RPC calls are related to blockchain exploration not mining, also the pattern inside mining.c for ZCash/Monero is too generic it doesn't specify any rpc methode.

utoni:
@mmaatuq 
So it is safe to remove the whole `mining.c` file?

mmaatuq:
> @mmaatuq So it is safe to remove the whole `mining.c` file?

I think so.

IvanNardi:
@mmaatuq, is it fine for you?

mmaatuq:
> @mmaatuq, is it fine for you?

yes, looks good.

lucaderi:
This PR removed the currency that is used by some people, so under the hood, you can change the code but the output must stay rich, hence I think you should preserve the currency name.

IvanNardi:
Let me try to understand better... Let start with two simple questions
AFAIK from previous discussions, we have `NDPI_PROTOCOL_MINING` for the generic "mining" stuff and specific protocols (ETHEREUM,  BITCOIN,...) for the "transactions".
First question: is that correct? @mmaatuq
If so, does this new dissector aim to identify Monero mining or Monero transaction, @utoni ?

utoni:
I think the dissector (and pcap) files are related to the node discovery thingy. I did not record any pcaps for mining/transactions yet.

IvanNardi:
> I think the dissector (and pcap) files are related to the node discovery thingy. I did not record any pcaps for mining/transactions yet.

Ok, so it makes sense to have a new/distinct dissector.
But, and that is my main issue, if the existing mining dissector still identifies *mining* stuff (even if it is not able to tell zcash from monero) I think that we should keep it (with its metadata) and keep having different logic for generic mining and specific cryptocurrency transaction/discovery/...

mmaatuq:
> Let me try to understand better... Let start with two simple questions AFAIK from previous discussions, we have `NDPI_PROTOCOL_MINING` for the generic "mining" stuff and specific protocols (ETHEREUM, BITCOIN,...) for the "transactions". First question: is that correct? @mmaatuq If so, does this new dissector aim to identify Monero mining or Monero transaction, @utoni ?
 yes correct, 
 to elaborate more on this
* patterns that were initially inside mining.c and moved to Bitcoin and Ethereum are related to the generic work of any blockchain node software that tries to connect to the blockchain network(discovery)  and publish, and receive transactions, no pattern is specific to mining.  The same goes for Monero/Zcash pattern it's generic and a lot of Monero rpc calls will have this pattern. 

* detect mining needs new dissectors/sub dissectors for each coin e.g (by looking for rpc calls that specifically are sent from a mining software like [getblocktemplate](https://developer.bitcoin.org/reference/rpc/getblocktemplate.html)  or mining protocols like stratum in case of bitcoin.

IvanNardi:
Proposal: add the new Monero dissector and keep the existing mining code. This way we add dissection capabilities without loosing anything.
Then, later, we might take a closer look at the mining stuff

IvanNardi (changes requested):
Could you:
* update `/doc/protocols.rst`
* restore the `zcash.pcap` test

please?

IvanNardi:
Thanks everyone!

PR #2059, opened 2023-07-26
Realtime output detected protocol

Bit-Warrior-X:

PR #2197, opened 2023-12-05
Add realtime protocol output to `ndpiReader`.

utoni:
This is basically a copy'n'pasta from #2059 with some small changes.
Not sure if we want to keep the "realtime protocol output" - behavior.
But at least the flow callback might be coming handy in the future.
I've stumbled on that "I wanna output something from the flow struct, but the memory is already free'd" - issue before and this is a nice workaround.

### DIFF
--- a/example/ndpiReader.c
+++ b/example/ndpiReader.c
@@ -92,7 +92,7 @@ static char* domain_to_check = NULL;
 static char* ip_port_to_check = NULL;
 static u_int8_t ignore_vlanid = 0;
 /** User preferences **/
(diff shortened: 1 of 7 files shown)


IvanNardi:
The feature is interesting and it might be useful, but I think it should better if it is disabled by default and the user can enable it via a `ndpiReader` flag/parameter

IvanNardi (approved):
We can add (now or later) a new test configuration (in `tests/cfgs`) with one pcap and the this new feature enabled

utoni:
I'll try to do that in a follow-up. :)

PR #2203, opened 2023-12-10
ndpiReader: fix `guessed_flow_protocols` statistic

IvanNardi:
Increment the counter only if the flow has been guessed


utoni (approved):
Always a good thing to improve the readability of the regression tests.

PR #2218, opened 2023-12-19
Rename NDPI_PROTOCOL_RPC to NDPI_PROTOCOL_MS_RPCH

0xA50C1A1:

Describe changes:

I think it's worth renaming this protocol ID to avoid confusion: the HTTP methods [RPC_IN_DATA and RPC_OUT_DATA](https://www.ibm.com/docs/en/sva/9.0.5?topic=configuration-microsoft-rpc-over-http-support) are only used by the MS-RPCH (Microsoft RPC over HTTP) protocol. Previously this protocol ID was also used for DCE/RPC, but now there is a separate one (`NDPI_PROTOCOL_DCERPC`) for it.


src/include/ndpi_protocol_ids.h
```diff
@@ -155,7 +155,7 @@ typedef enum {
   NDPI_PROTOCOL_YOUTUBE               = 124,
   NDPI_PROTOCOL_SKYPE_TEAMS           = 125,
   NDPI_PROTOCOL_GOOGLE                = 126,
-  NDPI_PROTOCOL_RPC                   = 127,
+  NDPI_PROTOCOL_MS_RPCH               = 127,
```
IvanNardi, line 158: Could you add an entry in `doc/protocols.rst`, please (in order)? Basically, with the same info that you already reported in the commit message.
Other than that, it makes a lot of sense. Thanks

IvanNardi: BTW, any chance that you have a pcap example with this kind of traffic? It seems quite difficult to find...

0xA50C1A1: > BTW, any chance that you have a pcap example with this kind of traffic? It seems quite difficult to find...

Unfortunately, I don't have a pcap example, but I heard that this protocol is used by MS Exchange.

0xA50C1A1: Ok, I found the protocol specs  and it looks like this check can be a little bit extended. See[ 2.1.1.1.1.1 RPC Connect Request](https://winprotocoldoc.blob.core.windows.net/productionwindowsarchives/MS-RPCH/%5bMS-RPCH%5d.pdf)

IvanNardi (approved):
Thanks

PR #2221, opened 2023-12-19
mining: a better identification logic

### LINKED ISSUE #2216: Mining dissector gives false positives on any JSON-RPC over TCP traffic
## Describe the bug
Well, this is totally screwed up... With such check, [this dissector](https://github.com/ntop/nDPI/blob/dev/src/lib/protocols/mining.c) will trigger on any JSON-RPC over TCP traffic, even non-mining related.

```
    if(ndpi_strnstr((const char *)packet->payload, "{", packet->payload_packet_len)
	      && (ndpi_strnstr((const char *)packet->payload, "\"method\":", packet->payload_packet_len)
		  || ndpi_strnstr((const char *)packet->payload, "\"blob\":", packet->payload_packet_len)
		  /* || ndpi_strnstr((const char *)packet->payload, "\"id\":", packet->payload_packet_len) - Removed as too generic */
		)
      )
```

Here's the [JSON-RPC](https://www.jsonrpc.org/specification) spec to prove my words.

### Expected behavior

This dissector is worth either deleting or reworking.

### Obtained behavior

Mining dissector gives false positives on [common JSON-RPC over TCP traffic](https://github.com/ntop/nDPI/files/13709766/jsonrpc.zip).

* OS name: Ubuntu
* OS version: 23.04
* Architecture: amd64
* nDPI version or commit hash: from the `dev` branch
* nDPI compilation flags used: --with-pcre2 --with-nbpf-path=/home/vladimir/PF_RING/userland/nbpf/
[config.log](https://github.com/ntop/nDPI/files/13709790/config.log)

- [x] The reported bug is reproducible using ndpiReader.

### If applicable, the used ndpiReader options:

``` shell
ndpiReader -v 2 -V 3 -i jsonrpc.pcap
```

[jsonrpc.zip](https://github.com/ntop/nDPI/files/13709766/jsonrpc.zip)

### Steps to reproduce the behavior:
1. Run 'ndpiReader -v 2 -V 3 -i jsonrpc.pcap'

IvanNardi:
It is quite simple (and not so efficient) but it should fix all the false positives reported in #2216. Add support for Ethereum mining.

Merge all the mining traces.
    
Remove duplicated function.

Close #2216


IvanNardi:
I know that it should be better, but I am not an expert of this stuff...
The main goal is to avoid all the reported false positives and have something better that the current implementation.
It might be improved later, if necessary...

What do you think? @mmaatuq @0xA50C1A1

0xA50C1A1:
> I know that it should be better, but I am not an expert of this stuff... The main goal is to avoid all the reported false positives and have something better that the current implementation. It might be improved later, if necessary...
> 
> What do you think? @mmaatuq @0xA50C1A1

Yeah, it looks a lot better now

utoni (approved):
LGTM. Even tough JSON de-serialization would be also nice. :)

IvanNardi:
> LGTM. Even tough JSON de-serialization would be also nice. :)

Yeah, but a quite different effort to implement :smile:

PR #2223, opened 2023-12-20
CI: update list of compilers

IvanNardi:
Try using latest gcc and clang versions.
We still care about RHEL7: since handling a RHEL7 runner on GitHub is quite complex, let try to use a similar version of gcc, at least


utoni (approved):
Good idea even if `gcc-4.9` is quite old and I do not understand why humans are not willing to upgrade their systems.

PR #2224, opened 2023-12-21
Add WebDAV detection support

0xA50C1A1:

Describe changes:

It's unlikely to find WebDAV over plain HTTP in the wild these days, but I guess it could be useful for enterprise networks where MITM is almost always used. The detection logic is based on HTTP request methods described in RFC4918.


src/lib/ndpi_utils.c
```diff
@@ -2199,6 +2230,11 @@ ndpi_http_method ndpi_http_str2method(const char* method, u_int16_t method_len)
       }
     }
     break;
+
+  case 'U':
+    if (method_len == 6)
+      return(NDPI_HTTP_METHOD_UNLOCK);
+    break;
   }
```
0xA50C1A1, line 2238: Well, it works, but doesn't seem optimal to me. @IvanNardi what do you think? Is there any way this could be improved?

src/lib/ndpi_utils.c
```diff
@@ -2177,17 +2184,41 @@ ndpi_http_method ndpi_http_str2method(const char* method, u_int16_t method_len)
...
     case 'A':return(NDPI_HTTP_METHOD_PATCH);
     case 'O':return(NDPI_HTTP_METHOD_POST);
     case 'U':return(NDPI_HTTP_METHOD_PUT);
+    case 'R':
+      if (method_len >= 5) {
+        if (strncmp(method, "PROPF", 5) == 0)
+          return(NDPI_HTTP_METHOD_PROPFIND);
+        else if (strncmp(method, "PROPP", 5) == 0)
+          return NDPI_HTTP_METHOD_PROPPATCH;
+      }
```
0xA50C1A1, line 2206: Well, it works, but doesn't seem optimal to me. @IvanNardi what do you think? Is there any way this could be improved?

doc/protocols.rst
```diff
@@ -386,3 +386,12 @@ References: `Protocol Specs: <https://opennetworking.org/wp-content/uploads/2014
...
 References: `Protocol Specs: <https://www.jsonrpc.org/specification>`_.
+
+
+.. _Proto 376:
+
+`NDPI_PROTOCOL_WEBDAV`
+======================
+WebDAV is a set of extensions to the HTTP protocol that allows WebDAV clients to collaboratively edit and manage files on remote Web servers.
+
+References: `RFC4918: <https://datatracker.ietf.org/doc/html/rfc4918>`_.
```
IvanNardi, line 397: Could you add a note (similar to HTTP2 and to your commit message) about the fact that it is unlikely to find it in cleartext?

src/lib/ndpi_utils.c
```diff
@@ -2161,6 +2161,13 @@ const char* ndpi_http_method2str(ndpi_http_method m) {
   case NDPI_HTTP_METHOD_RPC_CONNECT:  return("RPC_CONNECT");
   case NDPI_HTTP_METHOD_RPC_IN_DATA:  return("RPC_IN_DATA");
   case NDPI_HTTP_METHOD_RPC_OUT_DATA: return("RPC_OUT_DATA");
+  case NDPI_HTTP_METHOD_MKCOL:        return("MKCOL");
+  case NDPI_HTTP_METHOD_MOVE:         return("MOVE");
+  case NDPI_HTTP_METHOD_COPY:         return("COPY");
+  case NDPI_HTTP_METHOD_LOCK:         return("LOCK");
+  case NDPI_HTTP_METHOD_UNLOCK:       return("UNLOCK");
+  case NDPI_HTTP_METHOD_PROPFIND:     return("PROPFIND");
+  case NDPI_HTTP_METHOD_PROPPATCH:    return("PROPPATCH");
```
IvanNardi, line 2170: Could you update the list in the `#HTTP methods` section of `fuzz/dictionary.dict`, please?

src/lib/ndpi_utils.c
```diff
@@ -2177,17 +2184,41 @@ ndpi_http_method ndpi_http_str2method(const char* method, u_int16_t method_len)
...
+      return(NDPI_HTTP_METHOD_LOCK);
+    break;
+
+  case 'M':
+    if (strncmp(method, "MOVE", 4) == 0)
+      return(NDPI_HTTP_METHOD_MOVE);
+    else if (strncmp(method, "MKCO", 4) == 0)
+      return(NDPI_HTTP_METHOD_MKCOL);
+    break;
+
```
IvanNardi, line 2194: Check (only) for `method[1]` (like the `P` case)

src/lib/ndpi_utils.c
IvanNardi: I think that part it is fine as is.
The general goal is to have "consistent/coherent" code

src/lib/ndpi_utils.c
```diff
@@ -2199,6 +2230,11 @@ ndpi_http_method ndpi_http_str2method(const char* method, u_int16_t method_len)
       }
     }
     break;
+
+  case 'U':
+    if (method_len == 6)
+      return(NDPI_HTTP_METHOD_UNLOCK);
+    break;
```
IvanNardi: Direct return (like `G`, `H`...)

IvanNardi (changes requested):

src/lib/ndpi_utils.c
0xA50C1A1: > Check (only) for `method[1]` (like the `P` case)

Well, there are a two other HTTP methods that start with MK - `MKWORKSPACE` and `MKACTIVITY`. These methods are defined in RFC 3253 (Versioning Extensions to WebDAV), but I thought I wouldn't add it because this standard is marked as proposed. Just thought I'd take a little precaution.

src/lib/ndpi_utils.c
0xA50C1A1: > Direct return (like `G`, `H`...)

RFC 3253 also defines method starting with UN - `UNCHECKOUT`.

src/lib/ndpi_utils.c
0xA50C1A1: Hm... All of this shit only relates to WebDAV anyway, so I guess there's no harm in removing all those extra checks.

IvanNardi (approved):
Thank you

0xA50C1A1:
> Thank you

Thank you for the review as well.

PR #2227, opened 2023-12-22
Fix some warnings reported by CODESonar

IvanNardi:
Remove some unreached/duplicated code.

Add error checking for `atoi()` calls.

About `isdigit()` and similar functions. The warning reported is:
```
Negative Character Value help
isdigit() is invoked here with an argument of signed type char, but only
has defined behavior for int arguments that are either representable
as unsigned char or equal to the value of macro EOF(-1).
Casting the argument to unsigned char will avoid the undefined behavior.
In a number of libc implementations, isdigit() is implemented using lookup
tables (arrays): passing in a negative value can result in a read underrun.
```
Switching to our macros fix that.
Add a check to `check_symbols.sh` to avoid using the original functions
from libc.

### DIFF
--- a/src/lib/ndpi_main.c
+++ b/src/lib/ndpi_main.c
@@ -4539,7 +4539,7 @@ int ndpi_load_categories_file(struct ndpi_detection_module_struct *ndpi_str,
 int load_categories_file_fd(struct ndpi_detection_module_struct *ndpi_str,
 			    FILE *fd, void *user_data) {
   char buffer[512], *line, *name, *category, *saveptr;
-  int len, num = 0;
+  int len, num = 0, cat_id;
 
   if(!ndpi_str || !fd)
     return(-1);
@@ -4562,12 +4562,17 @@ int load_categories_file_fd(struct ndpi_detection_module_struct *ndpi_str,
       category = strtok_r(NULL, "\t", &saveptr);
 
       if(category) {
-	int rc = ndpi_load_category(ndpi_str, name,
-				    (ndpi_protocol_category_t) atoi(category),
-				    user_data);
+        const char *errstrp;
+        cat_id = ndpi_strtonum(category, 1, NDPI_PROTOCOL_NUM_CATEGORIES - 1, &errstrp, 10);
+        if(errstrp == NULL) {
 
-	if(rc >= 0)
-	  num++;
+          int rc = ndpi_load_category(ndpi_str, name,
+				      (ndpi_protocol_category_t)cat_id,
+				      user_data);
+
+          if(rc >= 0)
+           num++;
+        }
       }
     }
   }
@@ -4634,7 +4639,7 @@ int ndpi_load_category_file(struct ndpi_detection_module_struct *ndpi_str,
         line[i] = '\0';
         break;
       }
-      if (line[i] != '-' && line[i] != '.' && isalnum(line[i]) == 0
+      if (line[i] != '-' && line[i] != '.' && ndpi_isalnum(line[i]) == 0
           /* non standard checks for the sake of compatibility */
           && line[i] != '_')
         break;
@@ -4689,19 +4694,19 @@ int ndpi_load_categories_dir(struct ndpi_detection_module_struct *ndpi_str,
 
     /* Check if the format is <proto it>_<string>.<extension> */
     if((underscore = strchr(dp->d_name, '_')) != NULL) {
-      ndpi_protocol_category_t proto_id;
+      int cat_id;
+      const char *errstrp;
 
       underscore[0] = '\0';
-      proto_id = (ndpi_protocol_category_t)atoi(dp->d_name);
-
-      if((proto_id > 0) && (proto_id < (u_int16_t)NDPI_LAST_IMPLEMENTED_PROTOCOL)) {
+      cat_id = ndpi_strtonum(dp->d_name, 1, NDPI_PROTOCOL_NUM_CATEGORIES - 1, &errstrp, 10);
+      if(errstrp == NULL) {
 	/* Valid file */
 	char path[512];
 
 	underscore[0] = '_';
 	snprintf(path, sizeof(path), "%s/%s", dir_path, dp->d_name);
 
-	if (ndpi_load_category_file(ndpi_str, path, proto_id) < 0) {
+	if (ndpi_load_category_file(ndpi_str, path, (ndpi_protocol_category_t)cat_id) < 0) {
 	  NDPI_LOG_ERR(ndpi_str, "Failed to load '%s'\n", path);
 	  failed_files++;
 	}else
@@ -5975,9 +5980,6 @@ int ndpi_handle_ipv6_extension_headers(u_int16_t l3len, const u_int8_t **l4ptr,
 
     *nxt_hdr = (*l4ptr)[0];
 
-    if(*l4len < ehdr_len)
-      return(1);
-
     *l4len -= ehdr_len;
     (*l4ptr) += ehdr_len;
   }
@@ -6033,8 +6035,7 @@ static u_int8_t ndpi_detection_get_l4_internal(struct ndpi_detection_module_stru
   if(l3 == NULL || l3_len < sizeof(struct ndpi_iphdr))
     return(1);
 
-  if((iph = (const struct ndpi_iphdr *) l3) == NULL)
-    return(1);
+  iph = (const struct ndpi_iphdr *) l3;
 
   if(iph->version == IPVERSION && iph->ihl >= 5) {
     NDPI_LOG_DBG2(ndpi_str, "ipv4 header\n");
@@ -6483,7 +6484,7 @@ void ndpi_connection_tracking(struct ndpi_detection_module_struct *ndpi_str,
 	for(i=0; (i<packet->payload_packet_len)
 	      && (flow->flow_payload_len < ndpi_str->max_payload_track_len); i++) {
 	  flow->flow_payload[flow->flow_payload_len++] =
-	    (isprint(packet->payload[i]) || isspace(packet->payload[i])) ? packet->payload[i] : '.';
+	    (ndpi_isprint(packet->payload[i]) || ndpi_isspace(packet->payload[i])) ? packet->payload[i] : '.';
 	}
       }
     }
@@ -7019,7 +7020,7 @@ static void ndpi_reconcile_protocols(struct ndpi_detection_module_struct *ndpi_s
       (MS Teams uses Skype as transport protocol for voice/video)
     */
   case NDPI_PROTOCOL_MSTEAMS:
-    if(flow && (flow->l4_proto == IPPROTO_TCP)) {
+    if(flow->l4_proto == IPPROTO_TCP) {
       // printf("====>> NDPI_PROTOCOL_MSTEAMS\n");
 
       if(ndpi_str->msteams_cache)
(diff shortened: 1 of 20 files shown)


src/lib/protocols/protobuf.c
```diff
@@ -78,7 +78,7 @@ protobuf_dissect_varint(struct ndpi_packet_struct const * const packet,
   size_t i;
   *value = 0;
 
-  for (i = 0; i < 9; ++i)
+  for (i = 0; i < 10; ++i)
```
utoni: Did the protobuf dissector cause false-positives?

utoni (commented):
Maybe add `isdigit` to the symbol check script to make sure that nobody adds it again somewhere else.

src/lib/ndpi_main.c
```diff
@@ -4536,12 +4536,16 @@ int load_categories_file_fd(struct ndpi_detection_module_struct *ndpi_str,
       category = strtok_r(NULL, "\t", &saveptr);
 
       if(category) {
-	int rc = ndpi_load_category(ndpi_str, name,
-				    (ndpi_protocol_category_t) atoi(category),
-				    user_data);
+        cat_id = atoi(category);
```
0xA50C1A1, line 4539: Hmmm... Personally, I'd replace such stuff with `strtonum`. Ofc, this is a BSD specific function, but I guess it's easy to write our own implementation from scratch.

@IvanNardi @utoni what do you think about it?

src/lib/ndpi_main.c
```diff
@@ -4536,12 +4536,16 @@ int load_categories_file_fd(struct ndpi_detection_module_struct *ndpi_str,
...
+        if(cat_id > 0 && cat_id < NDPI_PROTOCOL_NUM_CATEGORIES) {
+
+          int rc = ndpi_load_category(ndpi_str, name,
+				      (ndpi_protocol_category_t)cat_id,
+				      user_data);
```
0xA50C1A1: Hmmm... Personally, I'd replace such stuff with `strtonum`. Ofc, this is a BSD specific function, but I guess it's easy to write our own implementation from scratch.

@IvanNardi @utoni what do you think about it?

utoni: I agree. I remember that I've tried to basically copy&paste the function from `libbsd` a long time ago, but gave up as it did not compile on MingW/MSVC. But if you find a proper solution, I'll approve. :)

0xA50C1A1: > I agree. I remember that I've tried to basically copy&paste the function from `libbsd` a long time ago, but gave up as it did not compile on MingW/MSVC. But if you find a proper solution, I'll approve. :)

Not giga concise, but it works. Requires only stdlib.h, limits.h and errno.h
```
int64_t ndpi_strtonum(const char* numstr, int64_t minval, int64_t maxval, const char** errstrp)
{
  int64_t retval = 0;
  char* endptr;

  if (minval > maxval) {
    *errstrp = "minval > maxval";
    return retval;
  }

  retval = (int64_t)strtoll(numstr, &endptr, 10);

  if (numstr == endptr || *endptr != '\0') {
    *errstrp = "invalid value";
     return 0;
  } else if ((ret == LLONG_MIN && errno == ERANGE) || (ret < minval)) {
    *errstrp = "value too small";
     return 0;
  } else if ((ret == LLONG_MAX && errno == ERANGE) || (ret > maxval ))
    *errstrp = "value too large";
     return 0;
  }

  return retval;
}
```
P.S I've tested it in Visual Studio 2019, it builds and works.

src/lib/protocols/protobuf.c
IvanNardi: The original code was:
```
for (i =0; i < 9; ++i)
{ 
  ...
}
if (i == 10)
  return -1;
```
and the if condition is always false.

First fix I tried:
```
for (i =0; i < 9; ++i)
{ 
  ...
}
if (i == 9)
  return -1;
```
but I got the following diff in the unit tests:
```
protobuf.pcap                                   	ERROR
 ../../../example/ndpiReader -A -p ../../../example/protos.txt -c ../../../example/categories.txt -r ../../../example/risky_domains.txt -j ../../../example/ja3_fingerprints.csv -S ../../../example/sha1_fingerprints.csv -G ../../../lists -q -K JSON -k /dev/null -t -v 2 -i pcap/protobuf.pcap -w /tmp/reader.45775.out  [old vs new]
1,3c1,4
< DPI Packets (TCP):	26	(5.20 pkts/flow)
< Confidence DPI              : 5 (flows)
< Num dissector calls: 707 (141.40 diss/flow)
---
> DPI Packets (TCP):	32	(6.40 pkts/flow)
> Confidence Unknown          : 1 (flows)
> Confidence DPI              : 4 (flows)
> Num dissector calls: 721 (144.20 diss/flow)
5c6
< LRU cache bittorrent: 0/0/0 (insert/search/found)
---
> LRU cache bittorrent: 0/3/0 (insert/search/found)
9c10
< LRU cache mining:     0/0/0 (insert/search/found)
---
> LRU cache mining:     0/1/0 (insert/search/found)
24c25,26
< Protobuf	60	4446	5
---
> Unknown	16	978	1
> Protobuf	44	3468	4
28,30c30,35
< 	3	TCP 127.0.0.1:51680 <-> 127.0.0.1:12345 [proto: 353/Protobuf][IP: 0/Unknown][ClearText][Confidence: DPI][DPI packets: 10][cat: Network/14][9 pkts/588 bytes <-> 7 pkts/390 bytes][Goodput ratio: 15/0][50.00 sec][bytes ratio: 0.202 (Upload)][IAT c2s/s2c min/avg/max/stddev: 0/10000 5714/7500 10000/10000 4949/4330][Pkt Len c2s/s2c min/avg/max/stddev: 54/54 65/56 72/66 8/4][Plen Bins: 100,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0]
< 	4	TCP 127.0.0.1:59030 <-> 127.0.0.1:12345 [proto: 353/Protobuf][IP: 0/Unknown][ClearText][Confidence: DPI][DPI packets: 4][cat: Network/14][5 pkts/377 bytes <-> 3 pkts/174 bytes][Goodput ratio: 25/0][3.00 sec][bytes ratio: 0.368 (Upload)][IAT c2s/s2c min/avg/max/stddev: 0/0 750/0 3000/0 1299/0][Pkt Len c2s/s2c min/avg/max/stddev: 54/54 75/58 149/66 37/6][Plen Bins: 0,0,100,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0]
< 	5	TCP 127.0.0.1:42358 <-> 127.0.0.1:12345 [proto: 353/Protobuf][IP: 0/Unknown][ClearText][Confidence: DPI][DPI packets: 4][cat: Network/14][5 pkts/349 bytes <-> 3 pkts/174 bytes][Goodput ratio: 19/0][< 1 sec][bytes ratio: 0.335 (Upload)][IAT c2s/s2c min/avg/max/stddev: 0/0 0/0 0/0 0/0][Pkt Len c2s/s2c min/avg/max/stddev: 54/54 70/58 121/66 26/6][Plen Bins: 0,0,100,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0]
---
> 	3	TCP 127.0.0.1:59030 <-> 127.0.0.1:12345 [proto: 353/Protobuf][IP: 0/Unknown][ClearText][Confidence: DPI][DPI packets: 4][cat: Network/14][5 pkts/377 bytes <-> 3 pkts/174 bytes][Goodput ratio: 25/0][3.00 sec][bytes ratio: 0.368 (Upload)][IAT c2s/s2c min/avg/max/stddev: 0/0 750/0 3000/0 1299/0][Pkt Len c2s/s2c min/avg/max/stddev: 54/54 75/58 149/66 37/6][Plen Bins: 0,0,100,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0]
> 	4	TCP 127.0.0.1:42358 <-> 127.0.0.1:12345 [proto: 353/Protobuf][IP: 0/Unknown][ClearText][Confidence: DPI][DPI packets: 4][cat: Network/14][5 pkts/349 bytes <-> 3 pkts/174 bytes][Goodput ratio: 19/0][< 1 sec][bytes ratio: 0.335 (Upload)][IAT c2s/s2c min/avg/max/stddev: 0/0 0/0 0/0 0/0][Pkt Len c2s/s2c min/avg/max/stddev: 54/54 70/58 121/66 26/6][Plen Bins: 0,0,100,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0]
> 
> 
> Undetected flows:
> 	1	TCP 127.0.0.1:51680 <-> 127.0.0.1:12345 [proto: 0/Unknown][IP: 0/Unknown][ClearText][Confidence: Unknown][DPI packets: 16][9 pkts/588 bytes <-> 7 pkts/390 bytes][Goodput ratio: 15/0][50.00 sec][bytes ratio: 0.202 (Upload)][IAT c2s/s2c min/avg/max/stddev: 0/10000 5714/7500 10000/10000 4949/4330][Pkt Len c2s/s2c min/avg/max/stddev: 54/54 65/56 72/66 8/4][Plen Bins: 100,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0]
```
I am not able to tell if that diff makes sense or not...
So I tried to fix using "10" and I got no differences...

Not sure which solution is best... @utoni what do you think?

src/lib/ndpi_main.c
IvanNardi: Switched to `ndpi_strtonum`

IvanNardi:
> Maybe add `isdigit` to the symbol check script to make sure that nobody adds it again somewhere else.

I added this checks to the script anyway, but I am not able to trigger the error... maybe because these functions are macros or are inlined and so we don't "see" them in the object file? :(

0xA50C1A1:
> > Maybe add `isdigit` to the symbol check script to make sure that nobody adds it again somewhere else.
> 
> I added this checks to the script anyway, but I am not able to trigger the error... maybe because these functions are macros or are inlined and so we don't "see" them in the object file? :(

I think it's better to use function poisoning for it.

IvanNardi:
> > > Maybe add `isdigit` to the symbol check script to make sure that nobody adds it again somewhere else.
> > 
> > 
> > I added this checks to the script anyway, but I am not able to trigger the error... maybe because these functions are macros or are inlined and so we don't "see" them in the object file? :(
> 
> I think it's better to use function poisoning for it.

I tried something like `#pragma GCC poison isdigit` and it works, but it generates a lot of this annoying warning

```
In file included from ndpi_geoip.c:30:
ndpi_private.h:17:20: warning: poisoning existing macro "isdigit"
   17 | #pragma GCC poison isdigit

```

If I try to poison a function (instead of a macro), no warnings, so it seems that warning is triggered only when poisoning  a macro; no idea how to disable it

utoni (approved):

src/lib/protocols/protobuf.c
IvanNardi: @utoni , I saw you approved this PR. Could you take also a look at the above comment, please?

utoni: Strange, looking at the code snipped it should work as you described..

utoni: `if (i == 10)` can be removed, I think that I just forgot to remove it in an earlier iteration

IvanNardi: Thanks. In the last push, I restored the original loop and remove the "if" check

PR #2236, opened 2024-01-02
ndpiReader: add breed stats on output used for CI

IvanNardi:


tests/cfgs/caches_cfg/result/teams.pcap.out
```diff
@@ -43,6 +43,11 @@ Microsoft365	136	52120	6
 Teams	595	215358	26
 Azure	2	294	1
 
+Safe                        564758
```
utoni: This value does refer to the processed bytes, right?
Having the amount flows that have that flow breed might also be a nice thing.

IvanNardi: Good idea! I added pkts/bytes/flows stats (in standard and "CI" outputs), like we already do for protocol statistics

PR #2243, opened 2024-01-04
Make some test traces smaller

IvanNardi:
Having smaller traces help fuzzing: we want the fuzzers to mutate "interesting" packets analyzed by nDPI, i.e. the first packets of each flows.
Try hard to keep the same classification and extraction capabilities


utoni:
This is quite a rough change to the regression tests. Does it make more sense to generate a single dedicated pcap file used for corpus generation?

IvanNardi:
> This is quite a rough change to the regression tests. Does it make more sense to generate a single dedicated pcap file used for corpus generation?

I manually reviewed session by session to remove uninteresting flows and to reduce the bigger one: I think it should be difficult to automate the process
Anyway, I share your concerns
I'll pull a simpler version of this PR with a simpler diff, that I hope will simplify the review process

utoni (approved):
Good job, thanks!

PR #2245, opened 2024-01-08
Fix default port conflicts between AJP and CiscoVPN

IvanNardi:
We shouldn't have the same default port associated with multiple protocols. The key reason is that we use only one of them (for classification by-port, for example) and which one we really use depends on the internal order of the protocols's initialization (which is unknown to the user).

We have 3 port collisions:
* 8009: AJP and CiscoVPN
* 445: SMBv1 and SMBv23
* 8080: HTTP_Connect and HTTP_Proxy

I think that the two last cases don't really matter...

About the first one: it seems that CiscoVPN is retired from 2011/2014 (see
https://www.cisco.com/c/en/us/obsolete/security/cisco-vpn-client.html) and superseded by AnyConnect, which uses different ports (see https://community.cisco.com/t5/network-security/what-protocols-are-used-on-anyconnect-mobility-client/td-p/4713525). Therefore it should be safe to remove 8009 from the CiscoVPN default ports.

### DIFF
--- a/src/lib/ndpi_main.c
+++ b/src/lib/ndpi_main.c
@@ -1661,7 +1661,7 @@ static void ndpi_init_protocol_defaults(struct ndpi_detection_module_struct *ndp
 			  ndpi_build_default_ports(ports_b, 0, 0, 0, 0, 0) /* UDP */);
   ndpi_set_proto_defaults(ndpi_str, 0 /* encrypted */, 0 /* nw proto */, NDPI_PROTOCOL_ACCEPTABLE, NDPI_PROTOCOL_CISCOVPN,
 			  "CiscoVPN", NDPI_PROTOCOL_CATEGORY_VPN,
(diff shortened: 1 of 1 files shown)


0xA50C1A1:
> We have 3 port collisions:
> 
>     * 445: SMBv1 and SMBv23

Is there any point in separating them? Maybe it would be worth aggregating these two IDs under one NDPI_SMB_CIFS id and set a risk if SMB version 1 is detected?

IvanNardi:
> > We have 3 port collisions:
> > ```
> > * 445: SMBv1 and SMBv23
> > ```
> 
> Is there any point in separating them? Maybe it would be worth aggregating these two IDs under one NDPI_SMB_CIFS id and set a risk if SMB version 1 is detected?

Good question! And we already set a risk with SMBv1... Let think about that as a possible follow-up...

utoni (approved):

PR #2249, opened 2024-01-09
New API for library configuration

IvanNardi:
This is the first step into providing (more) configuration options in nDPI.

The idea is to have a simple way to configure (most of) nDPI: only one function (`ndpi_set_config()`) to set any configuration parameters (in the present or on in the future) and we try to keep this function prototype as agnostic as possible.

You can configure the library:
* via API, using `ndpi_set_config()`
* via a configuration file, in a text format

* we don't break API/ABI compatibility of the library; even changing the parameter type (from integer to a list of integer, for example) doesn't break the compatibility.

Most of the existing configuration options has been ported to the new API and a few ones has been added for TLS, HTTP and DNS protocols.
More details are provided in the commit messages

The complete list of configuration options is provided in `doc/configuration_parameters.md`.

***API changes***

Almost all nDPI configuration is made via `ndpi_set_config()` and lots of configuration functions have been removed.
Important examples of API changes:

* Enumeration `ndpi_init_prefs` has been removed: `struct ndpi_detection_module_struct *ndpi_init_detection_module(ndpi_init_prefs prefs)` --> `struct ndpi_detection_module_struct *ndpi_init_detection_module(void)`. \
Example: `ndpi_s = ndpi_init_detection_module(ndpi_track_flow_payload);` --> `ndpi_s = ndpi_init_detection_module(); ndpi_set_config(ndpi_s, NULL, "flow.track_payload.enable", "1");`

* `ndpi_finalize_initialization` must be called at the end of the configuration/init phase and its (new) return value must be checked

* Enabling/disabling guessing algorithm in `ndpi_detection_giveup()` is done in general configuration/init phase and no more as specific parameter of this function

***Next***

This is only the first step. The idea is to move protocol configuration (removing `ndpi_set_protocol_detection_bitmask2()`) and file loading (custom protocols, signatures, custom categories...) to the new API.
#2190 showed that it is possible. However we must take another closer look at the iteration between custom rules/protocols/categories loading and the list of the disabled protocols


IvanNardi:
Placeholder

IvanNardi:
Placeholder2

doc/configuration_parameters.md
```diff
@@ -0,0 +1,36 @@
...
+| NULL         | "fully_encrypted_heuristic.enable"        | 1             | NULL      | NULL      | Enable/disable an heuristic to detect fully encrypted sessions, i.e. flows where every bytes of the payload is encrypted in an attempt to “look like nothing”. This heuristic only analyzes the first packet of the flow. See: https://www.usenix.org/system/files/sec23fall-prepub-234-wu-mingshi.pdf |
+| NULL         | "libgcrypt.init"                          | 1             | NULL      | NULL      | Enable/disable initialization of libgcrypt. When using the external libgcrypt (instead of the internal crypto code) the libgcrypt runtime must be initialized. If, for whatever reasons, the application alread does it, nDPI must be told to skip it |
```
utoni: Does this option actually work independently of the autoconf parameter?

IvanNardi: The behavior doesn't change with this PR: the libgcrypt initialization is a nop if libgcrypt is not configured via autoconf

example/ndpiReader.c
```diff
@@ -92,14 +92,21 @@ static char* domain_to_check = NULL;
...
-int nDPI_LogLevel = 0;
-char *_debug_protocols = NULL;
+
+struct cfg {
+  char *proto;
+  char *param;
+  char *value;
+};
+#define MAX_NUM_CFGS 32
+static struct cfg cfgs[MAX_NUM_CFGS];
```
utoni, line 104: Is it worth to make this a dynamic allocated array depending on the amount of used config params?

example/ndpiReader.c
```diff
@@ -874,81 +885,84 @@ void printCSVHeader() {
...
+        }
+        return 0;
       }
     }
   }
   ndpi_free(tmp_str);
   return -1;
 }
 
+int __add_cfg(char *proto, char *param, char *value, int dup)
```
utoni: What about:

```suggestion
int ndpi_add_cfg(char *proto, char *param, char *value, int dup)
```

or 

```suggestion
int reader_add_cfg(char *proto, char *param, char *value, int dup)
```

example/ndpiReader.c
```diff
@@ -1100,55 +1106,61 @@ static void parseOptions(int argc, char **argv) {
...
-      if(nDPI_LogLevel < NDPI_LOG_ERROR) nDPI_LogLevel = NDPI_LOG_ERROR;
-      if(nDPI_LogLevel > NDPI_LOG_DEBUG_EXTRA) {
-        nDPI_LogLevel = NDPI_LOG_DEBUG_EXTRA;
-        ndpi_free(_debug_protocols);
-        _debug_protocols = ndpi_strdup("all");
+    {
+      char buf[8];
+      int log_level;
+
+      log_level = atoi(optarg);
```
utoni: With `ndpi_strtonum` one could also check for conversion errors. :)

utoni (commented):
Thanks for this complex change!
It'll require some integration effort for applications on top of nDPI, but I guess it is worth to unify nDPI configuration.

doc/configuration_parameters.md
IvanNardi: @utoni , I will make the description clearer anyway, thanks!

utoni:
Just curious, because I was not able to find it in the code:

> 
>     * via a configuration file, in a text format
> 

So it is possible to pass what exactly? A file descriptor or a string as configuration file? Is it part of `ndpiReader` or `libnDPI`?

IvanNardi:
> So it is possible to pass what exactly? A file descriptor or a string as configuration file? Is it part of `ndpiReader` or `libnDPI`?

libnDPI, via file name: `ndpi_set_config(NULL, "filename.config", $FILENAME)`; you can find an example of the file in `example/config.txt`

src/lib/ndpi_main.c
```diff
@@ -4502,6 +4521,65 @@ static int ndpi_handle_rule(struct ndpi_detection_module_struct *ndpi_str,
...
+      break;
+
+    len = strlen(line);
+
+    if((len <= 1) || (line[0] == '#'))
+      continue;
+
+    line[len - 1] = '\0';
+
+    if(line[0] == ',') { /* First parameter might be missing */
```
utoni: That makes some config files/strings kinda ugly. Another possible way could be to use the return value of `strtok_r` to also count the amount of found `","`s.

IvanNardi: Let's take a look at `example/config.txt`. Are you suggetsing something like: `,packets_limit_per_flow,32` -> `packets_limit_per_flow,32`?

utoni: Yea, exactly. :D

doc/configuration_parameters.md
IvanNardi: Updated in the last iteration

example/ndpiReader.c
IvanNardi: Done. I choose the second option. I prefer to keep the "ndpi_" prefix for the public API of the library

example/ndpiReader.c

example/ndpiReader.c
IvanNardi: I think we can safely ignore that: 32 is plenty of space, the limit is always checked and we are in the application example... [if that code had been in the library, I would have used the realloc stuff...]

src/lib/ndpi_main.c
IvanNardi: Ok, it is a good idea. I will try that in the next iteration

IvanNardi: Done

utoni (approved):
:+1:

Issue #2246, opened 2024-01-08
Merging SMBv1 and SMBv23?


IvanNardi:
Should we have only one protocol id for SMB1 and SMBv2, both?
See: https://github.com/ntop/nDPI/pull/2245#issuecomment-1880926756

0xA50C1A1:
Well, that sounds logical. This seems to be the way [SSH detection](https://github.com/ntop/nDPI/blob/90b999e3f1a67babbcd7003fb0ccab005d45f343/src/lib/protocols/ssh.c#L71C1-L121C2) is implemented. I.e. one ID for all protocol versions and risk setting in case an obsolete version is detected.

PR #2251, opened 2024-01-09
Merge SMBv1 and SMBv23 into a single protocol ID

0xA50C1A1:

See #2246


tests/cfgs/default/result/1kxun.pcap.out
```diff
@@ -45,9 +45,8 @@ MpegDash	1	299	1
 Line	30	19034	3
 
 Safe                           124 28754         9            
-Acceptable                     622 514902        119          
+Acceptable                     627 516099        121          
 Fun                            948 1976493       53           
-Dangerous                        5 1197          2
```
IvanNardi, line 50: I think that might be the reason that the idea of merging smb1 and smb23 doesn't work...
We don't want this regression.
However, nDPI has only the concept of "breed per protocol", no "breed per flow", i.e. we cant' have right now two different flows with the same classification but with different breeds...

PR #2255, opened 2024-01-11
Improved uTorrent via utp (TCP-like streams over UDP).

utoni:


src/lib/protocols/bittorrent.c
```diff
@@ -68,7 +76,13 @@ static u_int8_t is_utpv1_pkt(const u_int8_t *payload, u_int payload_len) {
   if(h->h_version != 1)             return(0);
   if(h->h_type > 4)                 return(0);
   if(h->next_extension > 2)         return(0);
-  if(ntohl(h->window_size) > 65565) return(0);
```
utoni, line 71: Checking window size is not useful as different implementations enforce different window size's.

src/lib/protocols/bittorrent.c
```diff
@@ -531,7 +545,11 @@ static void ndpi_search_bittorrent(struct ndpi_detection_module_struct *ndpi_str
 
 	  if(is_utpv1_pkt(packet->payload, packet->payload_packet_len)) {
 	    bt_proto = ndpi_strnstr((const char *)&packet->payload[20], BITTORRENT_PROTO_STRING, packet->payload_packet_len-20);
-	    goto bittorrent_found;
+	    if(flow->packet_counter > 2) {
```
IvanNardi: Would make sense something like `if(flow->packet_counter > 2 || bt_proto) {` ?
If we find the BitTorrent string, maybe we can stop here. What do you think?

IvanNardi (approved):
Nice improvement!

PR #2257, opened 2024-01-12
Add KCP protocol dissector.

utoni:


src/lib/protocols/kcp.c
```diff
@@ -0,0 +1,107 @@
...
+  ndpi_int_kcp_add_connection(ndpi_struct, flow);
+}
+
+void init_kcp_dissector(struct ndpi_detection_module_struct *ndpi_struct,
+                        u_int32_t *id)
+{
+  ndpi_set_bitmask_protocol_detection("KCP", ndpi_struct, *id,
+    NDPI_PROTOCOL_KCP,
+    ndpi_search_kcp,
+    NDPI_SELECTION_BITMASK_PROTOCOL_V4_V6_TCP_OR_UDP_WITH_PAYLOAD_WITHOUT_RETRANSMISSION,
```
IvanNardi, line 101: Do you have an example of TCP traffic?

utoni: Not yet, but according to the source code basically any layer4 protocol can be used for KCP encapsulation.

IvanNardi (approved):

Issue #1828, opened 2022-12-10
Endianess issue


IvanNardi:
It seems nDPI has an endianess issue. The attached flow is classified as `BitTorrent` in little-endian machines, but as `RTP.Zoom` in big-endian ones.
See https://github.com/ntop/nDPI/actions/runs/3665656103/jobs/6196872432

[BTW, the correct one is definitely `RTP.Zoom`, but this ticket is mainly about the different results on different architectures.]

[big_endian_issue.zip](https://github.com/ntop/nDPI/files/10200836/big_endian_issue.zip)

***Updated example:*** https://github.com/ntop/nDPI/issues/1828

PR #2259, opened 2024-01-12
Bittorrent: improve detection of UTPv1

IvanNardi:
* fix header length check: we must take into account optional extension header
* it seems the check on UTPv1 header is weak only in the DATA case: wait for multiple packets only in that case


src/lib/protocols/bittorrent.c
```diff
@@ -69,24 +69,48 @@ static int search_bittorrent_again(struct ndpi_detection_module_struct *ndpi_str
...
 /* *********************************************** */
 
+static int get_utpv1_length(const u_int8_t *payload, u_int payload_len)
+{
+  struct ndpi_utp_hdr *h = (struct ndpi_utp_hdr*)payload;
+  unsigned int off;
+  u_int8_t ext_type = h->next_extension;
+
+  off = sizeof(struct ndpi_utp_hdr);
+  while(ext_type != 0 && off + 1 < payload_len) {
```
utoni, line 79: If I am not mistaken `libutp` has a hard limit of maximum allowed extensions. Was it 4? Well, I can not even remember stuff I read 2 days ago. :D

IvanNardi: 2 days is a looong time :)
I didn't find anything in `libutp` but putting an hard limit makes a lot of sense

utoni: From https://github.com/bittorrent/libutp/blob/master/utp_internal.cpp on line 510:

```C
	// extension bytes from SYN packet
	byte extensions[8];
```

// EDIT: There is no hard limit. But 8 could work. :)

IvanNardi: This field is used to save the *payload* of the extension type 2, only. see around line 1857. Am I wrong?

utoni: Yea, no hard limit for sure. :D

utoni (approved):

IvanNardi:
The endianess issue is unrelated to these changes (see https://github.com/ntop/nDPI/pull/2260)
Since I don't have time to look into that (see #1828 for tracking the issue), I remove the new trace: it is kind of cheating, but I don't want to have failures on the CI...

0xA50C1A1:
> The endianess issue is unrelated to these changes (see #2260) Since I don't have time to look into that (see #1828 for tracking the issue), I remove the new trace: it is kind of cheating, but I don't want to have failures on the CI...

Btw, what was the problem? I've got a s390x Alpine Linux VM, so I guess I can tinker with this stuff at my leisure.

IvanNardi:
> > The endianess issue is unrelated to these changes (see #2260) Since I don't have time to look into that (see #1828 for tracking the issue), I remove the new trace: it is kind of cheating, but I don't want to have failures on the CI...
> 
> Btw, what was the problem? I've got a s390x Alpine Linux VM, so I guess I can tinker with this stuff at my leisure.

The pcap attached in #2260 returns different results only on s390x CI test

0xA50C1A1:
> > > The endianess issue is unrelated to these changes (see #2260) Since I don't have time to look into that (see #1828 for tracking the issue), I remove the new trace: it is kind of cheating, but I don't want to have failures on the CI...
> > 
> > 
> > Btw, what was the problem? I've got a s390x Alpine Linux VM, so I guess I can tinker with this stuff at my leisure.
> 
> The pcap attached in #2260 returns different results only on s390x CI test

Ok, I'll create a PR when I figure out what's wrong.

IvanNardi:
> > > > The endianess issue is unrelated to these changes (see #2260) Since I don't have time to look into that (see #1828 for tracking the issue), I remove the new trace: it is kind of cheating, but I don't want to have failures on the CI...
> > > 
> > > 
> > > Btw, what was the problem? I've got a s390x Alpine Linux VM, so I guess I can tinker with this stuff at my leisure.
> > 
> > 
> > The pcap attached in #2260 returns different results only on s390x CI test
> 
> Ok, I'll create a PR when I figure out what's wrong.

I have seen similar issues for **years** so I don't think that the issue is the newer code....

PR #2261, opened 2024-01-15
ndpiReader: improve the check on max number of pkts processed per flow

IvanNardi:
Allow to disable this check.

I don't know how much sense these limits have in the application (especially with those default values...) since we have always had a hard limit on the library itself (`max_packets_to_process` set to 32). The only value might be that they provide different limits for TCP and UDP traffic.

Keep them for the time being...

### DIFF
--- a/example/ndpiReader.c
+++ b/example/ndpiReader.c
@@ -1269,7 +1269,8 @@ static void parseOptions(int argc, char **argv) {
 
     case 'T':
       max_num_tcp_dissected_pkts = atoi(optarg);
-      if(max_num_tcp_dissected_pkts < 3) max_num_tcp_dissected_pkts = 3;
+      /* If we enable that, allow at least 3WHS + 1 "real" packet */
+      if(max_num_tcp_dissected_pkts != 0 && max_num_tcp_dissected_pkts < 4) max_num_tcp_dissected_pkts = 4;
       break;
 
     case 'x':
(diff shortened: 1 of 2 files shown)


example/ndpiReader.c
```diff
@@ -1269,7 +1269,6 @@ static void parseOptions(int argc, char **argv) {
 
     case 'T':
       max_num_tcp_dissected_pkts = atoi(optarg);
-      if(max_num_tcp_dissected_pkts < 3) max_num_tcp_dissected_pkts = 3;
```
utoni, line 1272: Does it make sense for TCP to allow dissecting less than 3 packets if `SYN` and `ACK` packets also count?

IvanNardi: Probably not. Something like `if(max_num_tcp_dissected_pkts < 3 && max_num_tcp_dissected_pkts != 0) max_num_tcp_dissected_pkts = 3;`?

utoni: Either that one or stop counting TCP packets that do not carry any payload. Not sure what is the best here. IMHO your solution is a "low hanging fruit" compared to implementation effort and outcome.

IvanNardi: Let's keep the same logic; it is only an example showing what any real application could/might do

utoni (approved):

PR #2269, opened 2024-01-19
Add Mumble detection support

0xA50C1A1:

Describe changes:

Mumble remains popular (mostly because it's free software) even in the Discord era, so I thought it would be nice to add a protocol dissector for it.

Mumble uses both UDP (for audio stream) and TCP (auth, getting info about the server, etc.), but in the latter case there's nothing to look at as all traffic is TLS encrypted. Well, the detection logic is pretty simple and is based on UDP Ping packet inspection ([a description can be found here](https://wiki.mumble.info/wiki/Protocol)).


utoni (approved):
Thanks!

PR #2271, opened 2024-01-19
Allow multiple `struct ndpi_detection_module_struct` to share some state

IvanNardi:
Add the concept of "global context".

Right now every instance of `struct ndpi_detection_module_struct` (we
will call it "local context" in this description) is completely
independent from each other. This provide optimal performances in
multithreaded environment, where we pin each local context to a thread,
and each thread to a specific CPU core: we don't have any data shared
across the cores.

Each local context has, internally, also some information correlating
**different** flows; something like:
```
if flow1 (PeerA <-> Peer B) is PROTOCOL_X; then
  flow2 (PeerC <-> PeerD) will be PROTOCOL_Y
```
To get optimal classification results, both flow1 and flow2 must be
processed by the same local context. This is not an issue at all in the far
most common scenario where there is only one local context, but it might
be impractical in some more complex scenarios.

Create the concept of "global context": multiple local contexts can use
the same global context and share some data (structures) using it.
This way the data correlating multiple flows can be read/write from
different local contexts.
This is an optional feature, disabled by default.

Obviously data structures shared in a global context must be thread safe.
This PR updates the code of the LRU implementation to be, optionally,
thread safe.

Right now, only the LRU caches can be shared; the other main structures
(trees and automas) are basically read-only: there is little sense in
sharing them. Furthermore, these structures don't have any information
correlating multiple flows.

Every LRU cache can be shared, independently from the others, via
`ndpi_set_config(ndpi_struct, NULL, "lru.$CACHE_NAME.scope", "1")`.

It's up to the user to find the right trade-off between performances
(i.e. without shared data) and classification results (i.e. with some
shared data among the local contexts), depending on the specific traffic
patterns and on the algorithms used to balance the flows across the
threads/cores/local contexts.

Add some basic examples of library initialization in
`doc/library_initialization.md`.

This code needs libpthread as external dependency. It shouldn't be a big
issue; however a configure flag has been added to disable global context
support. A new CI job has been added to test it.

TODO: we should need to find a proper way to add some tests on
multithreaded enviroment... not an easy task...

***API changes*** 

If you are not interested in this feature, simply add a NULL parameter to
any `ndpi_init_detection_module()` calls.


IvanNardi:
This PR adds libpthread as mandatory dependency. I don't think this is a big issue. However, if anyone thinks otherwise, we can add a new flag on the configure script.

utoni:
> This PR adds libpthread as mandatory dependency. I don't think this is a big issue. However, if anyone thinks otherwise, we can add a new flag on the configure script.

True. But I would like having `libnDPI` building with minimal dependencies.

IvanNardi:
> > This PR adds libpthread as mandatory dependency. I don't think this is a big issue. However, if anyone thinks otherwise, we can add a new flag on the configure script.
> 
> True. But I would like having `libnDPI` building with minimal dependencies.

From the last commit message

```
This code needs libpthread as external dependency. It shouldn't be a big
issue; however a configure flag has been added to disable global context
support. A new CI job has been added to test it. Note taht ndpiReader
code uses global context and so we can't run internal tests without it.
```

Note that `ndpiReader` already depends on libpthread

IvanNardi:
> > > This PR adds libpthread as mandatory dependency. I don't think this is a big issue. However, if anyone thinks otherwise, we can add a new flag on the configure script.
> > 
> > 
> > True. But I would like having `libnDPI` building with minimal dependencies.
> 
> From the last commit message
> 
> ```
> This code needs libpthread as external dependency. It shouldn't be a big
> issue; however a configure flag has been added to disable global context
> support. A new CI job has been added to test it. Note taht ndpiReader
> code uses global context and so we can't run internal tests without it.
> ```
> 
> Note that `ndpiReader` already depends on libpthread

EDIT: with last push, ndpiReader (and internal tests) works even if nDPI has been compiled without global context support

utoni:
> Note that `ndpiReader` already depends on libpthread

I am usually building nDPI with `--with-only-libndpi`. :)

IvanNardi:
@utoni, are you fine with the new flag `--disable-global-context-support`?

utoni:
Yea, totally fine. :)

utoni (approved):
forgot to approve..

PR #2279, opened 2024-01-22
Improve MySQL detection

0xA50C1A1:

Describe changes:

Slight improvement of MySQL dissector: added MariaDB support and more stringent checks.


src/lib/protocols/mysql.c
```diff
@@ -30,50 +31,27 @@
...
+    if ((u_int32_t)(packet->payload_packet_len-4) == length && 
+        packet->payload[4] == 0x0A && ((memcmp(&packet->payload[5], "5.5.5-", 6) == 0) || 
+        (packet->payload[5] > 0x33 && packet->payload[5] < 0x39)))
+    {
+      if ((memcmp(&packet->payload[packet->payload_packet_len-10], "_password", 9) == 0) ||
+          (memcmp(&packet->payload[packet->payload_packet_len-10], "_kerberos", 9) == 0) ||
+          (memcmp(&packet->payload[packet->payload_packet_len-9], "_windows", 8) == 0) ||
+          (memcmp(&packet->payload[packet->payload_packet_len-8], "_simple", 7) == 0) ||
+          (memcmp(&packet->payload[packet->payload_packet_len-8], "_gssapi", 7) == 0) ||
+          (memcmp(&packet->payload[packet->payload_packet_len-5], "_pam", 4) == 0))
```
0xA50C1A1, line 50: Is it worth to add checks for other [authentication plugins](https://dev.mysql.com/doc/refman/8.0/en/authentication-plugins.html)?

0xA50C1A1:
@IvanNardi @utoni guys, what do you think about this? I bumped the minimum MySQL version in check from 1 to 4 since there's not many people using these ancient relics.

Well, also authentication plugin support was introduced only in MySQL 4, if memory serves me right.

src/lib/protocols/mysql.c
IvanNardi: I think that it is fine as-is, unless you have some traces with the other plugins

IvanNardi (approved):
Thanks

PR #2281, opened 2024-01-23
fuzz: extend fuzzing coverage

IvanNardi:


src/lib/ndpi_main.c
```diff
@@ -8883,13 +8775,13 @@ void change_category(struct ndpi_detection_module_struct *ndpi_str, struct ndpi_
 
 /* ********************************************************************************* */
 
+#ifdef CODE_UNUSED
```
utoni: Why not remove that `#ifdef`ed code?

IvanNardi: Great idea! Done

utoni (approved):
LGTM.
Note: In theory this PR is an API and ABI break (`ac_automa_finalized` and `check_for_email_address` although both were declared in a very private header file :smiley:)

// EDIT: Not forgetting about `ndpi_check_flow_func`

IvanNardi:
> LGTM. Note: In theory this PR is an API and ABI break (`ac_automa_finalized` and `check_for_email_address` although both were declared in a very private header file 😃)
> 
> // EDIT: Not forgetting about `ndpi_check_flow_func`

I don't consider changes in a private header as  API/ABI breaks, regardless of *where* this header is :)
Applications shouldn't use these kind of functions anyway

`ndpi_check_flow_func` was in a public header, though. And it was there since the OpenDPI era. However I am quite sure that it never worked in nDPI: to use it without calling `ndpi_detection_process_packet`, the application should have populate `ndpi_struct->packet` itself, but `ndpi_struct` has been an opaque structure for many years.
Bottom line: public function since the beginning, but never usable

PR #2286, opened 2024-01-25
Add Raft protocol dissector.

utoni:


src/lib/protocols/raft.c
```diff
@@ -0,0 +1,111 @@
...
+
+#define NDPI_CURRENT_PROTO NDPI_PROTOCOL_RAFT
+#include "ndpi_api.h"
+#include "ndpi_private.h"
+
+PACK_ON
+struct raft_header {
+  uint64_t msg_type;
+  uint64_t msg_length;
+} PACK_OFF;
```
IvanNardi, line 33: I think that sooner or later we need to take a closer look if/when all these PACK attributes are really needed...

IvanNardi (approved):

src/lib/protocols/raft.c
utoni: I think for that specific case, it should be fine to remove `PACK_*` iff we assume that structs in C are always 4-byte aligned.
My fear is that different CCs may have a different struct alignment.

Issue #2287, opened 2024-01-26
RESP protocol dissector implementation


0xA50C1A1:
@IvanNardi @utoni hey guys. Atm I'm working on a [RESP protocol](https://github.com/redis/redis-specifications/blob/master/protocol/RESP3.md) dissector, but I've encountered some issues and I'm looking for expert advice.

RESP PDUs start with `*` followed by the field length and `\r\n`, so the first thing that came to mind was this check:

```
  if (packet->payload_packet_len > 10 && packet->payload[0] == '*' &&
      ndpi_strnstr((const char *)packet->payload, "\r\n", packet->payload_packet_len))
  {
    NDPI_LOG_INFO(ndpi_struct, "found RESP\n");
    ndpi_set_detected_protocol(ndpi_struct, flow, NDPI_PROTOCOL_RESP,
                               NDPI_PROTOCOL_UNKNOWN, NDPI_CONFIDENCE_DPI);
    return;
  }
```

But I suspect that such a simple check could potentially cause a lot of false positives - even if I haven't seen it on pcap examples from nDPI, it doesn't guarantee anything. Or maybe I'm worrying for nothing?

I thought about adding a check like ```if (flow->l4.tcp.resp_valid_packets > 2) {...} if(flow->packet_counter > 5) { NDPI_EXCLUDE_PROTO(ndpi_struct, flow); } ```, but that breaks the detection of some flows since `* .. \r\n` is often located not at the start of the segment.

utoni:
Is it possible to use the field length for a successful detection?

// EDIT: Did not read the protocol specs yet. :smile:

0xA50C1A1:
> Is it possible to use the field length for a successful detection?
> 
> // EDIT: Did not read the protocol specs yet. 😄

Well, I'm not sure about that. Btw, here are some pcap samples [resp.zip](https://github.com/ntop/nDPI/files/14066740/resp.zip)

IvanNardi:
Something like (in pseudo code)?
```
if(payload[0] == '*' &&
    (from offeset 1, one or more bytes are a number in ascii format) &&
    (after the ascii characters there is "\r\n") {
        FOUND
        return;
}
EXCLUDE
```

It should be enough...

0xA50C1A1:
> Something like (in pseudo code)?
> 
> ```
> if(payload[0] == '*' &&
>     (from offeset 1, one or more bytes are a number in ascii format) &&
>     (after the ascii characters there is "\r\n") {
[...]

PR #2288, opened 2024-01-27
Fix RESP detection

0xA50C1A1:

Link to the related issue: #2287

Describe changes:

Fixed broken dissector, added RESP pcap sample and renamed the protocol ID from NDPI_PROTOCOL_REDIS to NDPI_PROTOCOL_RESP for correctness.


src/lib/protocols/resp.c
```diff
@@ -0,0 +1,81 @@
...
+    if (!isascii(packet->payload[offset]))
+      goto exclude;
+    offset++;
+  }
+
+  if (memcmp(&packet->payload[offset], "\r\n", 2) == 0) {
+    NDPI_LOG_INFO(ndpi_struct, "found RESP\n");
+    ndpi_set_detected_protocol(ndpi_struct, flow, NDPI_PROTOCOL_RESP,
+                               NDPI_PROTOCOL_UNKNOWN, NDPI_CONFIDENCE_DPI);
+  }
```
0xA50C1A1, line 66: @IvanNardi @utoni what do you think about that? Could it be optimized in some way?

IvanNardi: `ndpi_isdigit` instead of `isascii`? [if we are looking for a number...]

IvanNardi (changes requested):
I think you did something like
```
git rm src/lib/protocols/redis_net.c
git add src/lib/protocols/resp.c
```
Could you try reverting and doing
```
git move src/lib/protocols/redis_net.c src/lib/protocols/resp.c
```
?
I prefer to keep the file history

0xA50C1A1:
> I think you did something like
> 
> ```
> git rm src/lib/protocols/redis_net.c
> git add src/lib/protocols/resp.c
> ```
> 
> Could you try reverting and doing
> 
> ```
> git move src/lib/protocols/redis_net.c src/lib/protocols/resp.c
> ```
> 
> ? I prefer to keep the file history

I've tried. For some reason redis_net.c is still considered as deleted.

PR #2292, opened 2024-01-30
Provide a u64 wrapper for `ndpi_set_config()`

utoni:
An application might want to set a numeric value as an integer e.g. if it was already parsed from a string previously.

### DIFF
--- a/src/lib/ndpi_main.c
+++ b/src/lib/ndpi_main.c
@@ -10825,6 +10825,21 @@ ndpi_cfg_error ndpi_set_config(struct ndpi_detection_module_struct *ndpi_str,
   return NDPI_CFG_NOT_FOUND;
 }
 
+ndpi_cfg_error ndpi_set_config_u64(struct ndpi_detection_module_struct *ndpi_str,
+                                   const char *proto, const char *param, uint64_t value)
+{
+  char value_str[21];
+  int value_len;
+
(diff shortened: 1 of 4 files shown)


src/lib/ndpi_main.c
```diff
@@ -10825,6 +10825,34 @@ ndpi_cfg_error ndpi_set_config(struct ndpi_detection_module_struct *ndpi_str,
...
+  int value_len;
+  const struct cfg_param *c;
+
+  value_len = ndpi_snprintf(value_str, sizeof(value_str), "%llu", (unsigned long long int)value);
+  if (value_len <= 0 || value_len >= (int)sizeof(value_str))
+  {
+    return NDPI_CFG_INVALID_PARAM;
+  }
+
+  for (c = &cfg_params[0]; c && c->param; c++)
```
IvanNardi: I think this check is incomplete; we need to check `proto` too.
I think the best solution is remove the `for` loop entirey and let `ndpi_set_config()` to do all the checks

utoni: Yea, you're right. My fear was that one might call `ndpi_set_config_u64` with a `param` which does not expect a numeric value. But that case is already captured by the callbacks.

IvanNardi (approved):
Not very fond of exposing this helper, but I *do* understand its usefulness!

PR #2293, opened 2024-01-30
GitHub Actions: update to latest macOS runners

IvanNardi:
Add a simple job with macos-14 on M1.
https://github.blog/changelog/2024-01-30-github-actions-introducing-the-new-m1-macos-runner-available-to-open-source/)
There are some issues with external dependencies (they are installed but
autoconf script doens;t find them) so keep it simple.

On macos-13 it seems that:
* there is no `realpath` program (even if coreutils has been
installed...)
* most of the filesystem is read only (we can't write on /usr/lib).

So I change
```
make install DESTDIR=$(realpath _install)
ls -alhHR _install
```
to
```
DESTDIR=/tmp/ndpi make install
ls -alhHR /tmp/ndpi
```
for all the jobs

Fix a warning on GitHub logs:
```
Node.js 16 actions are deprecated. Please update the following actions
to use Node.js 20: actions/checkout@v3. For more information see:
https://github.blog/changelog/2023-09-22-github-actions-transitioning-from-node-16-to-node-20/.
```

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -16,11 +16,11 @@ jobs:
     env:
       CFLAGS: -Werror
     steps:
-      - uses: actions/checkout@v3
+      - uses: actions/checkout@v4
         with:
           fetch-depth: 0
       - name: Set up Python 3.9
-        uses: actions/setup-python@v4
+        uses: actions/setup-python@v5
         with:
           python-version: "3.9"
       - name: Install Ubuntu Prerequisites
@@ -47,7 +47,7 @@ jobs:
     name: Test Utils (ubuntu-latest)
     runs-on: ubuntu-latest
     steps:
-      - uses: actions/checkout@v3
+      - uses: actions/checkout@v4
         with:
           fetch-depth: 0
       - name: Install Ubuntu Prerequisites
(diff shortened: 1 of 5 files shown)


.github/workflows/build.yml
```diff
@@ -150,19 +150,19 @@ jobs:
             msan: "--with-memory-sanitizer"
             nBPF: ""
           - compiler: "cc"
-            os: macOS-latest
+            os: macos-latest
```
utoni: Shouldn't it `macOS-latest`? Dunno if it is case-sensitive.

.github/workflows/build.yml
```diff
@@ -305,20 +305,26 @@ jobs:
...
       - name: Print nDPI long help
         if: startsWith(matrix.arch, 'x86_64') && !startsWith(matrix.os, 'windows')
         run: |
           cd ./example && ./ndpiReader -H
       - name: Install nDPI
         if: startsWith(matrix.arch, 'x86_64') && !startsWith(matrix.os, 'windows')
         run: |
-          make install DESTDIR=$(realpath _install)
-          ls -alhHR _install
+          DESTDIR=/tmp/ndpi make install
```
utoni, line 326: No `make install` ?

utoni: Nevermind, did not see it..

.github/workflows/build.yml
IvanNardi: GitHub ignores case when comparing strings: https://docs.github.com/en/actions/learn-github-actions/expressions
Nonetheless I will restore the original code; this change is useless

utoni (approved):

PR #2299, opened 2024-02-02
Normalization of host_server_name

vel21ripn:
The ndpi_hostname_sni_set() function replaces all non-printable characters with the "?" character and removing whitespace characters at the end of the line.

### DIFF
--- a/src/include/ndpi_private.h
+++ b/src/include/ndpi_private.h
@@ -393,7 +393,7 @@ void change_category(struct ndpi_detection_module_struct *ndpi_struct,
 		     ndpi_protocol_category_t protocol_category);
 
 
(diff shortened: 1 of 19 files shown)


IvanNardi:
We can't unconditionally check for invalid characters: the example (from the CI results) is mdns where pretty much all characters are valid...

Should we change the prototype to `char *ndpi_hostname_sni_set(struct ndpi_flow_struct *flow, const u_int8_t *value, size_t value_len, int normalize)` and do all these new checks only if `normalize == 1`? [I will say it should be always 1 but with mdns]

vel21ripn:
I added a few conditions to the server name normalization.
For the mdns protocol, only translation to lower case is performed.
For all other cases, full normalization is performed.

IvanNardi (approved):
LGTM, thank you very much

PR #2301, opened 2024-02-02
collectd: fix extraction of hostname

IvanNardi:


IvanNardi:
Noticed that while reviewing https://github.com/ntop/nDPI/pull/2299

utoni (approved):

PR #2310, opened 2024-02-08
Improve normalization of `flow->host_server_name`

IvanNardi:
Follow-up of 4543385d107fcc5a7e8632e35d9a60bcc40cb4f4

Remove trailing spaces for any HTTP header (we already remove leading spaces)

We want:
* a "normalized" string in `flow->host_server_name`, but
* to parse the original string for flow risk checking

`ndpi_hostname_sni_set()` is a private function, so there is no need to export its flags.


utoni (commented):
What about `ndpi_is_valid_hostname` in QUIC and FastCGI?
Don't they need any changes?

IvanNardi:
> What about `ndpi_is_valid_hostname` in QUIC and FastCGI? Don't they need any changes?

Thanks!! Done in last revision

utoni (approved):

PR #2312, opened 2024-02-09
Add PE32/PE32+ risk detection (detect transmitted windows executables).

utoni:

Describe changes:

The risk detection is far from perfect e.g. the amount of packets it does look for is hard coded. I think it might be better to let the user decide how many packets should checked for a possible PE transmission. It also assumes that the first 0x3C + 4 bytes are transmitted as one packet e.g. w/o fragmentation/retransmission.


0xA50C1A1:
I think it makes sense to also add HTTP content-type check for `application/vnd.microsoft.portable-executable`.

utoni:
I think there is already such check in `src/lib/protocols/http.c`. But it may be useful to check the HTTP body for PE32/PE32+. What do you think?

0xA50C1A1:
> I think there is already such check in `src/lib/protocols/http.c`. But it may be useful to check the HTTP body for PE32/PE32+. What do you think?

Yeah, that's a great idea.

0xA50C1A1:
> I think there is already such check in `src/lib/protocols/http.c`.

Yes, it's there, but I meant something else: to classify flows with that content-type as `Portable_Executable`.

utoni:
> > I think there is already such check in `src/lib/protocols/http.c`.
> 
> Yes, it's there, but I meant something else: to classify flows with that content-type as `Portable_Executable`.

Yea, makes also sense. I'll do both.

IvanNardi:
I'll do a proper review later, but I have an initial question: is there any specific reasons why this PE stuff is treated as a new protocol instead of a new flow risk?
[It seems that the former allows a simpler code for the generic UDP/TCP case]
I might have a preference for the flow risk solution (PE seems a specific case of "binary transfer risk"] but I don't have strong opinions and I am fine with both solutions.  I would like to know your reasoning about that...

utoni:
> I'll do a proper review later, but I have an initial question: is there any specific reasons why this PE stuff is treated as a new protocol instead of a new flow risk? [It seems that the former allows a simpler code for the generic UDP/TCP case] I might have a preference for the flow risk solution (PE seems a specific case of "binary transfer risk"] but I don't have strong opinions and I am fine with both solutions. I would like to know your reasoning about that...

Initially, I did not want to create a new protocol id for that.
But: The PCAP file shows a plaintext executable xfer, which was not classified before.
At first I've just wanted to write a protocol dissector that does not classify a protocol, but set's a risk.
It seems that there is a limitation in the core which prevents to do such things. So I've decided to create a new protocol id for it.
But I am open to revert that new protocol id creation if we can at least set the risk if a PE32/PE32+ executable is transmitted.

IvanNardi:
> > I'll do a proper review later, but I have an initial question: is there any specific reasons why this PE stuff is treated as a new protocol instead of a new flow risk? [It seems that the former allows a simpler code for the generic UDP/TCP case] I might have a preference for the flow risk solution (PE seems a specific case of "binary transfer risk"] but I don't have strong opinions and I am fine with both solutions. I would like to know your reasoning about that...
> 
> Initially, I did not want to create a new protocol id for that. But: The PCAP file shows a plaintext executable xfer, which was not classified before. At first I've just wanted to write a protocol dissector that does not classify a protocol, but set's a risk. It seems that there is a limitation in the core which prevents to do such things. So I've decided to create a new protocol id for it. But I am open to revert that new protocol id creation if we can at least set the risk if a PE32/PE32+ executable is transmitted.

Just an idea. Your current code basically checks for PE transfer only for unknown traffic (because when/if the flow is classified, PE dissector is not called anymore). [let's ignore the HTTP case which seems easy to handle]
If we want that, we could remove the new dissector/protocol and add something like that

```
static ndpi_protocol ndpi_internal_detection_process_packet(...) {
[...]
  /* At the end of this function */
  if(ret.app_protocol == NDPI_PROTOCOL_UNKNOWN) {
    /* just the code that is right now in ndpi_search_portable_executable and if we found it we set a risk */
  }
}
```
This way, we don't need a new protocol and  we can set a risk.
Just an idea

utoni:
@IvanNardi 
Again, I forgot a PR.
Need some kind of stale PR alert or so..

src/lib/ndpi_main.c
```diff
@@ -8160,6 +8160,43 @@ static int ndpi_is_ntop_protocol(ndpi_protocol *ret) {
 
 /* ********************************************************************************* */
 
+static void ndpi_search_portable_executable(struct ndpi_detection_module_struct *ndpi_struct,
+                                            struct ndpi_flow_struct *flow)
+{
+  struct ndpi_packet_struct const * const packet = &ndpi_struct->packet;
+  static const uint16_t dos_signature = 0x4d5a; /* MZ */
+  static const uint32_t pe_signature = 0x50450000; /* PE */
+
```
IvanNardi, line 8170: Could you add here a link to the PE format (or to some page talking about it)? Just as a reference

src/lib/protocols/http.c
```diff
@@ -193,10 +193,10 @@ static void ndpi_validate_http_content(struct ndpi_detection_module_struct *ndpi
...
 
-      if(len >= 8 /* 4 chars for \r\n\r\n and at least 4 charts for content guess */) {
-	double_ret += 4;
-
-	ndpi_http_check_human_redeable_content(ndpi_struct, flow, double_ret, len);
+      if (len > 4 /* 4 chars for \r\n\r\n and at least 4 charts for content guess */) {
+        double_ret += 4;
+        len -= 4;
+        ndpi_http_check_human_redeable_content(ndpi_struct, flow, double_ret, len);
       }
```
IvanNardi: Is this change still related to the PE? If so, I think we should update the comment, at very least....

IvanNardi (changes requested):

src/lib/protocols/http.c
utoni: Actually not anymore. I guess it can be reverted.

IvanNardi (approved):
Thank you

PR #2321, opened 2024-02-19
Add a script to download/update the domain suffix list

IvanNardi:


utoni (commented):
What is the actual idea of those DNS suffixes for public institutions?

It is not used for classification, right?

IvanNardi:
> What is the actual idea of those DNS suffixes for public institutions?
> 
> It is not used for classification, right?

AFAIK, yes. Looking at the code, it seems to me it is used only for DGA detection (`ndpi_check_dga_name()` -> `ndpi_get_host_domain()`)

utoni (approved):

Issue #2258, opened 2024-01-12
Memory leak in ahocorasick


IvanNardi:
Oss-fuzz keeps reporting a memory leak in ahocorasick code, via `fuzz_filecfg_protocols` fuzzer.
Some examples (these reports should be public):
https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=64150&q=ndpi&can=1&sort=-id
https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=62269&q=ndpi&can=1&sort=-id
https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=61934&q=ndpi&can=1&sort=-id

It **seems** that the leak is about inserting duplicated patterns.

The stack reported is something like:
```
            #6 0x67f7c9 in ac_automata_add [ndpi/src/lib/third_party/src/ahocorasick.c:255](https://github.com/ntop/nDPI/blob/7b2bbb2309264766697507365231fd6ee5717e31/src/lib/third_party/src/ahocorasick.c#L255):19
	    #7 0x58df28 in ndpi_add_host_risk_mask [ndpi/src/lib/ndpi_main.c:4262](https://github.com/ntop/nDPI/blob/7b2bbb2309264766697507365231fd6ee5717e31/src/lib/ndpi_main.c#L4262):8
	    #8 0x592605 in ndpi_handle_rule [ndpi/src/lib/ndpi_main.c:4344](https://github.com/ntop/nDPI/blob/7b2bbb2309264766697507365231fd6ee5717e31/src/lib/ndpi_main.c#L4344):11
	    #9 0x592605 in load_protocols_file_fd [ndpi/src/lib/ndpi_main.c:5036](https://github.com/ntop/nDPI/blob/7b2bbb2309264766697507365231fd6ee5717e31/src/lib/ndpi_main.c#L5036):8
```

mmanoj:
@IvanNardi 

Is this issue still valid?

IvanNardi:
Yes, according to oss-fuzz. If you are interested I can provide an up-to-date artifact to reproduce the error

mmanoj:
@IvanNardi 

Thanks for the feedback, lets workout this issue as well together as we done same way of FPC feature.This way we can achieve more good results.

mmanoj:
@IvanNardi 
Can we work on this?

IvanNardi:
[...]

PR #2331, opened 2024-02-27
ahocorasick: improve matching with subdomains

### LINKED ISSUE #2330: ahocorasick: wrong match
In some case, we have wrong match returned by ahocorasick: "t.me" pattern triggers a match for "tt.me"

Context:
* we have the following line to match Telegram traffic
```
{ "t.me", "Telegram", NDPI_PROTOCOL_TELEGRAM, NDPI_PROTOCOL_CATEGORY_CHAT, NDPI_PROTOCOL_ACCEPTABLE, NDPI_PROTOCOL_DEFAULT_LEVEL },
```
* "t.me" and subdomains are correctly classified as Telegram
* "tt.me" is wrongly classified as Telegram

You can reproduce the issue with `ndpiReader -x DOMAIN_NAME`:

```
ivan@ivan-Latitude-E6540:~/svnrepos/nDPI(reader_check_domain)$ ./example/ndpiReader -x "t.me"
Match Found for string [t.me] -> P(185) B(1) C(9) => Telegram Acceptable Chat                         <---- OK
ivan@ivan-Latitude-E6540:~/svnrepos/nDPI(reader_check_domain)$ ./example/ndpiReader -x "foo.t.me"
Match Found for string [foo.t.me] -> P(185) B(1) C(9) => Telegram Acceptable Chat                  <---- OK
ivan@ivan-Latitude-E6540:~/svnrepos/nDPI(reader_check_domain)$ ./example/ndpiReader -x "tt.me"
Match Found for string [tt.me] -> P(185) B(1) C(9) => Telegram Acceptable Chat                        <---- WRONG

```

A comment in `ac_domain_match_handler()` in `src/lib/ndpi_main.c` seems about that issue:
```
      /*
	The patch below allows in case of pattern ws.amazon.com
	to avoid matching aws.amazon.com whereas a.ws.amazon.com
	has to match
      */
```

An example of traffic with "tt.me" pattern:  [tt.me.zip](https://github.com/ntop/nDPI/files/14409853/tt.me.zip)

IvanNardi:
The basic idea is to have the following logic:
* pattern "DOMAIN" matches the domain itself (i.e exact match) *and* any subdomains (i.e. "ANYTHING.DOMAIN")
* pattern "DOMAIN." matches *also* any strings for which is a prefix [please, note that this kind of match is handy but it is quite dangerous...]
* pattern "-DOMAIN" matches *also* any strings for which is a postfix

Examples:
* pattern "wikipedia.it":
  * "wikipiedia.it" -> OK
  * "foo.wikipedia.it -> OK
  * "foowikipedia.it -> NO MATCH
  * "wikipedia.it.com -> NO MATCH
* pattern "wikipedia.":
  * "wikipedia.it" -> OK
  * "foo.wikipedia.it -> OK
  * "foowikipedia.it -> NO MATCH
  * "wikipedia.it.com -> OK
* pattern "-wikipedia.it":
  * "wikipedia.it" -> NO MATCH
  * "foo.wikipedia.it -> NO MATCH
  * "0001-wikipedia.it -> OK
  * "foo.0001-wikipedia.it -> OK

Bottom line:
* exact match
* prefix with "." (always, implicit)
* prefix with "-" (only if explicitly set)
* postfix with "." (only if explicitly set)

That means that the patterns cannot start with '.' anymore.

Close #2330


lucaderi:
I am good with the patch even though I think that it should be better to just patch the engine and not the individual protocols. When ready we can merge it

IvanNardi:
Latest version changes the code only in the callback

IvanNardi:
The error reported by the fuzzing job is unrelated to this change: it is exactly the bug reported in #2258

PR #2332, opened 2024-02-27
Add BFD protocol dissector

0xA50C1A1:


src/lib/ndpi_main.c
```diff
@@ -2246,6 +2246,10 @@ static void ndpi_init_protocol_defaults(struct ndpi_detection_module_struct *ndp
 			  "DLEP", NDPI_PROTOCOL_CATEGORY_NETWORK,
 			  ndpi_build_default_ports(ports_a, 854, 0, 0, 0, 0) /* TCP */,
 			  ndpi_build_default_ports(ports_b, 854, 0, 0, 0, 0) /* UDP */);
+  ndpi_set_proto_defaults(ndpi_str, 1 /* cleartext */, 0 /* nw proto */, NDPI_PROTOCOL_ACCEPTABLE, NDPI_PROTOCOL_BFD,
+			  "BFD", NDPI_PROTOCOL_CATEGORY_NETWORK,
+			  ndpi_build_default_ports(ports_a, 0, 0, 0, 0, 0) /* TCP */,
+			  ndpi_build_default_ports(ports_b, 3784, 3785, 0, 0, 0) /* UDP */);
```
0xA50C1A1, line 2252: @IvanNardi Is it worth keeping these ports here? They're IANA registered, but can often be used by some other software. That's why I added port checking in the dissector code.

IvanNardi: Since these ports are IANA registered, I think we can keep them (even if their usefulness is quite limited for this kind of protocols, IMO)

IvanNardi (approved):
Thanks

PR #2336, opened 2024-03-05
STUN: add a parameter to configure how long the extra dissection lasts

IvanNardi:
Tradeoff: performance (i.e. number of packets) vs sub-classification


utoni (approved):

doc/configuration_parameters.md
```diff
@@ -31,6 +31,7 @@ TODO
 | "imap"       | "tls_dissection"                          | enable        | NULL      | NULL      | Enable/disable dissection of TLS packets in cleartext IMAP flows (because of opportunistic TLS, via STARTTLS msg) |
 | "pop"        | "tls_dissection"                          | enable        | NULL      | NULL      | Enable/disable dissection of TLS packets in cleartext POP flows (because of opportunistic TLS, via STARTTLS msg) |
 | "ftp"        | "tls_dissection"                          | enable        | NULL      | NULL      | Enable/disable dissection of TLS packets in cleartext FTP flows (because of opportunistic TLS, via AUTH TLS msg) |
+| "stun"       | "max_packets_extra_dissection"            | 4             | 0         | 255       | After a flow has been classified has STUN, nDPI might analyse more packets to look for a sub-classification or for metadata. This parameter set the upper limit on the number of these packets  |
```
utoni, line 34: Why 4 instead of 10 (as before)?

// The default is 14, right?
// Sorry, need my first coffee...

IvanNardi: Before the default was 4, with the exception of Telegram where it was 10.
Since I noticed that 4 or 10 doesn't change anything for Telegram in the unit tests, I kept 4 for everything, to keep it simple.

In case we find DTLS, we add 10 (fixed value, not configurable) more packets to the current value: the idea is that if there is DTLS, we might need more packets because, other than dissect the "usual" STUN packets, we also need to parse the DTLS handshake (Client Hello and certificate)

PR #2341, opened 2024-03-07
Add a specific protocol id for audio/video calls made using Google apps

IvanNardi:
Same logic already used for Signal/Whatsapp/Line/Facebook/...


doc/protocols.rst
```diff
@@ -40,6 +40,7 @@ Redis Serialization Protocol
 
 References: `Protocol Specs: <https://redis.io/docs/reference/protocol-spec/`_
 
+
 .. _Proto 201:
 
 `NDPI_PROTOCOL_GOOGLE_MEET`
```
utoni, line 46: Isn't that id now obsolete?

IvanNardi: it is still used for the generic, no realtime,  TLS traffic (web site, generic app flows,...)

utoni (approved):

PR #2346, opened 2024-03-14
LRU cache: move to 64 bits long keys

IvanNardi:
Tradeoff between key comparison efficiency (i.e. no `memcmp`) and key length.
At least in the ipv4 cases, we have no more different entries with the same key.

### DIFF
--- a/src/include/ndpi_api.h
+++ b/src/include/ndpi_api.h
@@ -1058,9 +1058,9 @@ extern "C" {
   /* LRU cache */
   struct ndpi_lru_cache* ndpi_lru_cache_init(u_int32_t num_entries, u_int32_t ttl, int shared);
   void ndpi_lru_free_cache(struct ndpi_lru_cache *c);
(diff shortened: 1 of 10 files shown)


utoni (changes requested):
It's probably better to replace all `%lx` with `%llx`.
I think Windows might not like `%lx`, because it expects an `unsigned long`, which is 32 bit unlike on Unix/Linux.

IvanNardi:
> It's probably better to replace all `%lx` with `%llx`. I think Windows might not like `%lx`, because it expects an `unsigned long`, which is 32 bit unlike on Unix/Linux.

Something like the changes pushed in the last version?

utoni (approved):
Exactly! :)

IvanNardi:
The failure on the fuzzing job is unrelated to this PR and seems related to the random failures on sanitizer jobs with recent github runner images (see 25c145f26d3e67dc280d439eceb71a6b321b8c68)
Note that the workaround used in 25c145f26d3e67dc280d439eceb71a6b321b8c68 doesn't work here because we are directly using a docker image provided by oss-fuzz

PR #2356, opened 2024-03-22
Add LoL: Wild Rift detection

0xA50C1A1:

Describe changes:

Regarding the detection logic: a UDP session is establishing with the game match start, and the first two packets in this session always have payload lengths of 69 and 359 bytes and contain signatures 0x4000000 (request) and 0x10000000 (response).


0xA50C1A1:
```
../include/ndpi_typedefs.h:1504:1: error: static assertion failed: "Size of the flow struct increased to more than 1008 bytes, please check if this change is necessary."
```

@IvanNardi @utoni 

Hmmm... Maybe it's somehow possible to add this stuff to the existing Riot Games dissector to avoid creating a new protocol ID.

Btw, here's a pcap sample with the traffic of the entire match if you're interested: [lol_wild_rift.zip](https://github.com/ntop/nDPI/files/14723994/lol_wild_rift.zip)

src/include/ndpi_typedefs.h
```diff
@@ -958,6 +958,9 @@ struct ndpi_flow_udp_struct {
   /* NDPI_PROTOCOL_MUMBLE */
   u_int8_t mumble_stage:1;
   u_int64_t mumble_ident;
+
+  /* NDPI_PROTOCOL_LOLWILDRIFT */
+  u_int8_t lolwildrift_stage:1;
```
IvanNardi: Try moving it up, as `u_int32_t lolwildrift_stage:1` (note the 32, not 8...)  just before the epic games stuff.
This way the size of the structure shouldn't change

0xA50C1A1: > Try moving it up, as `u_int32_t lolwildrift_stage:1` (note the 32, not 8...) just before the epic games stuff. This way the size of the structure shouldn't change

Ty, seems that it works.

IvanNardi:
> ```
> ../include/ndpi_typedefs.h:1504:1: error: static assertion failed: "Size of the flow struct increased to more than 1008 bytes, please check if this change is necessary."
> ```
> 
> @IvanNardi @utoni
> 
> Hmmm... Maybe it's somehow possible to add this stuff to the existing Riot Games dissector to avoid creating a new protocol ID.
> 
> Btw, here's a pcap sample with the traffic of the entire match if you're interested: [lol_wild_rift.zip](https://github.com/ntop/nDPI/files/14723994/lol_wild_rift.zip)

Sooner or later we need to decide if there is any value in having one protocol id for each games, or if we might switch to something different... (one id for all the games  + metadata with specific game indication? one specific id for each "sw house/company"? specific ids only for the most "famous" games?...)

IvanNardi (approved):

0xA50C1A1:
> Sooner or later we need to decide if there is any value in having one protocol id for each games, or if we might switch to something different... (one id for all the games + metadata with specific game indication? one specific id for each "sw house/company"? specific ids only for the most "famous" games?...)

Well, many games use either out-of-the-box stuff like RakNet or Steam Datagram Relay or some generic protocols (NetEase Games, Gaijin Entertainment, Tencent Games) for all games from a publisher/developer, so I don't see a problem with leaving a few IDs for something that stands out.

PR #2360, opened 2024-04-02
Fixed gcc version checking.

vel21ripn:
Using the "test" utility instead of "[" to determine the gcc compiler version.


utoni (approved):
That would have been my next topic. Thanks for fixing that one (and being faster then me).

PR #2361, opened 2024-04-03
STUN: rework sub-classification

IvanNardi:
The main goal is to have the "real" application (if any; i.e. Signal/Whatsapp/Telegram/...) always as "application" protocol and not as "master" one


src/lib/protocols/stun.c
```diff
@@ -293,7 +351,8 @@ int is_stun(struct ndpi_detection_module_struct *ndpi_struct,
                        flow->detected_protocol_stack[0]);
 
           if(1 /* TODO: enable/disable */ &&
```
utoni: Is that still needed?
Was the idea to make this configurable?

IvanNardi: It was.. at the time... I'll remove the comment at the next iteration of this patch

IvanNardi: Done

utoni (approved):

PR #2362, opened 2024-04-03
Patrica tree: fix handling of mac addresses

IvanNardi:
Found looking at the fuzzer statistics...

### DIFF
--- a/src/lib/third_party/src/ndpi_patricia.c
+++ b/src/lib/third_party/src/ndpi_patricia.c
@@ -238,6 +238,16 @@ static ndpi_prefix_t * ndpi_New_Prefix2 (int family, void *dest, int bitlen, ndp
       }
       memcpy (&prefix->add.sin, dest, sizeof(struct in_addr));
     }
+    else if(family == AF_MAC) {
+      default_bitlen = 48;
+      if(prefix == NULL) {
+        prefix = (ndpi_prefix_t*)ndpi_calloc(1, sizeof (ndpi_prefix_t));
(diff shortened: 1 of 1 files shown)


src/lib/third_party/src/ndpi_patricia.c
```diff
@@ -238,6 +238,16 @@ static ndpi_prefix_t * ndpi_New_Prefix2 (int family, void *dest, int bitlen, ndp
       }
       memcpy (&prefix->add.sin, dest, sizeof(struct in_addr));
     }
+    else if(family == AF_MAC) {
```
utoni, line 241: ~~Is it an apple exclusive address family or~~ does it refer to Ethernet?
And what's the purpose of that AF?

P.S.: I am missing my nth cup of coffee...

utoni (approved):

src/lib/third_party/src/ndpi_patricia.c
IvanNardi: I *think* that the goal is to have MAC addresses into the tree, in the same way we have ipv4 or ipv6 addresses
The constant `AF_MAC` is added for "compatibility" with `AF_INET` and `AF_INET6`, so that you can specify the tree "type" in the same way.
Note that I am guessing here... never use mac addresses and it seems `AF_MAC` is in nDPI pretty much since forever

utoni: ty for clarification <3

PR #2369, opened 2024-04-05
Calculate packet entropy for unknown protocols.

utoni:

Describe changes:

Maybe the max analyze packets should be configurable as I did not investigate any possible performance impact.


src/lib/ndpi_main.c
```diff
@@ -8546,13 +8546,27 @@ static ndpi_protocol ndpi_internal_detection_process_packet(struct ndpi_detectio
...
+  {
+    flow->entropy = ndpi_entropy(packet->payload, packet->payload_packet_len);
+    if(NDPI_ENTROPY_ENCRYPTED_OR_RANDOM(flow->entropy) != 0) {
+      char str[32];
+
+      snprintf(str, sizeof(str), "Entropy %.2f", flow->entropy);
+      ndpi_set_risk(flow, NDPI_SUSPICIOUS_ENTROPY, str);
+    }
+  }
+
```
IvanNardi, line 8618: We end up having an entropy value even for classified traffic, as reported in the unit tests diff (trivial example: entropy calculated in the first packet, but flow classified only on the second pkt)

Is that what we want?

BTW: I am right saying that, if we have that value, it is always calculated on the first packet of the flow?

utoni: Can you point me the diff test, because all I've found were Matches by IP/Port. I guess it's fine to calculate the entropy for all !DPI confidence flows.

IvanNardi: tests/cfgs/default/result/dnscrypt-v1-and-resolver-pings.pcap.out

I agree; no issues when the classification is !DPI

IvanNardi (approved):

PR #2370, opened 2024-04-05
STUN: improve extraction of Mapped-Address metadata

IvanNardi:
Enable parsing of Mapped-Address attribute for all STUN flows: that means that STUN classification might require more packets.

Add a configuration knob to enable/disable this feature.

Note that we can have (any) STUN metadata also for flows *not* classified as STUN (because of DTLS).

Add support for ipv6.

Restore the correct extra dissection logic for Telegram flows.


example/ndpiReader.c
utoni, line 1: Why not remove `intoaV4()` ?

IvanNardi: Done

utoni (approved):

PR #2375, opened 2024-04-07
Add Shellscript risk detection.

utoni:

- [X] I have read the contributing guide lines at https://github.com/ntop/nDPI/blob/dev/CONTRIBUTING.md
- [X] I have updated the documentation (in doc/) to reflect the changes made (if applicable)

I am not sure about the FTP Data detection. Sounds like a false positive to me.
Also the risk is not that accurate, but setting binary application xfer would also not be accurate.


IvanNardi:
The FTPData dissector identifies FTP data flows simply looking for some "magic words" at the beginning of the flow, to identify some specific file types. The basic idea is: if the payload is some known file type -> this is a FTP transfer.
Let avoid discussing now if that makes (still) sense... [spoiler: it doesn't...]

Since the Bash script is among the pattern that the FTPData dissector is looking for, I was not surprised that one of your flows matches it; I was surprised that some of them didn't' match!!! [The reason is: FTPData looks only for TCP and for "#!/b" pattern; your second TCP flow begin with "#! /", with a space...]

IvanNardi (approved):

IvanNardi:
Just to be clear: I am fine with the patch "as-is". But I am also fine if you decide to remove lines 121-123 from ftp_data.c...

utoni:
I think I'll remove those lines from the FTP dissector. It does not really makes sense to me.

IvanNardi:
@utoni, could you rebase, please?

IvanNardi (approved):
Thanks

PR #2376, opened 2024-04-08
Add The Elder Scrolls Online support

0xA50C1A1:

Describe changes:

In favor of The Elder Scrolls Online's 10th anniversary, which it had 4 days ago. This MMORPG is still popular even nowadays.

I'll leave it as a draft for now, because I'm not sure about using ```memmem()```. Although this is a GNU extension from glibc, this function is available in BSD, macOS and musl libc, also it's works pretty fast. But I'm not sure about the MSVC has this function, also it's not in the Linux kernel space, which will cause some issues for @vel21ripn's fork.


IvanNardi:
I you feel brave, you might try to add something like "ndpi_memmem" in `ndpi_utils.c` based on the implementation that you find in `src/lib/third_party/src/libinjection_sqli.c`....

0xA50C1A1:
> I you feel brave, you might try to add something like "ndpi_memmem" in `ndpi_utils.c` based on the implementation that you find in `src/lib/third_party/src/libinjection_sqli.c`....

Yea, it's a good idea and I thought about it, but the problem is that many simple copypasteable implementations have O(n^2) complexity. Well, if use it as in my case (just to find some sequence in the first 100 bytes of payload), then the complexity doesn't matter, but it's not good for something more serious.

0xA50C1A1:
https://github.com/skeeto/scratch/blob/d366cfc9d70b9bf53a432e21a6ce6efeeebecf80/misc/rabin-karp.c#L57-L97

Here is a bit better implementation published under [Unlicense](https://en.wikipedia.org/wiki/Unlicense), so I guess we can take it.

IvanNardi:
@0xA50C1A1, could you add the usual entry into `doc/protocols.rst`, please?

src/lib/protocols/teso.c
```diff
@@ -0,0 +1,91 @@
...
+
+  NDPI_EXCLUDE_PROTO(ndpi_struct, flow);
+}
+
+void init_teso_dissector(struct ndpi_detection_module_struct *ndpi_struct, u_int32_t *id)
+{
+  ndpi_set_bitmask_protocol_detection("TES_Online", ndpi_struct, *id,
+              NDPI_PROTOCOL_TESO,
+              ndpi_search_teso,
+              NDPI_SELECTION_BITMASK_PROTOCOL_TCP_WITH_PAYLOAD_WITHOUT_RETRANSMISSION,
```
IvanNardi: `NDPI_SELECTION_BITMASK_PROTOCOL_V4_V6_TCP_WITH_PAYLOAD_WITHOUT_RETRANSMISSION`

0xA50C1A1:
> @0xA50C1A1, could you add the usual entry into `doc/protocols.rst`, please?

Yea. For some reason I thought that I had already done it.

src/lib/protocols/teso.c
0xA50C1A1: > `NDPI_SELECTION_BITMASK_PROTOCOL_V4_V6_TCP_WITH_PAYLOAD_WITHOUT_RETRANSMISSION`

This game doesn’t seem to use IPv6, but I guess nothing bad will happen if I change selection bitmask to V4_V6.

doc/protocols.rst
```diff
@@ -706,3 +706,21 @@ References: `RFC <https://datatracker.ietf.org/doc/html/rfc6726>`_
...
+
+
+.. _Proto 407:
+
+`NDPI_PROTOCOL_LOLWILDRIFT`
+============================
+League of Legends: Wild Rift is a mobile MOBA game.
+
+References: `Main site <https://wildrift.leagueoflegends.com/>`_
+
```
IvanNardi, line 718: Too much :)

IvanNardi: Too much :)

src/lib/protocols/teso.c
IvanNardi: Regardless of this specific protocol, generally speaking, it is better to handle ipv6 traffic too, even if right now the server/service is only ipv4, because:
* it might support ipv6 in the future
* we might capture the traffic before a NAT64 translation

IvanNardi (approved):
Thank you

IvanNardi:
The error in the fuzzing is always #2258...

PR #2378, opened 2024-04-09
Add memmem() implementation

0xA50C1A1:

Describe changes:

Although this implementation has O(n*m) complexity in the worst case, in the typical nDPI scenario (searching for a string or byte sequence in a payload with a typical maximum length of 1460/1472 bytes) it performed much better than the other implementations.

Here is the result of a small benchmark searching for the string ```eso.live``` in an 852 byte long array (packet from The Elder Scrolls Online pcap sample):

image

### DIFF
--- a/src/lib/ndpi_main.c
+++ b/src/lib/ndpi_main.c
@@ -11442,9 +11442,8 @@ void* ndpi_memmem(const void* haystack, size_t haystack_len, const void* needle,
     return memchr(haystack, *(int*)needle, haystack_len);
   }
 
-  for (const u_int8_t* h = haystack; h <= (const u_int8_t*)haystack + 
-       haystack_len - needle_len; ++h)
-  {
+  const u_int8_t* h = NULL;
+  for (h = haystack; h <= (const u_int8_t*)haystack+haystack_len-needle_len; ++h) {
     if (!memcmp(h, needle, needle_len))
       return (void*)h;
   }
--- a/src/lib/ndpi_main.c
+++ b/src/lib/ndpi_main.c
@@ -11443,8 +11443,9 @@ void* ndpi_memmem(const void* haystack, size_t haystack_len, const void* needle,
   }
 
   const u_int8_t* h = NULL;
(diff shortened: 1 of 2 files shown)


0xA50C1A1:
Uh, I don't think it's related to my PR:

```
protocols/http.c:68:80: error: unused parameter ‘ndpi_struct’ [-Werror=unused-parameter]
   68 | static void ndpi_set_binary_data_transfer(struct ndpi_detection_module_struct *ndpi_struct,
      |      
```

utoni:
The error was introduced by a recent commit to dev. ;)

IvanNardi:
@0xA50C1A1 , could you rebase, please?

0xA50C1A1:
> @0xA50C1A1 , could you rebase, please?

Done.

src/lib/ndpi_main.c
```diff
@@ -11423,3 +11423,30 @@ char *ndpi_dump_config(struct ndpi_detection_module_struct *ndpi_str,
...
+    return NULL;
+  }
+
+  if (needle_len == 1) {
+    return memchr(haystack, *(int*)needle, haystack_len);
+  }
+
+  const u_int8_t* h = NULL;
+  for (h = haystack; h <= (const u_int8_t*)haystack+haystack_len-needle_len; ++h) {
+    if (!memcmp(h, needle, needle_len))
```
IvanNardi: Something like (not tested!)?
```
if(*h == needle[0] && !memcmp(h, needle, needle_len))
```
To avoid too many memcmp

0xA50C1A1: I guess it could be added, but I didn't see any noticeable change. Maybe my benchmark is wrong, or maybe the difference is less than 1 ns.

[memmem_bench.zip](https://github.com/ntop/nDPI/files/14931089/memmem_bench.zip)

screenshot1
screenshot2

IvanNardi (approved):
I think this implementation if fine for our usercase(s).
If needed, we can always improved it later

IvanNardi:
Thank you

PR #2392, opened 2024-04-13
Add strlcpy implementation

0xA50C1A1:

Describe changes:

I'll leave it as a draft to debate whether this stuff is even needed in nDPI... Well, I guess it would be useful.

`strlcpy` is a safe alternative to `strncpy`, originally added to BSD libc (also recently added to glibc). The main advantage of `strlcpy` over `strncpy` is that it always null-terminates the destination buffer, preventing issues that can occur with `strncpy` when the source string is longer than the specified buffer size.

Ofc, `strlcpy` isn't suitable for cases when it's necessary to copy a string without a null-terminator, such as when dealing with fixed-size data structures or binary data... But who uses `strncpy` instead of `memcpy` in such cases?

@IvanNardi @utoni what do you think?

### DIFF
--- a/example/ndpiReader.c
+++ b/example/ndpiReader.c
@@ -5630,6 +5630,42 @@ void strtonumUnitTest() {
 
 /* *********************************************** */
 
+void strlcpyUnitTest() {
+  // Test empty string
+  char dst_empty[10] = "";
+  assert(ndpi_strlcpy(dst_empty, "", sizeof(dst_empty)) == 0);
+  assert(dst_empty[0] == '\0');
+
+  // Basic copy test
+  char dst1[10] = "";
+  assert(ndpi_strlcpy(dst1, "abc", sizeof(dst1)) == 3);
+  assert(strcmp(dst1, "abc") == 0);
+
+  // Test with dst_len smaller than src_len
+  char dst2[4] = "";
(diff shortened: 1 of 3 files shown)


utoni:
A good idea in general. If you have some spare time, do you want to replace all the `strncpy()`s? :D

0xA50C1A1:
> A good idea in general. If you have some spare time, do you want to replace all the `strncpy()`s? :D

Yea, sure. I guess `sed` will suffice, since the arguments of `strncpy` and `strlcpy` are the same. But something tells me that things might go wrong :D

IvanNardi:
@0xA50C1A1, could you add a few unit tests, please? Take a look at `example/ndpiReader.c` around line 6550 to similar tests
Nothing fancy: testing empty string and input buffer equal to output buffer or smaller/grater of 1 byte.
You can use `assert` as in the other unit tests.

IvanNardi:
> that things might go wrong :D

VERY wrong... :D

0xA50C1A1:
> @0xA50C1A1, could you add a few unit tests, please? Take a look at `example/ndpiReader.c` around line 6550 to similar tests Nothing fancy: testing empty string and input buffer equal to output buffer or smaller/grater of 1 byte. You can use `assert` as in the other unit tests.

Done.

PR #2397, opened 2024-04-18
Add KNXnet/IP protocol support

41CEA55:

Describe changes:

KNXnet/IP is a communication protocol used in building automation and control systems, it allows devices connected to a KNX network to communicate over an Ethernet/IP-based network.

Unfortunately I have no examples of KNXnet/IP over TCP except this pcap with a single packet. And it's not properly detected because the `Total Length` field is not equal to the payload length.

[TcpSegmentation.zip](https://github.com/ntop/nDPI/files/15024332/TcpSegmentation.zip)


src/lib/protocols/knxnet_ip.c
```diff
@@ -0,0 +1,107 @@
...
+      (packet->payload[1] != 0x10))
+  {
+    NDPI_EXCLUDE_PROTO(ndpi_struct, flow);
+    return;
+  }
+
+  u_int16_t service_id = ntohs(get_u_int16_t(packet->payload, 2));
+  u_int16_t total_length = ntohs(get_u_int16_t(packet->payload, 4));
+
+  if (is_valid_knxnet_ip_service_id(service_id) && total_length == packet->payload_packet_len)
```
IvanNardi: Could we do something like, in pseudo-code:
```
if udp --> total_length == packet->payload_packet_len
else --> total_length <= packet->payload_packet_len
```
to try to handle also the TCP sample? If so, could you update the trace with the TCP flow, please?

IvanNardi (changes requested):

src/lib/protocols/knxnet_ip.c
41CEA55: > Could we do something like, in pseudo-code:
> 
> ```
> if udp --> total_length == packet->payload_packet_len
> else --> total_length <= packet->payload_packet_len
> ```
> 
> to try to handle also the TCP sample? If so, could you update the trace with the TCP flow, please?

Yeah, sure. I read the protocol specs and found that KNXnet/IP has the feature to pack multiple messages in a single TCP packet.

In this case we can use the `Total Length` field of the first message as an offset to check if there is another KNXnet/IP header (i.e. bytes `06 10` and another `Service ID`).

IvanNardi (approved):
Thanks for adding TCP support

PR #2401, opened 2024-04-19
Add BFCP protocol support

41CEA55:

Describe changes:

The Binary Floor Control Protocol (BFCP) is used in video conferencing systems to control the flow of shared content between participants. It allows users to manage who is presenting and viewing content during a meeting, ensuring smooth collaboration and communication.


41CEA55:
Unfortunately, I don't have any valid BFCP over TCP samples.

src/include/ndpi_typedefs.h
```diff
@@ -1496,6 +1496,10 @@ struct ndpi_flow_struct {
   /* NDPI_PROTOCOL_TINC */
   u_int8_t tinc_state;
 
+  /* NDPI_PROTOCOL_BFCP */
+  u_int8_t bfcp_stage;
+  u_int32_t bfcp_conference_id;
+
```
IvanNardi: Move above OPENVPN and use only one bit for `bfcp_stage`

src/lib/protocols/bfcp.c
```diff
@@ -0,0 +1,85 @@
...
+  }
+
+  u_int32_t conference_id = ntohl(get_u_int32_t(packet->payload, 4));
+  if (!flow->bfcp_stage) {
+    flow->bfcp_conference_id = conference_id;
+    flow->bfcp_stage = 1;
+    return;
+  }
+
+  if (flow->bfcp_conference_id == conference_id) {
```
IvanNardi: I think that  a `&& flow->bfcp_stage == 1` is missing here. Is that possible?

IvanNardi (changes requested):

src/lib/protocols/bfcp.c
41CEA55: > I think that a `&& flow->bfcp_stage == 1` is missing here. Is that possible?

It's possible, but it won't make much difference, since there is already a check `!flow->bfcp_stage` above with `return`.

41CEA55:
@IvanNardi Thanks for the review, I've fixed those issues.

src/lib/protocols/bfcp.c
IvanNardi: You are right! Sorry for the noise

IvanNardi:
@41CEA55, do you know if this protocol is really used over TCP? If not, I slightly prefer to support UDP only....

41CEA55:
> @41CEA55, do you know if this protocol is really used over TCP? If not, I slightly prefer to support UDP only....

According to RFC 8855, this protocol can use both TCP and UDP, but usually TCP is used in rare cases, for example if a UDP session couldn't be established.

41CEA55:
There are some TCP flows, but the problem is that they are unidirectional because they just consist only 1 packet.

[bfcp_test_packets.zip](https://github.com/ntop/nDPI/files/15073642/bfcp_test_packets.zip)

IvanNardi:
@41CEA55 , thank you.
Could you add these tcp packets to `bfcp.pcapng` (even if they are not classified), please?

41CEA55:
> @41CEA55 , thank you. Could you add these tcp packets to `bfcp.pcapng` (even if they are not classified), please?

Well, I got something better: this pcap file had a valid `FloorRequestQuery` and `FloorRequestStatus`, so I could simulate a session.

IvanNardi (approved):
Thank you very much!

PR #2406, opened 2024-04-23
DTLS: add support for Alert message type (similar to TLS)

IvanNardi:


utoni (approved):
Good catch, exactly what I need!

PR #2411, opened 2024-04-28
Workaround for fixing GitHub runners on macOS

IvanNardi:
GitHub switched "macos-latest" from "macos-12" to "macos-14", which is only on ARM64!
https://github.com/actions/runner/issues/3256
https://github.blog/changelog/2024-01-30-github-actions-macos-14-sonoma-is-now-available/

However we are having some issues build nDPI on macos-14 with external libraries:

```
configure: error: libgpg-error required (because of --with-local-libgcrypt) but not found or too old.
```
See: https://github.com/ntop/nDPI/actions/runs/8869020568/job/24350356867

```
ndpi_utils.c:69:10: fatal error: 'pcre2.h' file not found
         ^~~~~~~~~
1 error generated.
```
See: https://github.com/ntop/nDPI/actions/runs/8869020568/job/24349242251

Everything is still fine with macos-14 and no external dependencies

As workaround, test only macos-12 and macos-13 in our main matrix.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -76,8 +76,9 @@ jobs:
       fail-fast: true
       matrix:
         # ubuntu-latest == ubuntu-22.04
-        # macOS-latest == macos-12
-        os: ["ubuntu-latest", "ubuntu-20.04", "macOS-latest", "macos-13", "windows-latest"]
+        # macOS-latest == macos-14 on **ARM64**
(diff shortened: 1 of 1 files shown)


IvanNardi:
I had the same kind of issue when I added the first basic support for macos-14 on arm64: https://github.com/ntop/nDPI/commit/8fbef7fb5ab682b7aa9c92e4fd6e106698eb4ae1

utoni (approved):
Thanks. I'll need that as well. :)

PR #2427, opened 2024-05-08
Raknet/RTP: avoid Raknet false positives and harden RTP heuristic

IvanNardi:
There is some overlap between RTP and Raknet detection: give precedence to RTP logic.
Consequences:
* Raknet might require a little bit more packets for some flows (not a big issue)
* some very small (1-2 pkts) Raknet flows are not classified (not sure what do do about that..)


IvanNardi:
Without this patch, the 2 flows in the new trace are classified as Protobuf, which seems wrong

utoni (approved):
we're loosing two flows, but seems fine to me :)

PR #2428, opened 2024-05-08
Protobuf: fix false positives

IvanNardi:


utoni (approved):
we're loosing one protobuf flow, but it's just a one packet flow, so I guess it's alright

IvanNardi:
> we're loosing one protobuf flow, but it's just a one packet flow, so I guess it's alright

If you open that trace, that Protobuf classification seems wrong anyway....

PR #2438, opened 2024-05-11
Add Call of Duty Mobile support

0xA50C1A1:

Describe changes:

The game is quite popular, [with several tens of millions of players](https://activeplayer.io/call-of-duty-mobile/) playing it every day, so I guess that support for detection of its traffic would not be superfluous.

The main issues are:

- Virtually all the flows contain some random-looking data, suggesting the traffic is likely encrypted.
- However, I've noticed that at the start of each new match, a new UDP session is started, and the 4th packet within this session consistently begins with `0xCE` and contains the string `LOC_PREFAB_LOADOUTNAME_1` in the first 100 bytes of the payload.
- I also noticed some short UDP sessions with a destination port of 7500, consisting of 2 packets with a payload length of 12 bytes, containing some random data and the string `ping`.


src/lib/protocols/cod_mobile.c
```diff
@@ -0,0 +1,83 @@
...
+  {
+     if (ndpi_memmem(packet->payload, packet->payload_packet_len, "LOC_PREFAB_LOADOUTNAME_1",
+         NDPI_STATICSTRING_LEN("LOC_PREFAB_LOADOUTNAME_1")))
+     {
+       ndpi_int_cod_mobile_add_connection(ndpi_struct, flow);
+       return;
+     }
+  }
+
+  if (flow->packet_counter > 4) {
```
IvanNardi: If it is the fourth packet with the 0xCE pattern, can we exclude the protocol with ">=4"?

0xA50C1A1: > If it is the fourth packet with the 0xCE pattern, can we exclude the protocol with ">=4"?

Sure.

IvanNardi (approved):
Thank you

PR #2439, opened 2024-05-11
Revert `ndpi_strnstr()` optimization introduced in a813121e0

IvanNardi:
New implementation fails tests 11b, 12 and 13.
Revert to the original (BSD) implementation (with also some basic parameters check)


0xA50C1A1:
@IvanNardi I apologize for my inattention, I should have double-checked. However, this issue can be easily fixed by replacing `strnlen` with `strlen`. Neither the new nor the old implementation of `ndpi_strnstr` is suitable for handling non null-terminated strings anyway, so we won't be losing anything

Issue #2413, opened 2024-04-29
DEBUG_TRACE Build fails due to domainSearchUnitTest()


vicky16101991:
## Describe the bug
Looks like this change of adding ndpi_str to the function call ndpi_domain_classify_size is breaking while we are tyring to execute in the DEBUG_TRACE mode and also the function defination of ndpi_domain_classify_size ndpi_api.h is not having a signature to accept 2 arguments.
the fname is also not found and causing issues while compiling in DEBUG_TRACE mode.

### Expected behavior
Code should be getting compiled in DEBUG_TRACE Mode
### Obtained behavior
Compilation fails with errors.

## nDPI Environment (please complete the following information):
* OS name: Debian.
* OS version: 11
* Architecture: arm64
* nDPI version or commit hash: dev/ad117bf.
* nDPI compilation flags used: if you are building from source -DDEBUG_TRACE.

## How to reproduce the reported bug
Compile with the flag -DDEBUG_TRACE

### Reproducible using ndpiReader?
-The reported bug is reproducible using ndpiReader.

PR #2460, opened 2024-05-28
Added NDPI_PROTOCOL_NTOP assert and removed percentage comparison

dandare100:

- [X] I have read the contributing guide lines at https://github.com/ntop/nDPI/blob/dev/CONTRIBUTING.md
- [NA] I have updated the documentation (in doc/) to reflect the changes made (if applicable)

Link to the related [issue](https://github.com/ntop/nDPI/issues): https://github.com/ntop/nDPI/issues/2413

Describe changes:
When building with the DEBUG_TRACE flag there was a section in nDPI reader under the search domain unit tests that broke the build.
This section appears to print the percent of the test map size in relation to the size of a file. It could be to see the efficiency of the encoding/zipping of the map entries.
The file handle (fname) was not defined.
I went back in the commit history to investigate the effort of re-instating "fname" definition and realized it has been like that for a while.

Also added an assert to verify the classification id of the original ntop.org domain entry.

This pull request fixes the build but removes the percentage report line
If you can let me know the file name that the comparison should be done against, I am happy to submit another pull request so that the percentage report line is re-instated.

### DIFF
--- a/example/ndpiReader.c
+++ b/example/ndpiReader.c
@@ -5957,6 +5957,7 @@ void domainSearchUnitTest() {
   ndpi_domain_classify_add(ndpi_str, sc, NDPI_PROTOCOL_NTOP, ".ntop.org");
   ndpi_domain_classify_add(ndpi_str, sc, NDPI_PROTOCOL_NTOP, domain);
   assert(ndpi_domain_classify_hostname(ndpi_str, sc, &class_id, domain));
+  assert(class_id == NDPI_PROTOCOL_NTOP);
 
   ndpi_domain_classify_add(ndpi_str, sc, NDPI_PROTOCOL_CATEGORY_GAMBLING, "123vc.club");
   assert(ndpi_domain_classify_hostname(ndpi_str, sc, &class_id, "123vc.club"));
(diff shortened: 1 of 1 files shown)


example/ndpiReader.c
```diff
@@ -5966,17 +5967,6 @@ void domainSearchUnitTest() {
...
-  struct stat st;
-
-  if(stat(fname, &st) == 0) {
-    u_int32_t s = ndpi_domain_classify_size(ndpi_str, sc);
-
-    printf("Size: %u [%.1f %% of the original filename size]\n",
-	   s, (float)(s * 100) / (float)st.st_size);
-  }
-#endif
-
```
IvanNardi, line 5979: I think it is quite safe to assume that we can keep the ifdef block, removing the "stat" usage and keeping something like `if(trace) { ndpi_domain_classify_size + fprintf(trace, "Size...`, to be coherent with the general usage of the `trace` variable throughout the file.
@dandare100, does it make sense to you?

IvanNardi:
@dandare100, very sorry for the delay
Could you try:
```
--- a/example/ndpiReader.c
+++ b/example/ndpiReader.c
@@ -5950,7 +5950,6 @@ void domainSearchUnitTest() {
   char *domain = "ntop.org";
   u_int16_t class_id;
   struct ndpi_detection_module_struct *ndpi_str = ndpi_init_detection_module(NULL);
-  u_int8_t trace = 0;
 
   assert(ndpi_str);
   assert(sc);
@@ -5968,9 +5967,10 @@ void domainSearchUnitTest() {
   assert(ndpi_domain_classify_hostname(ndpi_str, sc, &class_id, "blog.ntop.org"));
   assert(class_id == NDPI_PROTOCOL_NTOP);
 
+#ifdef DEBUG_TRACE
   u_int32_t s = ndpi_domain_classify_size(sc);
   if(trace) printf("ndpi_domain_classify size: %u \n",s);
-  
+#endif
```
on top of your change, please? If it is ok, could you push that, please?

dandare100:
Hello. No problem on delay.
Forgive me if I am wrong but if I add the ifdef for DEBUG_TRACE the test will never run.

Because of the logic mentioned above :

```
if(!skip_unit_tests) {
#ifndef DEBUG_TRACE
    /* Skip tests when debugging */
```
Is this the intended behaviour ? If so, I will add and push

IvanNardi:
@dandare100, good catch, you are totally right!
You patch is fine as it is.
Thank you very much

IvanNardi (approved):

IvanNardi:
@dandare100, thanks for your contribution

dandare100:
Anytime, this is a really awesome, well thought out  project.
I really dig it

PR #2483, opened 2024-06-26
fuzz: pl7m: add a custom mutator for better fuzzing of pcap files

IvanNardi:
Pl7m is a custom mutator (used for structure aware fuzzing) for network traffic packet captures (i.e. pcap files).

The output of the mutator is always a valid pcap file, containing the same flows/sessions of the input file. That's it: the mutator only changes the packet payload after the TCP/UDP header, keeping all the original L2/L3 information (IP addresses and L4 ports).

See: https://github.com/IvanNardi/pl7m


lucaderi:
Thank you great work

Issue #2322, opened 2024-02-19
Add First Packet Classification capability


IvanNardi:
In some contexts/deployments (load-balancing or QoS involved, routing on different interfaces based on classification...) is quite important to have a flow classification on the very FIRST packet of the flow, even if that classification is not as precise as it could be if more packets are analyzed.

Note that, for TCP, "first packet" means (usually) the SYN packet...

mmanoj:
@IvanNardi 

This is good to have,I like to implement this feature, if you can guide me.

IvanNardi:
@mmanoj , very sorry for the delay.
FPC is a complex topic, but the most easy (and obvious) ways to achieve that are:
* via IP match, improving the IP lists. nDPI already has some ip lists of the most important services (see `/util` and `utils/asn_update.sh`. We might add some other protocols/lists 
* using the DNS information. If clientA triggers the DNS transaction hostnameXXX/PROTOCOL_Y -> IP_XXX, then it is likely that a future flow from clientA to IP_XXX is related to PROTOCOL_Y

If you are still interested, we can discuss about the details

mmanoj:
@IvanNardi Thx for the reply, Yes I'm interested and done some initial work with nDPIsimple integration as well.Can we discuss in detail with some actionable steps, so we can focus on the step by step implementation.

IvanNardi:
1) Extend the ip lists. Take a look at the existing lists (in `/util` and `utils/asn_update.sh`) and try adding other ones: new important/used applications/services, new streaming services, new game networks, missing CDN providers.... There are already examples about integrating a list from a csv or txt file, or from BGP info: the hard part is to find an official/reliable source for those ips...

2) Create a new LRU cache where the key is the pair "Client_IP - Resolved_IP" and the value is a protocol id:
 * for each DNS A/AAAA pair create a new entry, where the protocol is the classification associated to the required hostname
 * for each flow (at the very first packet) lookup into the cache using "SRC IP - DST IP" as key and, if a match is found, save the protocol id into `ndpi_flow_struct` structure as a "fpc result"

mmanoj:
@IvanNardi 

Thanks for the update,I will work on below plan.
Step #1. Start with action item #2 with existing IP list to implement PoC with nDPI-Simple integration example.

Please clarify below point: Is current nDPI DNS dissectors support to capture DNS answer (A/AAAA) record and IP address associated with it or need to develop that part? anyway I will lookup the code as well.

[...]

PR #2488, opened 2024-07-02
Add infrastructure for explicit support of First Packet Classification

IvanNardi:
Let's start with some basic helpers and with FPC based on flow addresses.

See: #2322


src/lib/ndpi_main.c
```diff
@@ -9383,6 +9407,25 @@ const char *ndpi_confidence_get_name(ndpi_confidence_t confidence)
...
+    return "Unknown";
+
+  case NDPI_FPC_CONFIDENCE_IP:
+    return "IP address";
+
+  case NDPI_FPC_CONFIDENCE_DNS:
+    return "DNS";
+
+  default:
+    return NULL;
```
utoni: Better to return something like `"n/a"` instead of a `NULL` or at least document that it may return `NULL` if not in sync.

IvanNardi: Good idea. Done

utoni (approved):

PR #2447, opened 2024-05-17
Replace ndpi_strnstr() implementation with an optimal one

0xA50C1A1:
Please sign (check) the below before submitting the Pull Request:

- [x] I have signed the ntop Contributor License Agreement at https://github.com/ntop/legal/blob/main/individual-contributor-licence-agreement.md
- [x] I have read the contributing guide lines at https://github.com/ntop/nDPI/blob/dev/CONTRIBUTING.md
- [x] I have updated the documentation (in doc/) to reflect the changes made (if applicable)

#2433 #2439

Describe changes:

I apologize for the buggy implementation I suggested last time. This time, I simply took the already quite optimal `ndpi_memmem()` code and reworked it to handle null-terminated strings. This implementation shows better results in the benchmark, even with the `strlen` and `strnlen` calls inside. I've thoroughly tested it, and your unit tests are passing correctly.

Main changes:

- **Efficient substring search**: The new implementation utilizes `memchr` to quickly locate the first character of the `needle` substring within the `haystack`. This avoids unnecessary character comparisons, leading to better performance, especially for long strings.
- **Early exit**: If the remainder of `haystack` is shorter than `needle`, then there's no point in doing unnecessary comparisons.
- **Single loop**: The new version uses a single loop to search for the needle

@IvanNardi @utoni what do you think, guys?

PR #2433, opened 2024-05-10
ndpi_strnstr optimization

0xA50C1A1:
Please sign (check) the below before submitting the Pull Request:

- [x] I have signed the ntop Contributor License Agreement at https://github.com/ntop/legal/blob/main/individual-contributor-licence-agreement.md
- [x] I have read the contributing guide lines at https://github.com/ntop/nDPI/blob/dev/CONTRIBUTING.md
- [x] I have updated the documentation (in doc/) to reflect the changes made (if applicable)

Describe changes:

On average, the new implementation is 2.5 times faster. Although, of course, there are test cases where it shows slightly worse results.

Also I wrote a small [benchmark](https://github.com/ntop/nDPI/files/15272200/benchmark.zip) to compare different implementations if you're interested.

PR #2494, opened 2024-07-11
Optimize performance of ndpi_strnstr() and possible bugfix

pasabanov:

#2433 #2439 #2447

Describe changes:

I made some minor changes to the current optimized substring search.

The main changes are:
1. Removed `len == 0` check from the first `if` statement. Now the function doesn't immediately return `NULL` if `len` is zero.  
I believe that the previous behavior was a bug, because if `needle_len` is zero, the function should return `haystack`, because the  empty substring does reside at the start of the empty string.
2. Changed the order of `hs_real_len` computation and `if (needle_len == 0)` check.
3. Removed the if `(needle_len == 1)` check because it only provides a performance benefit in specific cases. If you need to frequently search for single-character substrings, you might want to consider reverting this change.
4. Replaced variable `haystack_end` with variable `end_of_search` (the byte after the last interesting byte).  
Using `end_of_search` inside the `memchr` function instead of `haystack_end` eliminates the need for the `current + needle_len <= haystack_end` check, as it is always positive if `current` is not `NULL`.

I also added the new function to the performance tests.
In practice `ndpi_strnstr_opt2` is slightly faster than `ndpi_strnstr_opt` in most runs, though sometimes `ndpi_strnstr_opt` appears to be faster.

### DIFF
--- a/example/ndpiReader.c
+++ b/example/ndpiReader.c
@@ -5782,6 +5782,13 @@ void strnstrUnitTest(void) {
 
   /* Test 13 */
   assert(ndpi_strnstr("abcdef", "abc", 2) == NULL);
+
+  /* Test 14: zero length */
+  assert(strcmp(ndpi_strnstr("", "", 0), "") == 0);
+  assert(strcmp(ndpi_strnstr("string", "", 0), "string") == 0);
+  assert(ndpi_strnstr("", "str", 0) == NULL);
+  assert(ndpi_strnstr("string", "str", 0) == NULL);
+  assert(ndpi_strnstr("str", "string", 0) == NULL);
 }
 
 /* *********************************************** */
--- a/src/lib/ndpi_main.c
+++ b/src/lib/ndpi_main.c
@@ -9737,17 +9737,23 @@ void ndpi_dump_risks_score(FILE *risk_out) {
 
 char *ndpi_strnstr(const char *haystack, const char *needle, size_t len)
 {
-  if (!haystack || !needle || len == 0)
+  if (!haystack || !needle)
   {
     return NULL;
   }
 
-  size_t needle_len = strlen(needle);
-  size_t hs_real_len = strnlen(haystack, len);
+  const size_t needle_len = strlen(needle);
 
   if (needle_len == 0)
   {
     return (char *)haystack;
+  }
+
+  const size_t hs_real_len = strnlen(haystack, len);
+
+  if (needle_len == 1)
+  {
+    return (char *)memchr(haystack, *needle, hs_real_len);
   }
 
   if (needle_len > hs_real_len)
@@ -9755,29 +9761,24 @@ char *ndpi_strnstr(const char *haystack, const char *needle, size_t len)
     return NULL;
   }
 
-  if (needle_len == 1)
-  {
-    return (char *)memchr(haystack, *needle, hs_real_len);
-  }
+  const char *const end_of_search = haystack + hs_real_len - needle_len + 1;
 
   const char *current = haystack;
-  const char *haystack_end = haystack + hs_real_len;
-
-  while (current <= haystack_end - needle_len)
+  while (current < end_of_search)
   {
-    current = (const char *)memchr(current, *needle, haystack_end - current);
+    current = (const char *)memchr(current, *needle, end_of_search - current);
 
     if (!current)
     {
       return NULL;
     }
 
-    if ((current + needle_len <= haystack_end) && memcmp(current, needle, needle_len) == 0)
+    if (memcmp(current, needle, needle_len) == 0)
     {
       return (char *)current;
     }
 
-    current++;
+    ++current;
   }
 
   return NULL;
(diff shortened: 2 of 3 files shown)


pasabanov:
Actually,
```
  if (needle_len > hs_real_len)
  {
    return NULL;
  }
```
can be removed too because, in that case, the loop won't perform any iterations and the function will return NULL anyways.

But I'm nut sure if this will increase or decrease performance in your case.

IvanNardi:
>     1. Removed `len == 0` check from the first `if` statement. Now the function doesn't immediately return `NULL` if `len` is zero.
>        I believe that the previous behavior was a bug, because if `needle_len` is zero, the function should return `haystack`, because the  empty substring does reside at the start of the empty string.


Could you add a unit test about that case in `strnstrUnitTest()`, please?
@pasabanov, could you rebase?

pasabanov:
@IvanNardi, I've added unit test cases for zero length in `strnstrUnitTest()` and rebased.

pasabanov:
@IvanNardi, what do you think about removing the checks `if (needle_len == 1)`, which I've already removed, and `if (needle_len > hs_real_len)`, which I suggest removing?
They're part of the general case, so if you don't search for one-character needles or needles that are longer than haystacks often, then there might be no point in optimizing these cases.

0xA50C1A1:
> @IvanNardi, what do you think about removing the checks `if (needle_len == 1)`, which I've already removed, and `if (needle_len > hs_real_len)`, which I suggest removing? They're part of the general case, so if you don't search for one-character needles or needles that are longer than haystacks often, then there might be no point in optimizing these cases.

Actually there are some cases where `ndpi_strnstr` is used to search for a single character, that's why I added this optimization. Well, thanks anyway, great work!

**mgcp.c**
```
    endpoint = ndpi_strnstr((char const *)packet->payload + 5, " ", packet->payload_packet_len - 5);
    if (endpoint == NULL)
    {
      break;
    }
    endpoint++;

    mgcp = ndpi_strnstr(endpoint, " ", packet->payload_packet_len - ((u_int8_t const *)endpoint - packet->payload));
    if (mgcp == NULL)
    {
      break;
    }
    mgcp++;
 ```

**ethereum.c**
```
    } else if(ndpi_strnstr((const char *)packet->payload, "{", packet->payload_packet_len)
```

**xiaomi.c**
```
        ptr = ndpi_strnstr((const char *)&payload[offset], ":", len);
 ```

**tivoconnect.c**
```
 for (newline = ndpi_strnstr(payload, "\n", payload_len);
       newline != NULL;
       key = ++newline,
       newline = ndpi_strnstr(newline, "\n", payload_len - (newline - payload)))
  {
 ```

pasabanov:
> Actually there are some cases where `ndpi_strnstr` is used to search for a single character, that's why I added this optimization. Well, thanks anyway, great work!

Thanks!

In that case, I think I should add that check back.

What do you think about the second check `if (needle_len > hs_real_len)`? Is it needed too?  
It saves a fairly small number of instructions when the needle is longer than the haystack.

0xA50C1A1:
> > Actually there are some cases where `ndpi_strnstr` is used to search for a single character, that's why I added this optimization. Well, thanks anyway, great work!
> 
> Thanks!
> 
> In that case, I think I should add that check back.
> 
> What do you think about the second check `if (needle_len > hs_real_len)`? Is it needed too? It saves a fairly small number of instructions when the needle is longer than the haystack.

I dunno, I guess you could keep it. Usually the compiler manages skipping redundant checks.

pasabanov:
> I dunno, I guess you could keep it. Usually the compiler manages skipping redundant checks.

Ok. I've returned the removed check `if (needle_len == 1)`, but I switched its order with the `if (needle_len > hs_real_len)` check because I think that would be more efficient.

By the way, the compiler doesn't remove the check `if (needle_len > hs_real_len)` at any level of optimization. I checked that with latest versions of gcc and clang on godbolt.org.

0xA50C1A1:
> > I dunno, I guess you could keep it. Usually the compiler manages skipping redundant checks.
> 
> Ok. I've returned the removed check `if (needle_len == 1)`, but I switched its order with the `if (needle_len > hs_real_len)` check because I think that would be more efficient.
> 
> By the way, the compiler doesn't remove the check `if (needle_len > hs_real_len)` at any level of optimization. I checked that with latest versions of gcc and clang on godbolt.org.

Alright, it looks fine. You can take a look at `ndpi_memmem` if you want, because `ndpi_strnstr` is basically a copypaste of that function.

pasabanov:
> Alright, it looks fine. You can take a look at `ndpi_memmem` if you want, because `ndpi_strnstr` is basically a copypaste of that function.

I took a look at `ndpi_memmem` and saw that it has the same issues. I've created pull request #2499.

IvanNardi (approved):
LGTM

IvanNardi:
@pasabanov, thank you for this contribution
@0xA50C1A1, thanks for the review

PR #2500, opened 2024-07-15
Refactor ndpi_strnstr to use ndpi_memmem

0xA50C1A1:

Describe changes:

### DIFF
--- a/src/lib/ndpi_main.c
+++ b/src/lib/ndpi_main.c
@@ -9741,19 +9741,7 @@ char *ndpi_strnstr(const char *haystack, const char *needle, size_t len)
     return NULL;
   }
 
-  const size_t needle_len = strlen(needle);
-
-  if (needle_len == 0) {
-    return (char*)haystack;
-  }
-
-  const size_t hs_real_len = strnlen(haystack, len);
-
-  if (needle_len > hs_real_len) {
-    return NULL;
-  }
-
-  return (char *)ndpi_memmem(haystack, hs_real_len, needle, needle_len);
+  return (char *)ndpi_memmem(haystack, strnlen(haystack, len), needle, strlen(needle));
 }
 
 /* ****************************************************** */
@@ -11424,7 +11412,8 @@ void* ndpi_memmem(const void* haystack, size_t haystack_len, const void* needle,
   const u_int8_t *const end_of_search = (const u_int8_t *)haystack + haystack_len - needle_len + 1;
 
   const u_int8_t *current = (const u_int8_t *)haystack;
-  while (current < end_of_search) {
+
+  while (1) {
     /* Find the first occurrence of the first character from the needle */
     current = (const u_int8_t *)memchr(current, *(const u_int8_t *)needle, end_of_search - current);
 
--- a/src/lib/ndpi_main.c
+++ b/src/lib/ndpi_main.c
@@ -9737,51 +9737,23 @@ void ndpi_dump_risks_score(FILE *risk_out) {
 
 char *ndpi_strnstr(const char *haystack, const char *needle, size_t len)
 {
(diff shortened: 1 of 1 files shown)


IvanNardi:
That's exactly what I was thinking (t's a pleasure to hear from you again!)

IvanNardi:
@pasabanov, any thoughts?

pasabanov:
> @pasabanov, any thoughts?

Yes, I have a few thoughts.

1. Using `ndpi_memmem` in `ndpi_strnstr` slightly slows down the execution of `ndpi_strnstr` when `needle_len == 1`.  
It might be worth returning this optimization to `ndpi_strnstr`.
3. There is also the possibility to getting rid of the checks in `ndpi_strnstr` altogether, as they are already performed in `ndpi_memmem`. I can see the potential to write `ndpi_strnstr` as a one-liner.

Additionally, I have a thought about `ndpi_memmem`:
It might make sense to check `current < end_of_search` or, similarly, `needle_len <= haystack_len` only once, and then use an infinite loop instead of `while (current < end_of_search)`, because `memchr` on each iteration returns either `NULL` or a pointer between `current` and `end_of_search`.

And one more question: can you provide a link to your code style guide? I assume that you prefer the style without line breaks before curly braces.

0xA50C1A1:
>     2. There is also the possibility to getting rid of the checks in `ndpi_strnstr` altogether, as they are already performed in `ndpi_memmem`. I can see the potential to write `ndpi_strnstr` as a one-liner.

You mean the typical `NULL` checks? I guess I can remove them, since they are already present in `strlen` and `strnlen`, as well as in `ndpi_memmem`.

pasabanov:
> It might make sense to check `current < end_of_search` or, similarly, `needle_len <= haystack_len` only once

Actually, the check `needle_len <= haystack_len` in `ndpi_memmem` occurs in `if (!haystack || !needle || haystack_len < needle_len)`, so I think we can just replace `while (current < end_of_search)` with `while (1)` or another variant of an infinite loop.

0xA50C1A1:
> > It might make sense to check `current < end_of_search` or, similarly, `needle_len <= haystack_len` only once
> 
> Actually, the check `needle_len <= haystack_len` in `ndpi_memmem` occurs in `if (!haystack || !needle || haystack_len < needle_len)`, so I think we can just replace `while (current < end_of_search)` with `while (1)` or another variant of an infinite loop.

Thanks, that's a good idea, might increase performance a bit. I'll modify it now.

pasabanov:
> You mean the typical `NULL` checks? I guess I can remove them, since they are already present in `strlen` and `strnlen`, as well as in `ndpi_memmem`.

I mean all the checks: `if (!haystack || !needle)`, `if (needle_len == 0)` and `if (needle_len > hs_real_len)`. They're all performed in `ndpi_memmem`.

Oh, wait! I think that `strlen` and `strnlen` do not check for `NULL`, so you should keep the `if (!haystack || !needle)` check.

0xA50C1A1:
> Oh, wait! I think that `strlen` and `strnlen` do not check for `NULL`, so you should keep the `if (!haystack || !needle)` check.

Maybe it depends on the implementation, glibc has these checks. Alright, I'll keep them.

pasabanov:
I think it would be prettier to write `return (char *)ndpi_memmem(haystack, strnlen(haystack, len), needle, strlen(needle));`.

0xA50C1A1:
> I think it would be prettier to write `return (char *)ndpi_memmem(haystack, strnlen(haystack, len), needle, strlen(needle));`.

Alright, let's make this simpler.

IvanNardi (approved):

IvanNardi:
Thanks everyone

PR #2502, opened 2024-07-16
Add HLS support

0xA50C1A1:

Describe changes:

Nothing fancy, just a check against the HTTP Content-Type guaranteed to be used by that protocol.


src/lib/ndpi_main.c
```diff
@@ -1029,6 +1029,7 @@ static void ndpi_init_protocol_defaults(struct ndpi_detection_module_struct *ndp
 			      NDPI_PROTOCOL_APACHE_THRIFT,
 			      NDPI_PROTOCOL_JSON_RPC,
 			      NDPI_PROTOCOL_HL7,
+			      NDPI_PROTOCOL_HLS,
```
IvanNardi: We shouldn't need that (because we don't have a dissector for HLS; all the logic is in the HTTP code)
That should drastically simplify the diff of this PR

IvanNardi (changes requested):

src/lib/ndpi_main.c
0xA50C1A1: > We shouldn't need that (because we don't have a dissector for HLS; all the logic is in the HTTP code) That should drastically simplify the diff of this PR

Got it, I'll remove it.

0xA50C1A1:
I'll update the unit test results a bit later.

IvanNardi (approved):

IvanNardi:
Thank you

PR #2506, opened 2024-07-17
shell: reformatted, fixed inspections, typos

pasabanov:

#1986 #2006 #2113

Describe changes:

Reformatted shell scripts according to [ShellCheck](https://github.com/koalaman/shellcheck/).

I. Most common changes:
1. https://github.com/koalaman/shellcheck/wiki/SC2086
	`$var` → `"$var"`
	Note: this isn't always necessary and I've been careful not to substitute where it wasn't necessary in meaning.
2. https://github.com/koalaman/shellcheck/wiki/SC2006
	`` `command` `` → `$(command)`
3. https://github.com/koalaman/shellcheck/wiki/SC2004
	`$(( $a + $b ))` → `$(( a + b ))`
4. https://github.com/koalaman/shellcheck/wiki/SC2164
	`cd "$dir"` → `cd "$dir" || exit 1`
5. https://github.com/koalaman/shellcheck/wiki/SC2166
	`[ check1 -o check2 ]` → `[ check1 ] || [ check2 ]`
6. https://github.com/koalaman/shellcheck/wiki/SC2002
	`cat "${file}" | wc -c` → `< "${file}" wc -c`
	Note: this looks a bit uglier but works faster.

II. Some special changes:
1. In file `utils/common.sh`:
	https://github.com/koalaman/shellcheck/wiki/SC2112
	This script is interpreted by `sh`, not by `bash`, but uses the keyword `function`.
	So I replaced `#!/usr/bin/env sh` to `#!/usr/bin/env bash`.
2. After that I thought of replacing all shebangs to `#!/usr/bin/env bash` for consistency and cross-platform compatibility, especially since most of the files already use bash.
3. But in cases when it was `#!/bin/sh -e` or `#!/bin/bash -eu` another problem appears:
	https://github.com/koalaman/shellcheck/wiki/SC2096
	So I decided to make all shebangs look uniform:
	```
	#!/usr/bin/env bash
	set -e (or set -eu) (if needed)
	```
4. In file `tests/ossfuzz.sh`:
	https://github.com/koalaman/shellcheck/wiki/SC2162
	`read i` → `read -r i`
	Note: I think that there is no need in special treatment for backslashes, but I could be wrong.
5. In file `tests/do.sh.in`:
	https://github.com/koalaman/shellcheck/wiki/SC2035
	`ls *.*cap*` → `ls -- *.*cap*`
6. In file `utils/verify_dist_tarball.sh`:
	https://github.com/koalaman/shellcheck/wiki/SC2268
	`[ "x${TARBALL}" = x ]` → `[ -z "${TARBALL}" ]`
7. In file `utils/check_symbols.sh`:
	https://github.com/koalaman/shellcheck/wiki/SC2221
	`'[ndpi_utils.o]'|'[ndpi_memory.o]'|'[roaring.o]')` → `'[ndpi_utils.o]'|'[ndpi_memory.o]')`
8. In file `autogen.sh`:
	https://github.com/koalaman/shellcheck/wiki/SC2145
	`echo "./configure $@"` → `echo "./configure $*"`
	https://github.com/koalaman/shellcheck/wiki/SC2068
	`./configure $@` → `./configure "$@"`

III. `LIST6_MERGED` and `LIST_MERGED6`
	There were typos with this variables in files `utils/aws_ip_addresses_download.sh`, `utils/aws_ip_addresses_download.sh` and `utils/microsoft_ip_addresses_download.sh` where variable `LIST6_MERGED` was defined, but `LIST_MERGED6` was removed by `rm`.
	I changed all `LIST_MERGED6` to `LIST6_MERGED`.

Not all changes are absolutely necessary, but some may save you from future bugs.

### DIFF
--- a/autogen.sh
+++ b/autogen.sh
@@ -1,4 +1,4 @@
-#!/bin/sh
+#!/usr/bin/env sh
 
 rm -f configure config.h config.h.in
 
@@ -42,7 +42,6 @@ fi
 
 autoreconf -ivf
 
-echo "./configure $@"
+echo "./configure $*"
 chmod +x configure
-./configure $@
-
+./configure "$@"
--- a/packages/ubuntu/debian/postinst
+++ b/packages/ubuntu/debian/postinst
@@ -1,4 +1,5 @@
-#!/bin/sh -e
+#!/usr/bin/env sh
+set -e
 
 case "$1" in
     configure)
--- a/packages/version.sh
+++ b/packages/version.sh
@@ -1,9 +1,9 @@
-#!/bin/bash
+#!/usr/bin/env sh
 
-SCRIPTPATH="$(cd "$(dirname "$0")"; pwd -P)"
-RELEASE="$(cd ${SCRIPTPATH}; cat ../configure.ac|grep C_INIT|cut -c 20-|rev|cut -c 3-|rev)"
-MAJOR_RELEASE="$(cd ${SCRIPTPATH}; cat ../configure.ac|grep C_INIT|cut -c 20-|rev|cut -c 3-|rev|cut -d. -f1)"
-REVISION="$(cd ${SCRIPTPATH}; git rev-list --all |wc -l | tr -d '[[:space:]]')"
+SCRIPTPATH="$(cd "$(dirname "$0")" || exit 1; pwd -P)"
+RELEASE="$(cd "${SCRIPTPATH}" || exit 1; < ../configure.ac grep C_INIT | cut -c 20- | rev | cut -c 3- | rev)"
+MAJOR_RELEASE="$(cd "${SCRIPTPATH}" || exit 1; < ../configure.ac grep C_INIT | cut -c 20- | rev | cut -c 3- | rev | cut -d. -f1)"
+REVISION="$(cd "${SCRIPTPATH}" || exit 1; git rev-list --all | wc -l | tr -d '[[:space:]]')"
 
 get_release() {
 	echo "${RELEASE}"
--- a/utils/check_symbols.sh
+++ b/utils/check_symbols.sh
@@ -1,18 +1,17 @@
-#!/usr/bin/env sh
-
+#!/usr/bin/env bash
 set -e
 
-SCRIPT_DIR="$(realpath $(dirname ${0}))"
+SCRIPT_DIR="$(realpath "$(dirname "${0}")")"
 NDPI_LIB="${1:-${SCRIPT_DIR}/../src/lib/libndpi.a}"
 
 if [ ! -r "${NDPI_LIB}" ]; then
-    printf '%s\n' "${0}: nDPI static library '$(realpath ${NDPI_LIB})' not found."
+    printf '%s\n' "${0}: nDPI static library '$(realpath "${NDPI_LIB}")' not found."
     exit 1
 fi
 
 FAIL_COUNT=0
 CURRENT_OBJECT=''
-for line in `nm -P -u "${NDPI_LIB}"`; do
+for line in $(nm -P -u "${NDPI_LIB}"); do
     OBJECT="$(printf '%s' "${line}" | grep -E "^${NDPI_LIB}\[.*\.o\]:" | grep -oE "\[.*\.o\]" || true)"
     if [ ! -z "${OBJECT}" ]; then
         CURRENT_OBJECT="${OBJECT}"
@@ -39,7 +38,7 @@ for line in `nm -P -u "${NDPI_LIB}"`; do
                     'malloc'|'calloc'|'realloc'|'free') SKIP=1 ;;
                 esac
             ;;
-            '[ndpi_utils.o]'|'[ndpi_memory.o]'|'[roaring.o]')
+            '[ndpi_utils.o]'|'[ndpi_memory.o]')
                 case "${FOUND_SYMBOL}" in
                     'malloc'|'calloc'|'free') SKIP=1 ;;
                 esac
@@ -52,7 +51,7 @@ for line in `nm -P -u "${NDPI_LIB}"`; do
         esac
 
         if [ ${SKIP} -eq 0 ]; then
-            FAIL_COUNT="$(expr ${FAIL_COUNT} + 1)"
+            FAIL_COUNT="$((FAIL_COUNT + 1))"
             printf '%s: %s\n' "${CURRENT_OBJECT}" "${FOUND_SYMBOL}"
         fi
     fi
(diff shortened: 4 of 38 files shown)


utils/check_symbols.sh
```diff
@@ -39,7 +38,7 @@ for line in `nm -P -u "${NDPI_LIB}"`; do
                     'malloc'|'calloc'|'realloc'|'free') SKIP=1 ;;
                 esac
             ;;
-            '[ndpi_utils.o]'|'[ndpi_memory.o]'|'[roaring.o]')
+            '[ndpi_utils.o]'|'[ndpi_memory.o]')
```
IvanNardi, line 41: @pasabanov, could you explain why checking  `roaring.o` is no more necessary, please?

pasabanov: Because there is the same pattern before on line 37 (36).  
This pattern never matches.

IvanNardi: yeah, you are right. Thanks!

autogen.sh
```diff
@@ -1,4 +1,4 @@
-#!/bin/sh
+#!/usr/bin/env bash
```
utoni: Why bash?

autogen.sh
```diff
@@ -42,7 +42,6 @@ fi
 
 autoreconf -ivf
 
-echo "./configure $@"
+echo "./configure $*"
 chmod +x configure
-./configure $@
-
+./configure "$@"
```
utoni, line 47: Does `./autogen.sh --with-local-libgcrypt --with-pcre2` still work?

packages/ubuntu/debian/postinst
```diff
@@ -1,4 +1,5 @@
-#!/bin/sh -e
+#!/usr/bin/env bash
```
utoni: Again. Why bash?

packages/version.sh
```diff
@@ -1,9 +1,9 @@
-#!/bin/bash
+#!/usr/bin/env bash
 
-SCRIPTPATH="$(cd "$(dirname "$0")"; pwd -P)"
-RELEASE="$(cd ${SCRIPTPATH}; cat ../configure.ac|grep C_INIT|cut -c 20-|rev|cut -c 3-|rev)"
-MAJOR_RELEASE="$(cd ${SCRIPTPATH}; cat ../configure.ac|grep C_INIT|cut -c 20-|rev|cut -c 3-|rev|cut -d. -f1)"
-REVISION="$(cd ${SCRIPTPATH}; git rev-list --all |wc -l | tr -d '[[:space:]]')"
+SCRIPTPATH="$(cd "$(dirname "$0")" || exit 1; pwd -P)"
```
utoni, line 3: Why `exit` ?

autogen.sh
pasabanov: For consistency, since most of the scripts in the project use bash.  
I think it would be convenient to use the same shell everywhere.

It's not particularly important in this file, but in `utils/common.sh` it is important to either use `bash` or remove the `function` keyword.

If you are against it, I can replace `#!/usr/bin/env bash` with `#!/usr/bin/env sh` in the files where it is not crucial.

autogen.sh
pasabanov: As far as I can tell, yes. I managed to successfully build the project with these parameters just now.

As [SC2068](https://github.com/koalaman/shellcheck/wiki/SC2068) says:
> Let's say you have four arguments: `baz`, `foo bar`, `*` and `/*/*/*/*`
>
> "$@" will expand into exactly that: `baz`, `foo bar`, `*` and `/*/*/*/*`
>
> $@ will expand into multiple other arguments: `baz`, `foo`, `bar`, `file.txt`, `otherfile.jpg`, and (eventually) a list of most files on the system

So it should only affect whitespace and globbing for each argument separately.

packages/ubuntu/debian/postinst
pasabanov: For the same reason.

By the way, what do you think about replacing
```
#!/bin/sh -e
```
with
```
#!/usr/bin/env (sh|bash)
set -e
```
?

packages/version.sh
pasabanov: Actually, I think that in `SCRIPTPATH` there is no need to exit.  
For other variables, it makes sense because if we fail to change the directory, there is no need to evaluate the variables.  
However, since there are no destructive commands, this might not be necessary either.

packages/ubuntu/debian/postinst
utoni: I am always using `set -e`. So this is not an issue for me. :)

autogen.sh
utoni: IMHO it would be great to run `autogen.sh` with mostly every possible shell symlinked to `/bin/sh`.

IvanNardi: I agree. I prefer to keep generic sh for autogen and packages scripts: I don't want to add bash dependency, just to be extra cautious

pasabanov: OK. Then for `autogen.sh` and package scripts I'll change the shell back to `sh`.

IvanNardi: > OK. Then for `autogen.sh` and package scripts I'll change the shell back to `sh`.

Thanks

packages/ubuntu/debian/postinst
utoni: Ok, I think using bash here should be fine for Ubuntu/Debian based distros.

packages/version.sh
utoni: It complicates the code quite a bit. :cat:

packages/ubuntu/debian/postinst
pasabanov: I've already changed that back to `sh` after the discussion about `autogen.sh` :)

packages/version.sh
pasabanov: Maybe we should keep `exit` for cases when some `cd` is unseccessful?
It won't cause any issues in normal cases, but might prevent errors in defining the variables.

pasabanov: I didn't see your last response when I was writing my last comment.

So, do you think I should remove `exit`?

utoni: I like the idea in general. But as it's done now it complicates the code a bit. Anyway, let's do it as you've proposed. :)

utoni (approved):
LGTM

IvanNardi (approved):
Particularly appreciated the detailed commit message

pasabanov:
> Particularly appreciated the detailed commit message

You mean pull request message?

The commit message is not that detailed.

IvanNardi:
> > Particularly appreciated the detailed commit message
> 
> You mean pull request message?
> 
> The commit message is not that detailed.

Yes, PR message, sorry. I copied it in the commit message

PR #2499, opened 2024-07-14
ndpi_memmem: optimized, fixed bug, added tests

pasabanov:
Please sign (check) the below before submitting the Pull Request:

- [x] I have signed the ntop Contributor License Agreement at https://github.com/ntop/legal/blob/main/individual-contributor-licence-agreement.md
- [x] I have read the contributing guide lines at https://github.com/ntop/nDPI/blob/dev/CONTRIBUTING.md
- [x] I have updated the documentation (in doc/) to reflect the changes made (if applicable)

#2378 #2424 #2494

Describe changes:

Similar changes to [this PR (#2494)](https://github.com/ntop/nDPI/pull/2494).

1. Made the function `ndpi_memmem` return `haystack` when `needle_len == 0` instead of `NULL`. This is more logical and aligns with the default behavior for standard implementations, such as [GNU libc](https://github.com/bminor/glibc/blob/master/string/memmem.c) and [FreeBSD libc](https://github.com/freebsd/freebsd-src/blob/main/lib/libc/string/memmem.c).
2. Optimized performance by replacing the variable `haystack_end` with the variable `end_of_search`, which reduces the number of checks during the search.
3. Added unit tests for the function `ndpi_memmem`.

PR #2507, opened 2024-07-17
ndpi_strncasestr: optimization, fixes, tests

pasabanov:

#2494 #2499

Describe changes:

I checked `ndpi_strncasestr` function and found that it can also be improved.

I used the same approach as with `ndpi_strnstr` and `ndpi_memmem` functions.

Changes:
1. Renamed variables according to the function declaration.
2. Added a `NULL` check for `s` and `find`.
3. Added a check for an empty `find` (in this case, `s` is returned).
4. Replaced the variable `int i` with `const char *const end_of_search`, which allows reducing the number of checks in each iteration of the loop.
5. Added unit tests (copied from `strnstrUnitTest` and modified a bit).

### DIFF
--- a/src/lib/ndpi_main.c
+++ b/src/lib/ndpi_main.c
@@ -9765,18 +9765,26 @@ char *ndpi_strnstr(const char *haystack, const char *needle, size_t len)
  * Same as ndpi_strnstr but case-insensitive.
  * Please note that this function is *NOT* equivalent to strncasecmp().
  */
-const char * ndpi_strncasestr(const char *str1, const char *str2, size_t len) {
-  size_t str1_len = strnlen(str1, len);
-  size_t str2_len = strlen(str2);
-  int i; /* signed! */
+const char * ndpi_strncasestr(const char *s, const char *find, size_t len) {
 
-  for(i = 0; i < (int)(str1_len - str2_len + 1); i++){
-    if(str1[0] == '\0')
-      return NULL;
-    else if(strncasecmp(str1, str2, str2_len) == 0)
-      return(str1);
+  if (!s || !find) {
+    return NULL;
+  }
 
-    str1++;
+  const size_t find_len = strlen(find);
+
+  if (find_len == 0) {
+    return s;
+  }
+
+  const size_t s_len = strnlen(s, len);
+
+  const char *const end_of_search = s + s_len - find_len + 1;
+
+  for (; s < end_of_search; ++s) {
+    if (strncasecmp(s, find, find_len) == 0) {
+      return s;
+    }
   }
 
   return NULL;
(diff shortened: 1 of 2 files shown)


IvanNardi:
@0xA50C1A1, what do you think?

src/lib/ndpi_main.c
```diff
@@ -9758,18 +9758,26 @@ char *ndpi_strnstr(const char *haystack, const char *needle, size_t len)
...
+  }
+
+  const size_t s_len = strnlen(s, len);
+
+  const char *const end_of_search = s + s_len - find_len + 1;
+
+  for (; s < end_of_search; ++s) {
+    if (strncasecmp(s, find, find_len) == 0) {
+      return s;
+    }
```
0xA50C1A1: @pasabanov @IvanNardi I guess it would be nice to add such check to get rid of unnecessary `strncasecmp` calls.

```
  for (; s < end_of_search; ++s) {
    if (tolower((u_int8_t)*s) == tolower((u_int8_t)find[0])) {
      if (strncasecmp(s, find, find_len) == 0) {
        return s;
      }
    }
  }
```

pasabanov: I think that would improve performance in most cases.
I'll implement that.

pasabanov: But I'm not sure whether I should cast `char` to `unsigned char` for `tolower` or not.
Do you have any arguments in favor of casting?

IvanNardi: no cast, the prototype is `int tolower(int c);`

0xA50C1A1: > no cast, the prototype is `int tolower(int c);`

Oh, shit, I'm sorry. For some reason I thought the argument type was `unsigned char`.

pasabanov: Oh. I found the explanation in `man tolower`:

> If c is neither an unsigned char value nor EOF, the behavior of these functions is undefined.

So we should cast.

pasabanov: However, since we're using `char` in the code, I think it's better to cast to `unsigned char` rather than `u_int8_t`.

IvanNardi: > Oh. I found the explanation in `man tolower`:
> 
> > If c is neither an unsigned char value nor EOF, the behavior of these functions is undefined.
> 
> So we should cast.

It is possible that the parameter is EOF? If not, we can cast without any issues; otherwise we should at least check for it, IMO

pasabanov: > It is possible that the parameter is EOF? If not, we can cast without any issues; otherwise we should at least check for it, IMO

I think that placing EOF inside a string is so wrong that nobody would do that.

And in case EOF ends up in a string due to some bug, I believe it would be detected somehow else.
I don't think sacrificing performance for a scenario that may never occur is necessary.

pasabanov: I've added the optimization.
Since `find_len` is guaranteed to be greater than zero, we can call `strncasecmp` on the remaining parts of the strings without the first characters.
It should improve performance as well.

IvanNardi (approved):

IvanNardi:
Thanks everyone

Issue #2530, opened 2024-08-25
CNC-IP false positives


IvanNardi:
With latest code, there are some CNC-IP false positives. 
For examples:  [cncip_false_positives.zip](https://github.com/user-attachments/files/16740132/cncip_false_positives.zip)
Should we revert to the session-id match?

IvanNardi:
@0xA50C1A1, could you take a look, please?

IvanNardi:
Fixed in 3189f19b0fea497ad2108ef498d04818a015d329

PR #2531, opened 2024-08-25
Fix CNP-IP false positives

0xA50C1A1:

Link to the related [issue](https://github.com/ntop/nDPI/issues): #2530


IvanNardi:
@0xA50C1A1, thanks for fixing that!
Could you try to add these false positives flows to `tests/cfgs/default/pcap/false_positives.pcapng`, please?

src/include/ndpi_typedefs.h
```diff
@@ -1512,6 +1512,10 @@ struct ndpi_flow_struct {
   u_int8_t bfcp_stage:1;
   u_int32_t bfcp_conference_id;
 
+  /* NDPI_PROTOCOL_CNP_IP */
+  u_int8_t cnp_stage:1;
+  u_int32_t cnp_session_id;
```
utoni: Is it really required to add 8 more bytes to the flow struct?

0xA50C1A1: > Is it really required to add 8 more bytes to the flow struct?

It's because of alignment, unfortunately.

0xA50C1A1:
> @0xA50C1A1, thanks for fixing that! Could you try to add these false positives flows to `tests/cfgs/default/pcap/false_positives.pcapng`, please?

I replaced `NDPI_SELECTION_BITMASK_PROTOCOL_V4_V6_TCP_OR_UDP_WITH_PAYLOAD_WITHOUT_RETRANSMISSION` with `NDPI_SELECTION_BITMASK_PROTOCOL_V4_V6_UDP_WITH_PAYLOAD` for now, as even checking the Session ID didn't help. I'll think of something better when I find samples of CNP-IP over TCP.

src/include/ndpi_typedefs.h
IvanNardi: I think @utoni meant if it is possible to improve the detection without any state... like checking also the version field or waiting for multiple pkts (2-3) all matching the pattern....

IvanNardi: Since the false positives seems to be only on TCP, could we avoid the state for the time being and process only UDP flow? That should be  a easy patch... 
@0xA50C1A1

0xA50C1A1: > Since the false positives seems to be only on TCP, could we avoid the state for the time being and process only UDP flow? That should be a easy patch... @0xA50C1A1

Already done.

IvanNardi: It doesn't seem so... I mean remove `cnp_stage` and `cnp_session_id` and keep only `NDPI_SELECTION_BITMASK_PROTOCOL_V4_V6_UDP_WITH_PAYLOAD`...
Does it make sense?

0xA50C1A1: > It doesn't seem so... I mean remove `cnp_stage` and `cnp_session_id` and keep only `NDPI_SELECTION_BITMASK_PROTOCOL_V4_V6_UDP_WITH_PAYLOAD`... Does it make sense?

Oh yeah, sorry, I forgot about that.

IvanNardi (approved):
Thanks

PR #2534, opened 2024-08-30
TLS: better state about handshake

IvanNardi:
Keep track if we received CH or/and SH messages: useful with unidirectional flows

### DIFF
--- a/src/lib/protocols/tls.c
+++ b/src/lib/protocols/tls.c
@@ -909,24 +909,34 @@ static int processTLSBlock(struct ndpi_detection_module_struct *ndpi_struct,
 
   switch(packet->payload[0] /* block type */) {
   case 0x01: /* Client Hello */
-  case 0x02: /* Server Hello */
+    flow->protos.tls_quic.client_hello_processed = 1;
+    flow->protos.tls_quic.ch_direction = packet->packet_direction;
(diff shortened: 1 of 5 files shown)


src/lib/protocols/tls.c
```diff
@@ -909,24 +909,34 @@ static int processTLSBlock(struct ndpi_detection_module_struct *ndpi_struct,
 
   switch(packet->payload[0] /* block type */) {
   case 0x01: /* Client Hello */
+    flow->protos.tls_quic.client_hello_processed = 1;
+    flow->protos.tls_quic.ch_direction = packet->packet_direction;
+    processClientServerHello(ndpi_struct, flow, 0);
+    ndpi_int_tls_add_connection(ndpi_struct, flow);
+
+#ifdef DEBUG_TLS
+    printf("*** TLS [version: %02X][Client Hello]\n",
```
utoni, line 918: Still not a fan of `printf` in a library even it's default-disable-`#ifdef`'ed

utoni (approved):

src/lib/protocols/tls.c
IvanNardi: Yeah, I agree. I simply tried to use current style in the surrounding code...

PR #2536, opened 2024-08-30
bins: fix `ndpi_set_bin`, `ndpi_inc_bin` and `ndpi_get_bin_value`

IvanNardi:
When the required slot is too big, use the latest/bigger available bin, not in the first one.

### DIFF
--- a/src/lib/ndpi_analyze.c
+++ b/src/lib/ndpi_analyze.c
@@ -451,7 +451,7 @@ void ndpi_set_bin(struct ndpi_bin *b, u_int16_t slot_id, u_int64_t val) {
   if(!b || !b->u.bins8 || b->num_bins == 0)
     return;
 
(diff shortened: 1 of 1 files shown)


utoni (commented):
Doesn't it make more sense to let the library crash if someone uses it the wrong way instead of modifying already doomed code? :smiley:

IvanNardi:
Current code doesn't crash; it simply puts "big" values in the first slot.
While using this code to plot some histogram with excel, I found weird that "too big/ out-of-scale" values are shown in the far left of the histogram, instead of in the far right...

utoni:
> Current code doesn't crash; it simply puts "big" values in the first slot.

Sometimes it can be useful to let a library crash, especially if some third party uses the API in the wrong way. But I just saw that there is no documentation in `src/include/ndpi_api.h`, so I guess it's fine.

utoni (approved):

PR #2553, opened 2024-09-09
Add some heuristics to detect encrypted/obfuscated/proxied TLS flows 

IvanNardi:
Based on the paper: "Fingerprinting Obfuscated Proxy Traffic with
Encapsulated TLS Handshakes".
See: https://www.usenix.org/conference/usenixsecurity24/presentation/xue-fingerprinting

Basic idea:
* the packets/bytes distribution of a TLS handshake is quite unique
* this fingerprint is still detectable if the handshake is
encrypted/proxied/obfuscated

All heuristics are disabled by default.


IvanNardi:
Set as draft because we are waiting for some other commits to be merged before it.... This way we can start triggering the CI

mmanoj:
@IvanNardi 

Thanks for this effort, I will study the mentioned paper and see how we can extend this to detect vpn and anonymizers.Please advice if you already have ideas, so I can contribute as well.

PR #2562, opened 2024-09-19
Allow IP guess before port in ndpi_detection_giveup

liwilson1:
Add dpi.guess_ip_before_port which when enabled uses classification by-ip before classification by-port.

- [check] I have signed the ntop Contributor License Agreement at https://github.com/ntop/legal/blob/main/individual-contributor-licence-agreement.md
- [check] I have read the contributing guidelines at https://github.com/ntop/nDPI/blob/dev/CONTRIBUTING.md
- [check] I have updated the documentation (in doc/) to reflect the changes made (if applicable)

Describe changes:

We'd like to be able to specify the order in which the guess is chosen in ndpi_detection_giveup. This uses an enable/disable configuration parameter that when enabled, lets the protocol choose the guess by IP first, then port. 

Currently, the behavior is port before IP, which remains the same as the default for dpi.guess_ip_before_port is "disable".


IvanNardi (changes requested):
@liwilson1, thanks for your contribution, it makes a lot of sense!
Could you please improve it:
* adding a fuzzing test for this new parameter in `fuzz/fuzz_config.cpp`? You can "copy" the code used for "dpi.guess_on_giveup" as reference around line 319, with value interval 0, 1+ 1
* adding a new test configuration under `tests/cfg` with this feature enabled? Basically with config.txt with "--cfg=dpi.guess_ip_before_port,1" and as pcap only a symbolic link to `1kxun.pcap`

Let me know if I wasn't clear!

IvanNardi (approved):

IvanNardi:
@liwilson1, thanks for this useful patch! (I simply fix the indentation before the merge)
Thank you

PR #2569, opened 2024-09-23
Add enable/disable guessing using client IP/port

liwilson1:
Add configurable options for whether to include client port or client IP in the flow's protocol guesses. This defaults to include both client port/IP if the protocol is not guessed with the server IP/port.

This is intended for when flow direction detection is enabled, so we know that sport = client port, dport = server port.

- [X] I have read the contributing guidelines at https://github.com/ntop/nDPI/blob/dev/CONTRIBUTING.md
- [X] I have updated the documentation (in doc/) to reflect the changes made (if applicable)

Describe changes:

Hi, please let me know if there is a better way to do this. We would like to disable using client IP + port when deciding the flow->guessed_protocol_id and flow->guessed_protocol_id_by_ip values. 

I've created two new config options that enable/disable using the client port/ip when the server port/ip returns unknown.


src/lib/ndpi_main.c
```diff
@@ -4394,6 +4394,8 @@ void ndpi_exit_detection_module(struct ndpi_detection_module_struct *ndpi_str) {
 static default_ports_tree_node_t *ndpi_get_guessed_protocol_id(struct ndpi_detection_module_struct *ndpi_str,
                                                                u_int8_t proto, u_int16_t sport, u_int16_t dport) {
   default_ports_tree_node_t node;
+  /* Set use_sport to config value if direction detection is enabled */
+  int use_sport = ndpi_str->cfg.direction_detect_enabled ? ndpi_str->cfg.use_client_port_in_guess : 1;
```
IvanNardi, line 4398: I think we should be coherent here....

src/lib/ndpi_main.c
```diff
@@ -7422,6 +7424,7 @@ u_int16_t ndpi_guess_host_protocol_id(struct ndpi_detection_module_struct *ndpi_
 				      struct ndpi_flow_struct *flow) {
   struct ndpi_packet_struct *packet = &ndpi_str->packet;
   u_int16_t ret = NDPI_PROTOCOL_UNKNOWN;
+  int use_client = ndpi_str->cfg.use_client_ip_in_guess;
```
IvanNardi, line 7427: ... and here. Why using `ndpi_str->cfg.direction_detect_enabled` before and not now?
I think we can remove it; what do you think?

IvanNardi (changes requested):
Overall, a good idea!
Other than the issue reported below, could you change the pcap used for the tests? We should use a trace triggering different results with and without this new configuration option

Remove 1kxun.pcap from both tests and use a link to:
* iphone.pcap for disable_use_client_ip
* bot.pcap for disable_use_client_ip

src/lib/ndpi_main.c
liwilson1: Yeah I wasn't sure what would happen if direction detection was disabled? I thought the src_port wouldn't be matched to the client port.

liwilson1:
I've assumed the bot.pcap was for disable_use_client_port

src/lib/ndpi_main.c
IvanNardi: I'm not convinced, but let keep it this way for the time being

IvanNardi:
@liwilson1, thank you (I fixed the tests while merging; the instructions I gave you were confused, sorry)

PR #2588, opened 2024-10-13
Add monitoring capability

### LINKED ISSUE #2583: Enhance STUN implementation
In STUN only one address (out of the list below) is returned

    ndpi_address_port mapped_address, peer_address, relayed_address, response_origin, other_address;

This is to request to
- return more entries
- enhance STUN extra dissection so that it could be extended to X packets in order to collect all addresses

IvanNardi:
Allow nDPI to process the entire flows and not only the first N packets. Usefull when the application is interested in some metadata spanning the entire life of the session.

As initial step, only STUN flows can be put in monitoring.

See `doc/monitoring.md` for further details.

This feature is disabled by default.

Close #2583


doc/monitoring.md
```diff
@@ -0,0 +1,23 @@
+
+# Monitoring
+
+nDPI usually needs only a few packets per flow to get full classification and to get all the required metadata/flow_risks. After that point, nDPI stops processing the flow.
+However, in some use cases, it might be useful to allow nDPI to process the *entire* flow (i.e. *all* its packets, without any limits). Some examples:
+* to extract all the Stun metadata from a Stun flow
+* to extract all the request/replay pairs from a DNS flow
+In essence, monitoring allows the application to get the same metadata, multiple times, throughout the entire life of the session.
+
+If monitoring is enable in a flow:
```
lucaderi: enable -> enabled

doc/monitoring.md
```diff
@@ -0,0 +1,23 @@
+
+# Monitoring
+
+nDPI usually needs only a few packets per flow to get full classification and to get all the required metadata/flow_risks. After that point, nDPI stops processing the flow.
+However, in some use cases, it might be useful to allow nDPI to process the *entire* flow (i.e. *all* its packets, without any limits). Some examples:
+* to extract all the Stun metadata from a Stun flow
```
lucaderi: Stun -> STUN

doc/monitoring.md
```diff
@@ -0,0 +1,23 @@
...
+
+If monitoring is enable in a flow:
+* structures `ndpi_flow->protos`, `ndpi_flow->http`, `ndpi_flow->stun`,... are populated as usual, usually with the *first* instance of the specific metadata. Nothing changed.
+* packet by packet, the new structure `ndpi_flow->monitor` is populated with the metadata of the *current* packet. This information is lost when starting processing the next packet in the same flow; it is the responsibility of the application to get it.
+
+In other words:
+* "flow metadata" is saved in `ndpi_flow->protos`, `ndpi_flow->http`, `ndpi_flow->stun`, regardless of the monitoring feature being enabled or not
+* "(curent) packet metadata" is saved in `ndpi_flow->monitor`, only if monitor is enabled
+
+Monitoring must be explicit enabled with something like: `--cfg=stun,monitoring,1`
```
lucaderi: So in essence monitoring is per protocol. Any chance to use * or NULL, for all protocols?

lucaderi: Note add --cfg=any,monitoring,1 to the manual

utoni (commented):
Thanks for that feature Ivan. That is something that Cybersecurity folks need!

Do you think that it would make sense to also provide an API in which makes it possible to dynamically add a certain flow to the "special observation treatment" and dissect literally all packets of it?

IvanNardi:
> Do you think that it would make sense to also provide an API in which makes it possible to dynamically add a certain flow to the "special observation treatment" and dissect literally all packets of it?

@utoni, could you elaborate please? Do you mean something like: the application want to put on monitoring only THIS specific STUN flow and not all of them?

utoni:
Exactly, but I would like to use that specific behavior for other protocols e.g. suspicious TLS or BitTorrent flows.

IvanNardi:
> Exactly, but I would like to use that specific behavior for other protocols e.g. suspicious TLS or BitTorrent flows.

Right now the code has been updated only for STUN; to add support for monitoring to other protocols you need to update their dissectors code.

About the specific flow. The configuration is per protocols, so all the STUN flows or none of it, but the final decision is always of the application itself: the application can simply stop invoking nDPI for the "uninteresting" STUN flows.

That is the current state; we can always improve and extend it

doc/monitoring.md
IvanNardi: Done

doc/monitoring.md
IvanNardi: Done

doc/monitoring.md
IvanNardi: Done

PR #2600, opened 2024-10-19
Ignored non IP/IPv6 packets on ndpiReader 

lucaderi:

- [X ] I have signed the ntop Contributor License Agreement at https://github.com/ntop/legal/blob/main/individual-contributor-licence-agreement.md
- [X ] I have read the contributing guidelines at https://github.com/ntop/nDPI/blob/dev/CONTRIBUTING.md
- [ X] I have updated the documentation (in doc/) to reflect the changes made (if applicable)

Describe changes:
ndpiReader does not currently discard non IP/IPv6 pockets as it should


tests/cfgs/default/result/dingtalk.pcap.out
0xA50C1A1, line 1: Снимок экрана от 2024-10-19 16-37-39

I think it's because there's no Ethernet header.

lucaderi: The eth type is wrong but underneath there is an IPv4 protocol.

@IvanNardi will address the issue this PR was trying to solve and provide a better fix.

PR #2604, opened 2024-10-21
Fix `ndpi_tot_allocated_memory` calculation if `ndpi_calloc()` used

utoni:

### DIFF
--- a/src/lib/ndpi_memory.c
+++ b/src/lib/ndpi_memory.c
@@ -64,11 +64,11 @@ void *ndpi_malloc(size_t size) {
 
 void *ndpi_calloc(unsigned long count, size_t size) {
   size_t len = count * size;
-  void *p = ndpi_malloc(len);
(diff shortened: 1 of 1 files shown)


src/lib/ndpi_memory.c
```diff
@@ -64,7 +64,7 @@ void *ndpi_malloc(size_t size) {
 
 void *ndpi_calloc(unsigned long count, size_t size) {
   size_t len = count * size;
-  void *p = ndpi_malloc(len);
+  void *p = _ndpi_malloc ? _ndpi_malloc(len) : malloc(len);
 
   if(p) {
     memset(p, 0, len);
```
IvanNardi, line 70: `len` instead of `size` on the line below?

utoni: true..

IvanNardi (approved):
Good catch!

PR #2606, opened 2024-10-24
Add Paltalk protocol support

0xA50C1A1:

Describe changes:

Although Paltalk is no longer as popular as it was in the 2000s, it still has millions of users.


src/lib/ndpi_content_match.c.inc
```diff
@@ -1727,6 +1730,7 @@ static ndpi_tls_cert_name_match tls_certificate_match [] = {
   { "O=Sonos, Inc",                    NDPI_PROTOCOL_SONOS      },  
   { "CN=Snapchat Inc.",                NDPI_PROTOCOL_SNAPCHAT_CALL },
   { "CN=NVIDIA GameStream",            NDPI_PROTOCOL_GEFORCENOW },
+  { "O=PALTALK, INC.",                 NDPI_PROTOCOL_PALTALK },
```
IvanNardi: Is that really necessary? I'd prefer to avoid adding entries in this list if not really necessary...
For example, there is TLS traffic without SNI?

0xA50C1A1: Probably not.

IvanNardi (approved):
Thanks! Sorry for the delay :)

PR #2629, opened 2024-11-26
Add a configuration file to ndpiReader

### LINKED ISSUE #2608: Add a configuration file to ndpiReader
Allow to put ndpiReader options into a text file.
Something like: `ndpiReader --cfg-file=options.txt -i file.pcap`
where, for example:
```
cat options.txt
-I -d -U -p file.protos
```
Note that the *library* can be already configured via a text file; this ticket is about `ndpiReader` options

IvanNardi:
Example:
./example/ndpiReader --conf=./example/calls.conf -i ./tests/pcap/signal_videocall.pcapng -v2

Close #2608

### DIFF
--- a/example/Makefile.in
+++ b/example/Makefile.in
@@ -96,6 +96,7 @@ distdir:
 		-o -name '*.h' \
 		-o -name '*.py' \
 		-o -name '*.txt' \
+		-o -name '*.conf' \
 		-o -name '*.csv' | xargs -I'{}' cp '{}' '$(distdir)/{}'
 
 distclean: clean
--- a/example/calls.conf
+++ b/example/calls.conf
@@ -1,0 +1,8 @@
+#Useful ndpiReader configuration to analyse audio/video calls traffic
+
+#Generic limits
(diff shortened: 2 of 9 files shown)


utoni (commented):
Is it necessary that all config file options need to start with `--cfg=`? A more intuitive config file would be something like `key.subkey = value`.

IvanNardi:
> Is it necessary that all config file options need to start with `--cfg=`? A more intuitive config file would be something like `key.subkey = value`.

The big advantage of the proposed solution is to keep the same format between `ndpiReader` command line options and the configuration file: you can simply copy the string between them.

BTW, not ALL config file options start with `--cfg=` but only the libnDPI ones; `ndpiReader` options are different (for example `-U 0 -p proto.file -P 4:8:10:16:25`) but you can copy also the latter as-is into the configuration file.

In other words, for me, the goal of this file is not to be automatically generated, but simply to provide you an easy way to collect all your options together  and call them easily via command line

utoni (approved):

PR #2649, opened 2024-12-19
QUIC: extract "max idle timeout" parameter

IvanNardi:
Even if it is only the proposed value by the client (and not the negotiated one), it might be use as hint for timeout by the (external) flows manager


utoni (approved):
Interesting. Did not know about `ICSL`.
https://datatracker.ietf.org/doc/html/draft-ietf-quic-transport-00 gave me some context

PR #2660, opened 2025-01-07
Update some CI jobs

IvanNardi:
* Move ThreadSanitizer job to the scheduled jobs (once a day): all our tests are intrinsically mono-thread and this job takes quite some time
    
* Two explicit jobs to test LTO and Gold linker, used by oss-fuzz
    
* Two explicit jobs for Windows (with msys2)
    
* Run address sanitizer only on the 4 main jobs: newest/oldest gcc/clang
    
* Reduce the time used by fuzzing jobs. Note that oss-fuzz is continuosly fuzzing our code!
    
* Move the no x86_64 jobs to a dedicated file
    
This way, the main matrix is a little bit simpler and the CI jobs last a little shorter

### DIFF
--- a/.github/workflows/cifuzz.yml
+++ b/.github/workflows/cifuzz.yml
@@ -18,7 +18,7 @@ jobs:
         uses: google/oss-fuzz/infra/cifuzz/actions/run_fuzzers@master
         with:
           oss-fuzz-project-name: 'ndpi'
-          fuzz-seconds: 1200
+          fuzz-seconds: 600
           dry-run: false
           sanitizer: ${{ matrix.sanitizer }}
       - name: Check Crash (fails when a crash is detected)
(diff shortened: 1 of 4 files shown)


.github/workflows/cifuzz.yml
```diff
@@ -18,7 +18,7 @@ jobs:
         uses: google/oss-fuzz/infra/cifuzz/actions/run_fuzzers@master
         with:
           oss-fuzz-project-name: 'ndpi'
-          fuzz-seconds: 1200
+          fuzz-seconds: 600
```
utoni, line 21: Does this influence fuzzing time within `oss-fuzz`?

IvanNardi: No, it only changes the duration of "our" fuzzing in the GitHub Actions

utoni (approved):
Great work. Separating x64 and !x64 builds definitely improves readability.

PR #2661, opened 2025-01-08
Improve documentation

IvanNardi:
Integrate .md files into official documentation
See: https://stackoverflow.com/questions/46278683/include-my-markdown-readme-into-sphinx/68005314#68005314


utoni:
Do you think it's useful to have a CI job that generates the documentation?

IvanNardi:
No I don't. But since we already have it, it might be useful to test that
the documentation is properly built. Doing that every day is overwhelming,
tough, imo

PR #2666, opened 2025-01-11
HTTP: fix entropy calculation

IvanNardi:
We calculate HTTP entropy according to "Content-type:" header, see `ndpi_validate_http_content()` on HTTP code


utoni (approved):

utoni:
makes sense

Issue #2551, opened 2024-09-07
Remove JA3C/S fingerprints


IvanNardi:
Since we have support for the new fingerprint JA4, should we remove the oldest version?

IvanNardi:
See for example https://suricon.net/wp-content/uploads/2024/12/SuriCon2024-Eric-Leblond_Google-killed-JA3-should-we-be-scared.pdf for an explanation why JA3C is no more useful

IvanNardi:
The current idea is:
* remove JA3C: we already have JA4C
* keep JA3S for the time being: we don't have another server fingerprint right now, and if we create a new one, having that code around will ease the transaction from JA3S to the new finferprint

PR #2667, opened 2025-01-11
Remove JA3C output from ndpiReader

IvanNardi:
Removing JA3C is an big task. Let's start with a simple change having an huge impact on unit tests: remove printing of JA3C information from ndpiReader.

This way, when we will delete the actual code, the unit tests diffs should be a lot simpler to look at.

Note that the information if the client/server cipher is weak or obsolete is still available via flow risk

See: #2551


utoni (approved):
:+1:

PR #2669, opened 2025-01-11
Fix compilation on latest mac versions with external libraries

### LINKED ISSUE #2412: Compilation errors on MacOS-14 on ARM64 with external libraries
See:
https://github.com/ntop/nDPI/commit/8fbef7fb5ab682b7aa9c92e4fd6e106698eb4ae1
https://github.com/ntop/nDPI/commit/9e549ba781a87cb0131c439939e27fd5fd050f10

IvanNardi:
Fix also Performance job on ubuntu-latest/24.04: see similar fix in 957a05050.
    
Close: #2412

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -76,9 +76,8 @@ jobs:
     strategy:
       fail-fast: true
       matrix:
-        # macOS-latest == macos-14 on **ARM64**. Also macos-15 is on arm64
-        # There are some issues with external dependencies on macOS-14/15. Disable it for the time being
-        os: ["ubuntu-20.04", "ubuntu-22.04", "ubuntu-24.04", "macOS-13"]
+        # macos-14 and 15 are on **ARM64**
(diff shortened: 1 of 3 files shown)


.github/workflows/build.yml
```diff
@@ -215,18 +200,10 @@ jobs:
...
-        if: ${{ !startsWith(matrix.os, 'macos-14') && !startsWith(matrix.os, 'macos-15') }}
         run: |
           make -j all
           make -C example ndpiSimpleIntegration
           make -C rrdtool
-      - name: Build nDPI (MacOS M1)
-        if: startsWith(matrix.os, 'macos-14') || startsWith(matrix.os, 'macos-15')
-        run: |
-          make -j all
-          make -C example ndpiSimpleIntegration
```
utoni, line 227: You do not want to build it at least once on Mac?

IvanNardi: macos builds use the standard step "Build nDPI" just above, as well as ubuntu builds; no more need for a specific step....
We will build on macos-13,14,15 (standard compilation) and on latest with some external dependencies

PR #2680, opened 2025-01-14
ndpiReader: add some global statistics about FPC

IvanNardi:
Enabled via `--dump-fpc-stats` option


IvanNardi:
Screenshot from 2025-01-14 18-14-20

PR #2723, opened 2025-02-14
reworked ntp info extraction

koltiradw:

Describe changes:
Updated ntp fields by header format.
image

### DIFF
--- a/src/include/ndpi_typedefs.h
+++ b/src/include/ndpi_typedefs.h
@@ -1439,8 +1439,8 @@ struct ndpi_flow_struct {
     } dns;
 
     struct {
-      u_int8_t request_code;
       u_int8_t version;
+      u_int8_t mode;
     } ntp;
 
     struct {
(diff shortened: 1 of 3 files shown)


src/include/ndpi_typedefs.h
```diff
@@ -1439,8 +1439,8 @@ struct ndpi_flow_struct {
     } dns;
 
     struct {
-      u_int8_t request_code;
       u_int8_t version;
+      u_int8_t mode;
```
IvanNardi, line 1443: It seems reasonable.
We lost the `request_code` attribute for the old v2, but we now have correct `version` and `mode` for all 3 versions

IvanNardi: If someone is really interested in that attribute for v2, we can re-add it later

IvanNardi (approved):
LGTM, thanks

PR #2768, opened 2025-03-14
NTP metadata export

koltiradw:


IvanNardi:
Any updates on this? The features seems interesting, but It seems abandoned, too...

koltiradw:
@IvanNardi, hi!
I've been busy for a long time, but I'm ready to do it again this week.

koltiradw:
@IvanNardi 
I think the PR is ready.

src/include/ndpi_typedefs.h
```diff
@@ -1716,10 +1719,16 @@ struct ndpi_flow_struct {
...
+    struct ntp_info {
+      u_int8_t leap_indicator: 2, version: 3, mode: 3;
+      u_int8_t stratum;
+      int8_t ppol, precision;
+      float root_delay, root_dispersion;
+      char ref_id[20];
+      uint64_t ref_time, org_time, rec_time, trans_time;
+    } ntp_info;
+
+    struct ntp_info ntp[2];
```
IvanNardi: Maybe defining `struct ntp_info` outside `struct ndpi_flow_struct`? We want two instances, not three

src/lib/protocols/ntp.c
```diff
@@ -29,34 +29,94 @@
...
-    uint8_t version = (packet->payload[0] & 56) >> 3;
-
-    if (version <= 4) {
-      flow->protos.ntp.version = version;
-      flow->protos.ntp.mode = packet->payload[0] & 7;
+    get_ntp_flags(flow, packet, flow->l4.udp.ntp_stage);
     
+    if (flow->protos.ntp[flow->l4.udp.ntp_stage].version <= 4) {
+      get_ntp_info(flow, packet, flow->l4.udp.ntp_stage);
+      flow->l4.udp.ntp_stage = 1;
```
IvanNardi: You must pay attention here. If you set `flow->protos.ntp` (via `get_ntp_flags()`) you MUST set the protocol.
In other word, you can write/red to `flow->protos.ntp` ONLY if you are going to set the protocol in the same packet (or if you already set it in previous packets).
Here if `if (flow->protos.ntp[flow->l4.udp.ntp_stage].version <= 4)` is false, you write to the union, but you didn't set the protocol, IMO, and that is wrong

IvanNardi (changes requested):
Sorry for the delay.
Could you take a look at these comments, please?

src/include/ndpi_typedefs.h
IvanNardi: Or, better, you can leave inside, with something like
```
    struct ntp_info {
      u_int8_t leap_indicator: 2, version: 3, mode: 3;
      u_int8_t stratum;
      int8_t ppol, precision;
      float root_delay, root_dispersion;
      char ref_id[20];
      uint64_t ref_time, org_time, rec_time, trans_time;
    } ntp[2];
```

koltiradw:
I thought it might make sense to get only the flags and ref_id for ntp versions 3 and 4, and only the flags for version 2.
 If you have any ideas on how to use other fields, then I won't change anything.

IvanNardi (approved):
Thanks for not abandoning this task!

PR #2773, opened 2025-03-19
Rework the old MapleStory code to identify traffic from generic Nexon games

IvanNardi:
Remove `NDPI_PROTOCOL_MAPLESTORY` and add a generic `NDPI_PROTOCOL_NEXON`


IvanNardi:
Note that we have never had a pcap matching the old patterns...

IvanNardi:
Thanks, @0xA50C1A1

PR #2776, opened 2025-03-25
Rework the old Starcraft code to identify traffic from generic Blizzard games

IvanNardi:
Remove `NDPI_PROTOCOL_STARCRAFT` and add a generic `NDPI_PROTOCOL_BLIZZARD`.


IvanNardi:
@0xA50C1A1, what do you think?

0xA50C1A1:
> @0xA50C1A1, what do you think?

I think it should work for Warcraft 3 Reforged as well. I'll check it out when I get a chance.

PR #2799, opened 2025-04-22
Add Microsoft Delivery Optimization protocol

0xA50C1A1:

Describe changes:

Added detection for Microsoft Delivery Optimization, an internal Windows protocol for sharing update packages between Windows peers.

I used the existing `NDPI_PROTOCOL_WINDOWS_UPDATE` protocol ID since Delivery Optimization is related, but I'm not sure if that's a good approach - maybe I should have added a new ID?


IvanNardi:
> 
> I used the existing `NDPI_PROTOCOL_WINDOWS_UPDATE` protocol ID since Delivery Optimization is related, but I'm not sure if that's a good approach - maybe I should have added a new ID?

@0xA50C1A1, sorry for the delay. I think it is best to use a new id. Could you add it (with a new entry in the documentation), please?

0xA50C1A1:
> > I used the existing `NDPI_PROTOCOL_WINDOWS_UPDATE` protocol ID since Delivery Optimization is related, but I'm not sure if that's a good approach - maybe I should have added a new ID?
> 
> @0xA50C1A1, sorry for the delay. I think it is best to use a new id. Could you add it (with a new entry in the documentation), please?

Done.

IvanNardi (approved):
Thank you

PR #2819, opened 2025-05-16
Remove Half-Life 2 support; improve Source Engine protocol detection

0xA50C1A1:

Describe changes:

According to the [official Steam news](https://store.steampowered.com/news/app/320/view/529838805680652975), Half-Life 2 Deathmatch has transitioned to using **Steam Networking** (Steam Datagram Relay) since February 2025. Importantly, the original Half-Life 2 did not support multiplayer, so the old protocol is essentially obsolete.

This is further confirmed in the [Valve developer wiki](https://developer.valvesoftware.com/wiki/Source_Multiplayer_Networking#Steam_Networking), which states that Steam Networking is the current standard for multiplayer in Source Engine games.

I also reworked the Source Engine protocol dissector (it's actually called A2S) based on information from Valve's developer wiki. Although Valve is slowly dropping this protocol in favor of Steam Datagram Relay, it is still used in games such as Arma 3, ARK: Survival Evolved, and 7 Days to Die.


IvanNardi (approved):
LGTM

Issue #2827, opened 2025-05-19
Some games not so popular/used anymore....


IvanNardi:
* NDPI_PROTOCOL_DOFUS — this MMORPG still exists, albeit with reduced popularity. Updating the dissector would be worthwhile, but I don’t have any pcap samples to work with.

* NDPI_PROTOCOL_GUILDWARS — GW1 servers are still running, though it would be great to add GW2 traffic detection. I have pcap samples, but couldn’t find any clear signatures so far.

* NDPI_PROTOCOL_CROSSFIRE — while EU servers have closed, the game remains popular in Asia. Updating the dissector would be useful, though I lack traffic samples. Also, there’s a mobile version available, but only in Chinese, as far as I know.

From https://github.com/ntop/nDPI/pull/2826#issuecomment-2890679970, credits to @0xA50C1A1

IvanNardi:
I don't have strong opinions on these 3 ids... @utoni: what do you think? Keep them? Delete them?

IvanNardi:
I would also add NDPI_PROTOCOL_ARMAGETRON: I updated the code a few weeks ago, but I am not sure if this game still "deserves" to be supported....

0xA50C1A1:
> I would also add NDPI_PROTOCOL_ARMAGETRON: I updated the code a few weeks ago, but I am not sure if this game still "deserves" to be supported....

I don't have a strong opinion about it either.

On the one hand this game is very niche (only 194 members in the [Reddit community](https://www.reddit.com/r/armagetronad/) at the moment), but it's not abandoned and is available on Steam as [Retrocycles](https://store.steampowered.com/app/1306180/Retrocycles/).

IvanNardi:
Ok, so something like: keep only NDPI_PROTOCOL_ARMAGETRON because the code works with current game version and remove the others for the opposite reason: current code doesn't classify latest games versions.
If someone is still interested, we can work on those three later

0xA50C1A1:
Hi @IvanNardi, @utoni,

I’ve recently updated the CrossFire and GuildWars dissectors (#2834, #2836). However, I’m currently stuck with the Dofus dissector because I’m unable to play this game due to regional restrictions.

Could either of you take a look and consider updating the Dofus dissector? If not, perhaps we should consider dropping support for it.

IvanNardi:
> Hi [@IvanNardi](https://github.com/IvanNardi), [@utoni](https://github.com/utoni),
> 
[...]

PR #2836, opened 2025-05-21
Drop GW1 support and add basic GW2 detection

0xA50C1A1:

Link to the related [issue](https://github.com/ntop/nDPI/issues): #2827

Describe changes:

Dropped Guild Wars 1 support since I couldn't properly verify if the dissector still works - servers might be up, but the protocol could have changed.

Instead, added basic support for Guild Wars 2, which is still actively played and should be more relevant.


IvanNardi (approved):
Thank you

PR #2843, opened 2025-05-21
A new interface for dissectors registration

### LINKED ISSUE #2136: Increase the number of protocols supported
Due to the old OpenDPI code, supporting many new protocols is a problem. This code needs to be rewritten in order to make it more modern.

starting ndpiReader with the enclosed protocol file reproduces the problem.
[ndpiProtocol.txt](https://github.com/ntop/nDPI/files/13280033/ndpiProtocol.txt)

IvanNardi:
We use `registr_dissector()` instead of `ndpi_set_bitmask_protocol_detection()`.

Every file in `src/lib/protocols/*.c` is a dissector.

Every dissector can handle multiple protocols.

The real goal is this small change:
```
 struct call_function_struct {
-  NDPI_PROTOCOL_BITMASK detection_bitmask;
```
i.e. getting rid of another protocol bitmask: this is mandatory to try to fix #2136 (see also e845e8205b68752c997d05224d8b2fd45acde714)

As a nice side effect, we remove a bitmask comparison in the hot function
`check_ndpi_detection_func()`

TODO: change logging configuration from per-protocol to per-dissector

### DIFF
--- a/src/include/ndpi_api.h
+++ b/src/include/ndpi_api.h
@@ -36,11 +36,6 @@ extern "C" {
 #define NDPI_DETECTION_ONLY_IPV4              ( 1 << 0 )
 #define NDPI_DETECTION_ONLY_IPV6              ( 1 << 1 )
 
-#define ADD_TO_DETECTION_BITMASK              1
-#define NO_ADD_TO_DETECTION_BITMASK           0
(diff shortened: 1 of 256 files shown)


0xA50C1A1:
That's an impressive work

PR #2849, opened 2025-05-23
Add ndpi_memcasecmp, refactor mail protocol dissectors

0xA50C1A1:

Describe changes:

Spotted a potential overflow of `u_int32_t mail_imap_stage:3` in the IMAP dissector: it was being incremented a bunch without bounds checking. Never triggered on CI but doesn’t mean it’s safe, so I fixed that.

While I was there, saw a lot of manual case-insensitive string checks across IMAP, POP, and SMTP. Wrote a little utility: `ndpi_memcasecmp` and refactored the dissectors to use it. Should make this a bit cleaner and less error-prone.


src/lib/protocols/mail_imap.c
```diff
@@ -29,6 +29,15 @@
 #include "ndpi_api.h"
 #include "ndpi_private.h"
 
+/* Safely increments IMAP stage counter preventing 3-bit mail_imap_stage overflow.
+ * Even though current tests don't trigger overflow, better safe than sorry. */
+#define NDPI_SAFE_INC_IMAP_STAGE(flow) \
```
IvanNardi: Just a minor issue. Could you remove the NDPI prefix here? I prefer to keep it only for the public API (at least, for the new functions/defines...)

IvanNardi (approved):
LGTM!

PR #2854, opened 2025-05-26
BFCP: fix check on payload length and extract metadata

### LINKED ISSUE #2745: BFCP pcap
@0xA50C1A1, is it possible that the file `tests/cfgs/default/pcap/bfcp.pcapng` contains malformed/invalid packets? May it be some kind of "artificial/handmade" traffic?

I am trying to improve BFCP detection and this traffic looks invalid to me....

IvanNardi:
We should be able to identified this protocol on the first packet, without keeping any state

Close #2745


IvanNardi:
@0xA50C1A1, could you take a look, please?

0xA50C1A1:
> @0xA50C1A1, could you take a look, please?

Perfect. What do you think about extracting the user ID and conference ID as metadata, could that be useful?

src/lib/protocols/bfcp.c
```diff
@@ -45,7 +45,13 @@ static void ndpi_search_bfcp(struct ndpi_detection_module_struct *ndpi_struct,
   u_int8_t version = (packet->payload[0] >> 5) & 0x07;
   u_int8_t reserved = (packet->payload[0] >> 3) & 0x01;
```
0xA50C1A1: Should we check reserved bits?

IvanNardi:
> > @0xA50C1A1, could you take a look, please?
> 
> Perfect. What do you think about extracting the user ID and conference ID as metadata, could that be useful?

I don't know, but it would be trivial to do. I'll push a new version

src/lib/protocols/bfcp.c
IvanNardi: I think so, but this check is wrong; we should check last 3 bits...

PR #2860, opened 2025-05-30
Add Hamachi protocol detection support

0xA50C1A1:

Describe changes:

While Hamachi isn't as widely used as it was in its 2010s heyday, it's still actively maintained by LogMeIn (acquired by GoTo in 2020) with paid subscriptions available.

Note: The TCP implementation currently only detects service traffic (via initial packet signature). Full TCP session detection could be added later - I've included sample pcaps in the attached archive for anyone interested in expanding this.

[hamachi.zip](https://github.com/user-attachments/files/20523913/hamachi.zip)

Traffic characteristics observed:
- UDP: First 4 bytes and bytes 8-9 remain constant per session but differ between peers. Bytes 4-7 appear to be a packet counter (increments by 1 in subsequent packets).
- TCP: First 4 bytes contain (packet_length-4) in big-endian format.


src/lib/protocols/hamachi.c
```diff
@@ -0,0 +1,181 @@
...
+    /* Both directions verified - final consistency check before protocol identification */
+    if (hamachi_l != flow->l4.udp.hamachi_long[dir] ||
+        hamachi_s != flow->l4.udp.hamachi_short[dir])
+    {
+      goto exclude_hamachi;
+    }
+
+    ndpi_int_hamachi_add_connection(ndpi_struct, flow);
+    return;
+  }
```
0xA50C1A1: @IvanNardi I'm not entirely sure about the state machine logic - would appreciate your feedback on this implementation.

IvanNardi (approved):
LGTM

src/lib/protocols/hamachi.c
```diff
@@ -0,0 +1,171 @@
...
+  NDPI_LOG_DBG(ndpi_struct, "search Hamachi over UDP\n");
+
+  /* Session starts with two such packets containing 4 bytes in 
+   * the middle - probably key exchange or something. Skip.
+   */
+  if (packet->payload_packet_len == 76 &&
+      get_u_int64_t(packet->payload, 0) == 0)
+  {
+    return;
+  }
```
IvanNardi, line 74: @0xA50C1A1, I am not sure If I am getting that right...
**IF** the first two packets should always have length 76 and zero at the beginning, couldn't we exclude this protocol if this condition is not true for the first 2 packets?

0xA50C1A1: @IvanNardi Yeah, you're right! By what I've seen, when using a relay, the session always starts with two packets like that - 76 bytes in length, with 4 bytes in the middle and the rest filled with zeros.

Initially, I didn't exclude the protocol here because I thought the code might also work with P2P traffic (where the session starts differently), but it turns out the packet format is different for P2P so this code doesn't really fit there.

So yes, we can safely exclude this protocol if this condition doesn't hold for the first two packets.

0xA50C1A1: The reason I used return is that those initial packets were messing up the detection logic - цу actually need to compare bytes 0–3 and 8–9 in both directions only **after** those two packets.

@IvanNardi what do you think about this solution?

```
  if (packet->payload_packet_len == 76)
  {
    if (get_u_int64_t(packet->payload, 0) != 0 &&
        get_u_int64_t(packet->payload, 68) != 0)
    {
      goto exclude_hamachi;
    }
    return;
  }
```

0xA50C1A1: The reason I used return is that those initial packets were messing up the detection logic - we actually need to compare bytes 0–3 and 8–9 in both directions only **after** those two packets.

@IvanNardi what do you think about this solution?

```
  if (packet->payload_packet_len == 76)
  {
    if (get_u_int64_t(packet->payload, 0) != 0 &&
        get_u_int64_t(packet->payload, 68) != 0)
    {
      goto exclude_hamachi;
    }
    return;
  }
```

IvanNardi: Something like (untested):
```
if(flow->packet_counter <=2) {
   if(packet->payload_packet_len == 76 &&
       get_u_int64_t(packet->payload, 0) != 0 &&
       get_u_int64_t(packet->payload, 68) != 0)) {
    return;
  }
  exclude;
}
```
? @0xA50C1A1

PR #2861, opened 2025-06-02
Hamachi: improve handshake check

0xA50C1A1:

Describe changes:

Follow up #2860


src/lib/protocols/hamachi.c
```diff
@@ -64,13 +64,15 @@ static void search_hamachi_udp(struct ndpi_detection_module_struct* ndpi_struct,
...
-    return;
+  /* Skip initial 76-byte handshake (relay mode only) */
+  if (flow->packet_counter <= 2 && packet->payload_packet_len == 76) {
+    /* Quick zero-check for first 8 and last 8 bytes */
+    if (get_u_int64_t(packet->payload, 0) != 0 && 
+        get_u_int64_t(packet->payload, 68) != 0)
+    {
+      goto exclude_hamachi;
+    }
+    return; /* Likely Hamachi handshake */
```
IvanNardi, line 75: I think that we should exclude it also when the packet length of the first two packets is not 76
```
if (flow->packet_counter <= 2) {
  if(packet->payload_packet_len == 76 &&
     get_u_int64_t(packet->payload, 0) == 0 && 
     get_u_int64_t(packet->payload, 68) == 0)
    return;  /* Likely Hamachi handshake */
  exclude;
}
```

0xA50C1A1: Maybe. However, in the case of a direct connection between peers (i.e. without relay), the handshake packets are 72 bytes long and contain no zeros.

That said, I didn't consider this scenario due to its rarity - after all, who needs Hamachi if the peers are already on the same network? Additionally, the first 18 packets differ by 8-9 bytes.

[hamachi_direct.zip](https://github.com/user-attachments/files/20556884/hamachi_direct.zip)

0xA50C1A1: By the way, won't this break the mid-flow detection?

IvanNardi: I think a generic mid-flow detection doesn't work already, because of the check on the sequence number. Am I wrong?

0xA50C1A1: > I think a generic mid-flow detection doesn't work already, because of the check on the sequence number. Am I wrong?

I've never seen the sequence number go beyond 0xFFFF. By the way, each direction has its own sequence number.

I picked a few packets from the mid-flow, they are detected as Hamachi just fine.

[hamachi_midflow.zip](https://github.com/user-attachments/files/20566210/hamachi_midflow.zip)

IvanNardi: Ok, you convinced me :) let's try to detect mid-flows....

In the end, is that PR still needed? I got lost... :D @0xA50C1A1

0xA50C1A1: > Ok, you convinced me :) let's try to detect mid-flows....
> 
> In the end, is that PR still needed? I got lost... :D @0xA50C1A1

Well, we can get it back the way it was.

IvanNardi (approved):

IvanNardi:
@0xA50C1A1, thanks for your explanations

Issue #2594, opened 2024-10-17
'ndpi_add_host_url_subprotocol', need custom category and breed


UnveilTech:
Hello Ivan,

proposal of improvment:
to be able to define a category and breed with the function 'ndpi_add_host_url_subprotocol'
- file: _ndpi_main.c_
- main function: _static int ndpi_handle_rule(struct ndpi_detection_module_struct *ndpi_str, char *rule)_
- used function: _ndpi_add_host_url_subprotocol(ndpi_str, value, subprotocol_id, NDPI_PROTOCOL_CATEGORY_UNSPECIFIED, NDPI_PROTOCOL_ACCEPTABLE, 0);_

it should be fine to use category and breed we need instead of **NDPI_PROTOCOL_CATEGORY_UNSPECIFIED** and **NDPI_PROTOCOL_ACCEPTABLE** as by default here.

what do you think ?

we understand it could modify the format of the proto.txt, but it'll be a great +

bye Fred

0xA50C1A1:
I've tried to implement this before, but didn't create a PR. I think the main issue with the new proto.txt format is: how to add category and breed in a way that doesn't make it ambiguous?

UnveilTech:
Hi Vladimir,
quick examples:
'host:"api-global.netflix.com"@Netflix=|Network|Safe'
'host:"api-global.netflix.com"@Netflix=1024|Network|Safe'
'host:"api-global.netflix.com"@Netflix=|Network'
'host:"api-global.netflix.com"@Netflix=1024||Safe'
'host:"api-global.netflix.com"@Netflix=1024,Network,Safe'
'host:"api-global.netflix.com"@Netflix,Network,Safe'  (no number after the CustomProtocol means the next part is a category)
etc...

there are dozens of scenarios here.
you can use the mark you think it could be the best or the easiest to cut the string... :o)

another nice way could be to use a json format (ex.: proto.json) to define each host/values... maybe more complicated but much more powerful.

bye Fred

IvanNardi:
I don't have preferences... I'll even agree with something really simple as:
```
host:"api-global.netflix.com"@Netflix=1024,5             #Category 5
host:"api-global.netflix.com"@Netflix,5                        #Category 5
host:"api-global.netflix.com"@Netflix=1024,5 ,3         #Category 5, Breed 3. If you want to specify a breed you need to specify a category too, even if 0
host:"api-global.netflix.com"@Netflix,5,3                     #Category 5, Breed 3.
```
[...]

PR #2872, opened 2025-06-06
Add category and breed support for custom rules

0xA50C1A1:

Link to the related [issue](https://github.com/ntop/nDPI/issues): #2594

Describe changes:

@IvanNardi @UnveilTech, could you please review this? 

Added category and breed support for all @-rules as requested in the issue. Works fine on my machine, but want to make sure I didn't mess anything up.


IvanNardi (changes requested):
Could you add some examples in protos.txt?
Changing the CustomProtocolD/E/F/... rules to add all possible combinations

UnveilTech:
Hi Ivan,

"could you please review this?
Added category and breed support for all @-rules as requested in the issue. Works fine on my machine, but want to make sure I didn't mess anything up."

not sure to get what you mean... github processes are not familiar to us, please explain, thanks
bye Fred

example/protos.txt
```diff
@@ -63,6 +63,22 @@ ipv6:[fe80::76ac:b9ff:fe6c:c124]:12717@CustomProtocolG=2050
...
+host:"api.example.com"@CustomProtocolN=4000,cat=5,breed=2
+host:"*.test.example.org"@CustomProtocolO=4001,cat=101,breed=6
+
+ip:192.0.2.1:443@CustomProtocolP=5000,cat=14,breed=0
+ip:203.0.113.0/24@CustomProtocolQ=5001,cat=107,breed=3
+
+
+tcp:9999@CustomProtocolR=6000,cat=8,breed=3
+udp:53239@CustomProtocolS=6001,cat=14,breed=0
+
```
IvanNardi, line 70: @0xA50C1A1 , could you add one pcap with one flow/pkt for *each* of these rules, please?
I would like some basic tests for this new feature

0xA50C1A1: > @0xA50C1A1 , could you add one pcap with one flow/pkt for _each_ of these rules, please? I would like some basic tests for this new feature

Okay. By the way, I had to roll back the string-based `cat=` and `breed=` parsing because it was causing issues with special characters (like `/`) and spaces in the rules. The original code replaces these with underscores, which breaks string matching for some category/breed names.

IvanNardi: @0xA50C1A1, are you saying that there are characters in the category/breed names that we should avoid?

0xA50C1A1: > @0xA50C1A1, are you saying that there are characters in the category/breed names that we should avoid?

Yeah, there are a couple of examples.

Breed names:

```
  case NDPI_PROTOCOL_POTENTIALLY_DANGEROUS:
    return("Potentially Dangerous");
  case NDPI_PROTOCOL_TRACKER_ADS:
    return("Tracker/Ads");
```

Categories:

```
  "Blog/Forum",
  "CNR/Proxy",
  "Hw/Sw",
```

I noticed that protocol names don't use such characters.

IvanNardi: If it simplify this patch, we can definitely remove "/" and " " from categories and breeds names... (in a separate PR)

0xA50C1A1:
@IvanNardi I added a pcap sample and fixed a couple bugs. Can you please take a look? Maybe I missed something.

IvanNardi (approved):
Great work

IvanNardi:
I added some examples for the fuzzing

PR #2908, opened 2025-06-29
Refactor: make src_name/dst_name dynamically allocated to reduce RAM usage in struct ndpi_flow_info

fabiodepin:

- [ X ] I have signed the ntop Contributor License Agreement at https://github.com/ntop/legal/blob/main/individual-contributor-licence-agreement.md
- [ X ] I have read the contributing guidelines at https://github.com/ntop/nDPI/blob/dev/CONTRIBUTING.md

Link to the related [issue](https://github.com/ntop/nDPI/issues): -

Describe changes:
- Changed ndpi_flow_info: replaced fixed-size char arrays (always INET6_ADDRSTRLEN) for src_name and dst_name with char* pointers.
- Now IPv4 flows use only INET_ADDRSTRLEN when needed, instead of always reserving IPv6 size.
- Updated ndpi_patchIPv6Address to handle pointer safely.

Impact:
- Saves up to ~92 bytes per flow when reverse lookup is not used.
- In pure IPv4 scenarios this avoids overallocating IPv6 buffer unnecessarily.

### DIFF
--- a/example/reader_util.c
+++ b/example/reader_util.c
@@ -207,8 +207,10 @@ void ndpi_payload_analyzer(struct ndpi_flow_info *flow,
 #ifdef DEBUG_PAYLOAD
     printf("[hashval: %u][proto: %u][vlan: %u][%s:%u <-> %s:%u][direction: %s][payload_len: %u]\n",
 	   flow->hashval, flow->protocol, flow->vlan_id,
-	   flow->src_name, flow->src_port,
-	   flow->dst_name, flow->dst_port,
+     flow->src_name ? flow->src_name : "",
+     flow->src_port,
+     flow->dst_name ? flow->dst_name : "",
+     flow->dst_port,
 	   src_to_dst_direction ? "s2d" : "d2s",
 	   payload_len);
 #endif
@@ -602,6 +604,8 @@ void ndpi_flow_info_free_data(struct ndpi_flow_info *flow) {
   ndpi_free_bin(&flow->payload_len_bin);
 #endif
 
+  if(flow->src_name)        ndpi_free(flow->src_name);
+  if(flow->dst_name)        ndpi_free(flow->dst_name);
   if(flow->tcp_fingerprint) ndpi_free(flow->tcp_fingerprint);
   if(flow->risk_str)        ndpi_free(flow->risk_str);
   if(flow->flow_payload)    ndpi_free(flow->flow_payload);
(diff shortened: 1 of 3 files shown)


example/reader_util.c
```diff
@@ -913,18 +917,37 @@ static struct ndpi_flow_info *get_ndpi_flow_info(struct ndpi_workflow * workflow
...
-                  newflow->src_name, sizeof(newflow->src_name));
-        newflow->dst_ip6 = *(struct ndpi_in6_addr *)&iph6->ip6_dst;
-        inet_ntop(AF_INET6, &newflow->dst_ip6,
-                  newflow->dst_name, sizeof(newflow->dst_name));
-        /* For consistency across platforms replace :0: with :: */
-        ndpi_patchIPv6Address(newflow->src_name), ndpi_patchIPv6Address(newflow->dst_name);
+      if (version == 4 || version == 6) {
+        uint16_t needed_len = (version == 4) ? INET_ADDRSTRLEN : INET6_ADDRSTRLEN;
+        if (!newflow->src_name || newflow->src_name_len != needed_len) {
+          if (newflow->src_name) ndpi_free(newflow->src_name);
```
IvanNardi: I do like the general idea behind the patch!
However, are these check really necessary? I *think* this code is triggered only once per flow, when the flow is created... could you double check please?

src/lib/ndpi_utils.c
```diff
@@ -911,24 +911,40 @@ char* ndpi_ssl_version2str(char *buf, int buf_len,
...
+ * This function normalizes an IPv6 address string that was generated
+ * by inet_ntop(), replacing any occurrence of ":0:" with "::" if no "::"
+ * is already present. This is a legacy fix for systems or cases where
+ * zero compression might not be fully applied.
+ *
+ * @param str Pointer to a null-terminated IPv6 address string buffer.
+ *            Must be allocated and writable.
+ */
 void ndpi_patchIPv6Address(char *str) {
-  int i = 0, j = 0;
```
IvanNardi: Why did you change this function? Could you elaborate please?

example/reader_util.c
fabiodepin: Hello,

Thanks for the feedback! 
Actually, without this check the fuzz test was crashing with a segfault, probably because it simulates a version value different from 4 or 6. So this condition avoids invalid memory operations in those edge cases.

I still have the log here:
```
==47==ERROR: MemorySanitizer: SEGV on unknown address 0x000000000000 (pc 0x7fd8fb3b3178 bp 0x000000000000 sp 0x7ffc7c3c3ab8 T47)
==47==The signal is caused by a WRITE memory access.
==47==Hint: address points to the zero page.
    #0 0x7fd8fb3b3178  (/lib/x86_64-linux-gnu/libc.so.6+0x18a178) (BuildId: 0323ab4806bee6f846d9ad4bccfc29afdca49a58)
    #1 0x7fd8fb3674c0 in inet_ntop (/lib/x86_64-linux-gnu/libc.so.6+0x13e4c0) (BuildId: 0323ab4806bee6f846d9ad4bccfc29afdca49a58)
    #2 0x557a244540f8 in inet_ntop /src/llvm-project/compiler-rt/lib/msan/../sanitizer_common/sanitizer_common_interceptors.inc:2650:15
    #3 0x557a244c8c24 in get_ndpi_flow_info /src/ndpi/fuzz/../example/reader_util.c:933:9
    #4 0x557a244c0050 in packet_processing /src/ndpi/fuzz/../example/reader_util.c:1815:12
    #5 0x557a244c0050 in ndpi_workflow_process_packet /src/ndpi/fuzz/../example/reader_util.c:2759:10
    #6 0x557a2449fdca in LLVMFuzzerTestOneInput /src/ndpi/fuzz/fuzz_readerutils_workflow.cpp:83:7
    #7 0x557a24395310 in fuzzer::Fuzzer::ExecuteCallback(unsigned char const*, unsigned long) /src/llvm-project/compiler-rt/lib/fuzzer/FuzzerLoop.cpp:614:13
    #8 0x557a24394b35 in fuzzer::Fuzzer::RunOne(unsigned char const*, unsigned long, bool, fuzzer::InputInfo*, bool, bool*) /src/llvm-project/compiler-rt/lib/fuzzer/FuzzerLoop.cpp:516:7
    #9 0x557a24396ac2 in fuzzer::Fuzzer::ReadAndExecuteSeedCorpora(std::__Fuzzer::vector<fuzzer::SizedFile, std::__Fuzzer::allocator<fuzzer::SizedFile>>&) /src/llvm-project/compiler-rt/lib/fuzzer/FuzzerLoop.cpp:829:7
    #10 0x557a24396db2 in fuzzer::Fuzzer::Loop(std::__Fuzzer::vector<fuzzer::SizedFile, std::__Fuzzer::allocator<fuzzer::SizedFile>>&) /src/llvm-project/compiler-rt/lib/fuzzer/FuzzerLoop.cpp:867:3
    #11 0x557a24385eeb in fuzzer::FuzzerDriver(int*, char***, int (*)(unsigned char const*, unsigned long)) /src/llvm-project/compiler-rt/lib/fuzzer/FuzzerDriver.cpp:914:6
    #12 0x557a243b12c2 in main /src/llvm-project/compiler-rt/lib/fuzzer/FuzzerMain.cpp:20:10
2025-06-29 03:26:53,518 - root - INFO - Fuzzer: fuzz_readerutils_workflow. Detected bug.
2025-06-29 03:26:53,518 - root - INFO - Trying to reproduce crash using: /tmp/tmpa5aa3d3t/crash-ad3ee257193550b65d98465331c0d8ef0fb84715.
    #13 0x7fd8fb24d082 in __libc_start_main (/lib/x86_64-linux-gnu/libc.so.6+0x24082) (BuildId: 0323ab4806bee6f846d9ad4bccfc29afdca49a58)
    #14 0x557a2437876d in _start (build-out/fuzz_readerutils_workflow+0x1f576d)

DEDUP_TOKEN: inet_ntop--__interceptor_inet_ntop
MemorySanitizer can not provide additional info.
SUMMARY: MemorySanitizer: SEGV (/lib/x86_64-linux-gnu/libc.so.6+0x18a178) (BuildId: 0323ab4806bee6f846d9ad4bccfc29afdca49a58) 
==47==ABORTING
```

IvanNardi: Sorry, I was not clear... My point was about :
```
if (!newflow->src_name || newflow->src_name_len != needed_len) {
          if (newflow->src_name) ndpi_free(newflow->src_name);
```
I do expect that `newflow->src_name` is always NULL here....

src/lib/ndpi_utils.c
fabiodepin: This change is not necessary, I will send the adjust it later tonight.
Only NULL validation would be appropriate here, but not necessary.

example/reader_util.c
fabiodepin: Yes, it is true.
It should always be NULL here. I will remove the additional validation and then test. Then I will submit again.

IvanNardi (approved):
LGTM, thank you!

PR #2913, opened 2025-07-09
Add the concept of protocols stack: more than 2 protocols per flow

IvanNardi:
The idea is to remove the limitation of only two protocols ("master" and "app") in the flow classifcation.
This is quite handy expecially for STUN flows and, in general, for any flows where there is some kind of transitionf from a cleartext protocol to TLS: HTTP_PROXY -> TLS/Youtube; SMTP -> SMTPS (via STARTTLS msg).

In the vast majority of the cases, the protocol stack is simply Master/Application.

Examples of real stacks (from the unit tests)  different from the standard "master/app":
* "STUN.WhatsAppCall.SRTP": a WA call
* "STUN.DTLS.GoogleCall": a Meet call
* "Telegram.STUN.DTLS.TelegramVoip": a Telegram call
* "SMTP.SMTPS.Google": a SMTP connection to Google server started in cleartext and updated to TLS
* "HTTP.Google.ntop": a HTTP connection to a Google domain (match via "Host" header) and to a ntop server (match via "Server" header)

The logic to create the stack is still a bit coarse: we have a decade of code try to push everything in only two protocols... Therefore, the content of the stack is still **highly experimental** and might change in the next future; do you have any suggestions?

It is quite likely that the legacy fields "master_protocol" and "app_protocol" will be there for a long time.

Add some helpers to use the stack:
```
ndpi_stack_get_upper_proto();
ndpi_stack_get_lower_proto();
bool ndpi_stack_contains(struct ndpi_proto_stack *s, u_int16_t proto_id);
bool ndpi_stack_is_tls_like(struct ndpi_proto_stack *s);
bool ndpi_stack_is_http_like(struct ndpi_proto_stack *s);
```

Be sure new stack logic is compatible with legacy code:
```
assert(ndpi_stack_get_upper_proto(&flow->detected_protocol.protocol_stack) ==
       ndpi_get_upper_proto(flow->detected_protocol));
assert(ndpi_stack_get_lower_proto(&flow->detected_protocol.protocol_stack) ==
       ndpi_get_lower_proto(flow->detected_protocol));
```


lucaderi:
The idea makes sense. However I would like to see some function/macro for reading the bottom (master_protocol) and top (app_protocol) to ease migration of existing applications.

IvanNardi:
> The idea makes sense. However I would like to see some function/macro for reading the bottom (master_protocol) and top (app_protocol) to ease migration of existing applications.

Done. Also add the list of new helpers to the commit message

Issue #2914, opened 2025-07-09
JA4 ALPN fingerprint uses incorrect characters


drnpkr:
## Bug Description

The JA4 fingerprint implementation in nDPI has multiple issues with ALPN parsing that cause incorrect fingerprints incompatible with other JA4 implementations.

## Expected Behavior

According to the JA4 specification and the comment in src/lib/protocols/tls.c at line 2140: "first and last characters of first ALPN extension value"

For ALPN values:
- "http/1.1" → should use h1 (first='h', last='1')
- "h2;http/1.1" → should use h2 (first='h', last='2' from first protocol)

## Actual Behavior

nDPI has multiple parsing issues:
1. **Correct handling** of multi-protocol ALPNs: "h2;http/1.1" produces h2
2. Uses second character instead of last: "http/1.1" **produces ht instead of h1**
3. String modification interference: nDPI's internal null terminator insertion corrupts the last character

## Impact

This bug causes nDPI JA4 fingerprints to be incompatible with other JA4 implementations:

Multi-protocol ALPN (h2;http/1.1):
  - ✅ nDPI: t13d1516h2_8daaf6152771_d8a2da3f94cd
  - ✅ Wireshark: t13d1516h2_8daaf6152771_d8a2da3f94cd

Single protocol ALPN (http/1.1):
  - ❌ nDPI: t13d3113**ht**_e8f1e7e78f70_1b3407e2c936
  - ✅ Wireshark: t13d3113**h1**_e8f1e7e78f70_1b3407e2c936

Example PCAP to reproduce the issue: http://pekar.s.cnl.sk/ndpi/ja4_sample_flows.pcap

PR #2915, opened 2025-07-09
Fix JA4 fingerprinting

drnpkr:
According to the JA4 specification/description (line 2140), the ALPN field should contain the first and last characters of the first ALPN extension value.

Currently, nDPI uses the first and second characters (alpn[0] and alpn[1]), which produces incorrect fingerprints that do not match other JA4 implementations like Wireshark.

For example, with ALPN 'http/1.1':
- Current (incorrect): 'ht' (first + second char)
- Fixed (correct):     'h1' (first + last char)

This change ensures nDPI's JA4 implementation conforms to the official specification and maintains interoperability with other JA4 tools.

Fixes: Incorrect JA4 ALPN fingerprint generation

- [ X] I have signed the ntop Contributor License Agreement at https://github.com/ntop/legal/blob/main/individual-contributor-licence-agreement.md
- [ X] I have read the contributing guidelines at https://github.com/ntop/nDPI/blob/dev/CONTRIBUTING.md

Link to the related [issue](https://github.com/ntop/nDPI/issues): https://github.com/ntop/nDPI/issues/2914

Describe changes:
  - **File**: `src/lib/protocols/tls.c`
  - **Line**: 2213
  - **Change**: `ja->client.alpn[1]` → `ja->client.alpn[strlen(ja->client.alpn)-1]`


drnpkr:
The previous fix attempted to use strlen(ja->client.alpn)-1 but this was
insufficient because nDPI modifies the ALPN string by:
-  Adding null terminators that truncate the last character
-  Converting semicolons to dashes, affecting multi-protocol ALPNs

The latest fix:
- Adds alpn_original_last field to store the true last character
- Captures the last character of the FIRST ALPN protocol only (before ;/,)
- Preserves the original character before nDPI's string modifications

Now correctly implements JA4 spec: first + last characters of first ALPN protocol
Examples:
- ALPN 'h2;http/1.1' -> 'h2' (not 'h.' or 'h1')
- ALPN 'http/1.1' -> 'h1' (not 'ht' or 'h.')

IvanNardi:
@drnpkr, could you update the unit tests results, please?
Something like `NDPI_FORCE_UPDATING_UTESTS_RESULTS=1 ./tests/do.sh` and push the changes...

drnpkr:
@IvanNardi sure thing! I already executed and pushed.

drnpkr:
@IvanNardi I also noticed one additional potential bug: I noticed that nDPI currently marks flows as having SNI (d) even when there is no actual TLS extension type 0x0000 (SNI) in the packet.

The technical details here https://github.com/FoxIO-LLC/ja4/blob/main/technical_details/JA4.md say: If the ClientHello contains an SNI extension (type `0x0000`), the SNI flag is `d`. Otherwise (no SNI), it is `i`. For some flows, currently nDPI incorrectly claims SNI present.

The bug is in line 2208 of src/lib/protocols/tls.c: https://github.com/drnpkr/nDPI/blob/6646cf113a43af6308dd7ce0289877d9cdc13523/src/lib/protocols/tls.c#L2208

`ja_str[3] = ndpi_isset_risk(flow, NDPI_NUMERIC_IP_HOST) ? 'i' : 'd', ja_str_len = 4;`

In my understanding:
  - nDPI uses NDPI_NUMERIC_IP_HOST risk to determine if SNI contains an IP address
  - If NDPI_NUMERIC_IP_HOST is set → `i` (IP address in SNI)
  - If NDPI_NUMERIC_IP_HOST is NOT set → `d` (assumes domain SNI present)

However, when there is NO SNI extension at all, NDPI_NUMERIC_IP_HOST is never set (it is only set when SNI exists AND contains an IP). Therefore, flows without any SNI extension incorrectly get `d` instead of `i`.

This is a relatively easy fix. This PR now fixes also https://github.com/ntop/nDPI/issues/2916

IvanNardi (approved):
LGTM, thanks for these fixes!

PR #2918, opened 2025-07-23
build tests in oss-fuzz

AdamKorcz:
I changed the OSS-Fuzz build script in https://github.com/google/oss-fuzz/pull/13597 to add support for Chronos. Propagating the changes here to remove the .diff in OSS-Fuzz.

OSS-Fuzz's `run_tests.sh` is used as part of Chronos with cached builds: https://github.com/google/oss-fuzz/tree/master/infra/experimental/chronos#check-tests

Describe changes:

### DIFF
--- a/tests/ossfuzz.sh
+++ b/tests/ossfuzz.sh
@@ -42,9 +42,7 @@ fi
 
 # build project
 cd ndpi
-# Set LDFLAGS variable as workaround for the "missing
-# dependencies errors" in the introspector build. See #8939
-RANLIB=llvm-ranlib LDFLAGS="-L/usr/local/lib -lpcap" ./autogen.sh --enable-fuzztargets --enable-tls-sigs
+RANLIB=llvm-ranlib ./autogen.sh --enable-fuzztargets --enable-tls-sigs
 make -j$(nproc)
 # Copy fuzzers
 ls fuzz/fuzz* | grep -v "\." | while read -r i; do cp "$i" "$OUT"/; done
--- a/tests/ossfuzz.sh
+++ b/tests/ossfuzz.sh
@@ -42,9 +42,9 @@ fi
 
 # build project
 cd ndpi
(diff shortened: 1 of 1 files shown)


tests/ossfuzz.sh
```diff
@@ -42,9 +42,9 @@ fi
 
 # build project
 cd ndpi
-# Set LDFLAGS variable and `--with-only-libndpi` option as workaround for the
-# "missing dependencies errors" in the introspector build. See #8939
-RANLIB=llvm-ranlib LDFLAGS="-L/usr/local/lib -lpcap" ./autogen.sh --enable-fuzztargets --with-only-libndpi --enable-tls-sigs
+# Set LDFLAGS variable as workaround for the "missing
+# dependencies errors" in the introspector build. See #8939
+RANLIB=llvm-ranlib LDFLAGS="-L/usr/local/lib -lpcap" ./autogen.sh --enable-fuzztargets --enable-tls-sigs
```
IvanNardi: I think we can completely remove  the LDFLAGS value and th entire comment.... (we needed to explicitly link to libpcap because of `--with-only-libndpi`, but you just rmeove that...)

IvanNardi: Pushed this change in a separate commit

AdamKorcz:
Thanks Ivan. What do you prefer to do about CIFuzz? As I see it we can:

1. Force the PR through.
2. Revert the changes in OSS-Fuzz and retrigger the CI here so it passes.

IvanNardi:
> Thanks Ivan. What do you prefer to do about CIFuzz? As I see it we can:
> 
>     1. Force the PR through.
> 
>     2. Revert the changes in OSS-Fuzz and retrigger the CI here so it passes.

I'll merge this PR now and I am going to prepare a patch for oss-fuzz tomorrow. Not a big deal if the fuzzing fails for a few days, IMHO

IvanNardi:
@AdamKorcz, thanks for your contribution and for all your work on oss-fuzz!

Issue #2906, opened 2025-06-25
Extract MDNS Metadata


lucaderi:
MDNS is often used to advertise host services as in the enclosed packets.

This ticket is to request to extend struct ndpi_flow_struct { . } adding a new entry for MDNS (similar to SSDP an other protocols) that in case of advertised services (as in the enclosed pcap) they are reported in the flow metadata.

[mdns.pcapng.zip](https://github.com/user-attachments/files/20907809/mdns.pcapng.zip)

Gianla:
Should the MDNS entry be structured similarly to the DNS one, or should it only contain advertised host service names?

Regarding metadata parsing, would it be better to extend /src/lib/protocols/dns.c, or should I implement a new protocol dissector for MDNS?

lucaderi:
Use the dns dissector, no need to create fresh code as it is already doing what's necessary.

Gianla:
I've tried to add the MDNS entry outside the union because i don't want it to interfere with the dns entry, but at this point at compile time i get "Size of the flow struct increased to more than 1232 bytes, please check if this change is necessary.".
Before changing the hardcoded limit of 1232 bytes, i want to ask if that is the correct way or if it should be done in another way.

IvanNardi:
Is it possible to add a new specific  mdns entry in the union, at the same
level of dns or tls_quic?

Gianla:
For the mdns.pcap.ng file linked here, is it intended behaviour that process_aswers() is never called? You can check it from this simple fork which is only debugging process_aswers and process_queries: https://github.com/Gianla/nDPI-fork
I can't copy answers and save them in a metadata field anyways if this code is never reached. Changing dns.c too much (for example search_dns() function) doesn't seem like the best option to me.

IvanNardi:
@Gianla 
Two points.
First. Right now `process_answers()` is called only for reply messages and you don't have replay messages in that pcap... Let's try something trivial like:
```
--- a/src/lib/protocols/dns.c
+++ b/src/lib/protocols/dns.c
@@ -821,6 +821,8 @@ static void search_dns(struct ndpi_detection_module_struct *ndpi_struct, struct
     if(rc == -1)
       printf("[DNS] Error queries (query msg)\n");
 #endif
+    off = rc;
+    rc = process_answers(ndpi_struct, flow, &dns_header, off, &proto);
   } else {
```
Just a POC; no proper error checking!
[...]

PR #2929, opened 2025-08-03
Implemented new MDNS feature

Gianla:
Patch to implement this [requested feature](https://github.com/ntop/nDPI/issues/2906).


example/ndpiReader.c
```diff
@@ -1985,6 +1985,18 @@ static void printFlow(u_int32_t id, struct ndpi_flow_info *flow, u_int16_t threa
 	}
       }
 
+    if((flow->detected_protocol.proto.master_protocol == NDPI_PROTOCOL_MDNS) || (flow->detected_protocol.proto.app_protocol == NDPI_PROTOCOL_MDNS)) {
```
IvanNardi: && flow->ndpi_flow->mdns_metadata.num_services > 0

Remove the "none" stuff below

IvanNardi (changes requested):
Please, update the unit test results and push the changes

example/ndpiReader.c
```diff
@@ -1985,6 +1985,16 @@ static void printFlow(u_int32_t id, struct ndpi_flow_info *flow, u_int16_t threa
 	}
       }
 
+    if(((flow->detected_protocol.proto.master_protocol == NDPI_PROTOCOL_MDNS) ||
+       (flow->detected_protocol.proto.app_protocol == NDPI_PROTOCOL_MDNS)) &&
+       ((flow->ndpi_flow) && (flow->ndpi_flow->mdns_metadata.num_services > 0))) {
```
IvanNardi: This logic is wrong. Here, you need to print data from `flow`, not from `flow->ndpi_flow` (thus, the check is useless). You need to save all information in `flow`; take a look at `process_ndpi_collected_info()`.
In other words: you can't access `flow->ndpi_flow` here

src/lib/ndpi_main.c
```diff
@@ -7596,6 +7596,13 @@ void ndpi_free_flow_data(struct ndpi_flow_struct* flow) {
     if(flow->monit)
       ndpi_free(flow->monit);
 
+  	if(flow_is_proto(flow, NDPI_PROTOCOL_MDNS)) {
+  	  for(int i = 0; i < flow->mdns_metadata.num_services; ++i) {
+  		ndpi_free(flow->mdns_metadata.services[i]);
+      }
+    }
```
IvanNardi: Fix the indentation. You don't need to check for `NDPI_PROTOCOL_MDNS`

src/lib/protocols/dns.c
```diff
@@ -415,6 +440,25 @@ static int process_answers(struct ndpi_detection_module_struct *ndpi_struct,
                  We unconditionally update it at the end of the for loop */
               x = orig_x;
               found = 1;
+
+              if (ignore_checks) {
+                u_int8_t *num_services = &flow->mdns_metadata.num_services;
```
IvanNardi: If you really want to use a local variable here, why a pointer? Use a plain integer type...

src/lib/protocols/dns.c
```diff
@@ -817,10 +865,19 @@ static void search_dns(struct ndpi_detection_module_struct *ndpi_struct, struct
     flow->protos.dns.transaction_id = dns_header.tr_id;
 
     rc = process_queries(ndpi_struct, flow, &dns_header, off);
+    if(rc == -1) {
 #ifdef DNS_DEBUG
-    if(rc == -1)
       printf("[DNS] Error queries (query msg)\n");
 #endif
+    } else if (proto.master_protocol == NDPI_PROTOCOL_MDNS) {
```
IvanNardi: I think you can remove that check; the change should be fine also for DNS and LLMNR

IvanNardi (changes requested):
Some changes required; please remember to update the unit tests results before pushing.
Thanks!

example/ndpiReader.c
Gianla: Wasn't the feature supposed to extend `ndpi_flow_struct` (not `ndpi_flow_info`)? In this case, the check isn't useless because some tests trying to access that variable gave `segmentation fault`.

src/lib/protocols/dns.c
Gianla: I will use the original reference. It was just to make it shorter.

example/ndpiReader.c
IvanNardi: Yes it was and you correctly update the library.
But if you want to print this info from ndpiReader (i.e. the application) you need to access only `ndpi_flow_info`, not `ndpi_flow_struct`: as you noticed, the latter is not always available at that time.
That is the reason why `process_ndpi_collected_info()` copies ALL metadata from `ndpi_flow_struct` to `ndpi_flow_info`:  the application needs a local copy of that data

Gianla: Thank you for your reply. So should i add `mdns_metadata` also to `ndpi_flow_info` is that right?

IvanNardi: yes, like all the other metadata

src/lib/protocols/dns.c
IvanNardi: Ok! Thank you [removed a wrong previous message; sorry for the confusion]

example/ndpiReader.c
Gianla: Thanks, that is clear now. Should the metadata's copy happen under
`else if(ndpi_stack_contains(&flow->detected_protocol.protocol_stack, NDPI_PROTOCOL_MDNS)) {`
or should I create a new branch?

IvanNardi: It should be fine

Gianla: I'm duping the strings in `mdns_metadata.services` since the freeing appears to be happening twice, but sometimes the strings do not get freed even if the freeing function is the same as `ndpi_flow_struct` (as far as I'm understanding). At the same time, if I just copy the addresses, I get a `segmentation fault` because the freeing is happening twice sometimes. Am I missing something?

Gianla: Sorry for the previous comment. Now that I changed [this](https://github.com/ntop/nDPI/pull/2929#discussion_r2251278929) the duping was also happening for non-mdns protocols. Moving the freeing for cycle out of the check solved the issue.

Gianla: Deleted previous replies, sorry for the confusion. I'm still having memory problems. If I dup a string inside `process_ndpi_collected_info()`, where should I free it?

src/lib/protocols/dns.c
Gianla: Because of this change now the output states `MDNS Advertised services: ...` even for more general protocols. Should we keep the condition in `ndpiReader` or it's okay this way?

IvanNardi: How is that possible? 
You populate `flow->mdns_metadata` only if `ignore_checks` is true, and than happens only with MDNS traffic...
What am I missing?

Gianla: Sorry, my bad.

koltiradw:
@Gianla, hi!
Can you also add the serialization of mdns services like [here](https://github.com/ntop/nDPI/blob/8e1b17215dc7a0ff0541503385664298dfacf3b9/src/lib/ndpi_utils.c#L1351)?

IvanNardi:
@Gianla , I pushed the correct unit tests results (hopefully!) into your branch

IvanNardi:
> @Gianla, hi! Can you also add the serialization of mdns services like [here](https://github.com/ntop/nDPI/blob/8e1b17215dc7a0ff0541503385664298dfacf3b9/src/lib/ndpi_utils.c#L1351)?

Yes, please. There is already a `case NDPI_PROTOCOL_MDNS:` branch

Gianla:
> @Gianla , I pushed the correct unit tests results (hopefully!) into your branch

Thank you so much!

Gianla:
@koltiradw @IvanNardi like [this](https://github.com/ntop/nDPI/commit/0f3bc5d9bdb5870515d94c2368be17a3a2b4301e)?

IvanNardi:
> @koltiradw @IvanNardi like [this](https://github.com/ntop/nDPI/commit/0f3bc5d9bdb5870515d94c2368be17a3a2b4301e)?

Fine for me. @koltiradw ?

koltiradw:
> > @koltiradw @IvanNardi like [this](https://github.com/ntop/nDPI/commit/0f3bc5d9bdb5870515d94c2368be17a3a2b4301e)?
> 
> Fine for me. @koltiradw ?

Looks great! Thx!

Gianla:
> > > @koltiradw @IvanNardi like [this](https://github.com/ntop/nDPI/commit/0f3bc5d9bdb5870515d94c2368be17a3a2b4301e)?
> > 
> > 
> > Fine for me. @koltiradw ?
> 
> Looks great! Thx!

Amazing, thank you!

IvanNardi (approved):
LGTM, thaks for your contribution!
Let's wait for @lucaderi approval....

tests/cfgs/default/result/anyconnect-vpn.pcap.out
```diff
@@ -70,21 +70,21 @@ JA Host Stats:
...
+	13	UDP [fe80::408:3e45:3abc:1552]:5353 -> [ff02::fb]:5353 [proto: 8/MDNS][Stack: MDNS][IP: 0/Unknown][MDNS advertised services: lp-rkerur-osx (9)._companion-link._tcp.local, lp-rkerur-osx (9)._companion-link._tcp.local][ClearText][Confidence: DPI][FPC: 8/MDNS, Confidence: DPI][DPI packets: 6][cat: Network/14][Breed: Acceptable][9 pkts/1628 bytes -> 0 pkts/0 bytes][Goodput ratio: 66/0][25.40 sec][Hostname/SNI: _raop._tcp.local][_raop._tcp.local][bytes ratio: 1.000 (Upload)][IAT c2s/s2c min/avg/max/stddev: 819/0 3174/0 11263/0 3646/0][Pkt Len c2s/s2c min/avg/max/stddev: 152/0 181/0 206/0 24/0][PLAIN TEXT (companion)][Plen Bins: 0,0,33,22,44,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0]
```
lucaderi: Duplicates

lucaderi:
IMHO this code is not useful in the current state. MSDN is used to advertise services and report information about them. The current implementation in essence maps the MDNS names into a table (with duplicates), it does not report any information about the record that is exactly the purpose of MDNS. Example see https://github.com/ntop/nDPI/pull/2929/files#r2253543262 it does not say anything about the service (e.e. Android.local), names are duplicates, and in essence this is like DNS.

So I consider this a draft PR, but it wouldn't merge the code until I see an improvement with the current DNS implementation. This is my advice at least.

Gianla:
What should be added next?

Gianla:
> IMHO this code is not useful in the current state. MSDN is used to advertise services and report information about them. The current implementation in essence maps the MDNS names into a table (with duplicates), it does not report any information about the record that is exactly the purpose of MDNS. Example see https://github.com/ntop/nDPI/pull/2929/files#r2253543262 it does not say anything about the service (e.e. Android.local), names are duplicates

Thank you for the feedback.
I can easily fix the duplicates — I had already considered it, but initially decided to stay as close as possible to the raw packet data, so I didn't include any filtering or deduplication logic.

Regarding the service information (e.g., Android.local), I’m not sure where exactly I should extract that from in the packet. Could you clarify which specific fields or record types I should include in the structure?

> and in essence this is like DNS

Note that, as far as I understood the code, for flows with `n` service(s), the name is overwritten by `n-1` times and the previous names (if any) aren't stored in the struct.

PR #2932, opened 2025-08-05
Add Mudfish protocol dissector

utoni:

Note: This dissector does not support the full protocol, as more (PCAP) data is needed.


utoni:
@IvanNardi 
Any ideas why the assertion in https://github.com/ntop/nDPI/actions/runs/16742464127/job/47393479034?pr=2932#logs is failing? It does not fail locally using the same configuration.

IvanNardi:
> @IvanNardi Any ideas why the assertion in https://github.com/ntop/nDPI/actions/runs/16742464127/job/47393479034?pr=2932#logs is failing? It does not fail locally using the same configuration.

Hi @utoni!
Let's take a step back. About the classification `IMO/Mudfish`: is that real? I mean, does Mudfish use IMO as "master"/"cover" protocol (to hide itself?) ? Or are there same issues about valid Mudfish flows wrongly classified as IMO too early by nDPI?

IvanNardi:
BTW, you need the `NDPI_EXTENDED_SANITY_CHECKS` defines, which is always defined in CI

utoni:
> > @IvanNardi Any ideas why the assertion in https://github.com/ntop/nDPI/actions/runs/16742464127/job/47393479034?pr=2932#logs is failing? It does not fail locally using the same configuration.
> 
> Hi @utoni!
> Let's take a step back. About the classification `IMO/Mudfish`: is that real? I mean, does Mudfish use IMO as "master"/"cover" protocol (to hide itself?) ? Or are there same issues about valid Mudfish flows wrongly classified as IMO too early by nDPI?

Yea, Mudfish uses "single byte" ping packets with 0x50 as payload. IMO uses a similar packet structure but not explicitly using 0x50. That's were the false positive was coming from.

IvanNardi:
> > > @IvanNardi Any ideas why the assertion in https://github.com/ntop/nDPI/actions/runs/16742464127/job/47393479034?pr=2932#logs is failing? It does not fail locally using the same configuration.
> > 
> > 
> > Hi @utoni!
> > Let's take a step back. About the classification `IMO/Mudfish`: is that real? I mean, does Mudfish use IMO as "master"/"cover" protocol (to hide itself?) ? Or are there same issues about valid Mudfish flows wrongly classified as IMO too early by nDPI?
> 
> Yea, Mudfish uses "single byte" ping packets with 0x50 as payload. IMO uses a similar packet structure but not explicitly using 0x50. That's were the false positive was coming from.

So I'd try to avoid IMO stuff at all costs, even if that means to update/change IMO dissector, adding some exceptions

utoni:
I do not like this "new" protocol stack. It's not transparent why it's there and what is the difference to the "old" one.

Also the obscure `-DNDPI_EXTENDED_SANITY_CHECKS` confuses me even more.
The assertion is triggered while running:
`NDPI_FORCE_PARALLEL_UTESTS=1 NDPI_FORCE_UPDATING_UTESTS_RESULTS=1 NDPI_DISABLE_FUZZY=1 ./tests/do.sh`, but **not** if running: `./example/ndpiReader -i tests/pcap/mudfish.pcap`

IvanNardi:
I'll try to answer to the stack part later. Sorry...

In the meantime: 

> Also the obscure `-DNDPI_EXTENDED_SANITY_CHECKS` confuses me even more.
> The assertion is triggered while running:
> `NDPI_FORCE_PARALLEL_UTESTS=1 NDPI_FORCE_UPDATING_UTESTS_RESULTS=1 NDPI_DISABLE_FUZZY=1 ./tests/do.sh`, but **not** if running: `./example/ndpiReader -i tests/pcap/mudfish.pcap`

The assertion is triggered only with `-v2` option, because the stack is retrieved by `ndpiReader` (and so it is checked...) only if/when  printing the flow information... (test.sh implicitly uses `-v2`)
Do you think it would be best if we check that unconditionally (i.e. with or without `-v2`)?

IvanNardi:
@utoni, about the stack...
The basic idea is to allow more than 2 protocols in flow classification: it is useless in 95% of the cases (because as stack you simple have master.app), but quite meaningful in some complex cases as STUN and general proxy/tunnel.
I tried to explain more in the original commit: https://github.com/ntop/nDPI/commit/8dd2220116c96c00066dfd88344effe432c8f5a8
It is quite an experimental feature, so any suggestions are welcomed!

In the vast majority of the case, it is completely transparent. 
Here you get an error, because this patch is trying to use the subclassification feature in some "non-standard" way: you explicitly tell that Mudfish is an IMO subprotocol (via `ndpi_set_proto_subprotocols()`) but when that data path is used, you forced `Mudfish` as classification, instead of the expected `IMO.Mudfish` (expected because you are in the "subprotocol" data path).
If you are going to ignore the IMO stuff altogether, better to avoid IMO match/false-positive in the first place and completly remove the sub-protocol code.
I am even fine with a specific case in IMO code: something like if the unique byte is 0x50, skip IMO and let Mudfish dissector to kick-in....
[it seems IMO code needs some upgrade anyway...]

utoni:
> I am even fine with a specific case in IMO code: something like if the unique byte is 0x50, skip IMO and let Mudfish dissector to kick-in....

That's something I always try to avoid. Some dissectors i.e. TLS, QUIC or STUN contain dissection logic of other protocols. While it's fine for those complex ones, IMO and Mudfish are more of the simple kind and there should be no need that IMO contains Mudfish related code or detection patterns.

Maybe another API function similar to `ndpi_set_proto_subprotocols()` might be better, i.e. `ndpi_set_similar_protocols()`, which does basically the same, but instead of setting a subprotocol, it replaces the "main" protocol.

IvanNardi:
> > I am even fine with a specific case in IMO code: something like if the unique byte is 0x50, skip IMO and let Mudfish dissector to kick-in....
> 
> That's something I always try to avoid. Some dissectors i.e. TLS, QUIC or STUN contain dissection logic of other protocols. While it's fine for those complex ones, IMO and Mudfish are more of the simple kind and there should be no need that IMO contains Mudfish related code or detection patterns.
> 
> Maybe another API function similar to `ndpi_set_proto_subprotocols()` might be better, i.e. `ndpi_set_similar_protocols()`, which does basically the same, but instead of setting a subprotocol, it replaces the "main" protocol.

Let's see if I got that right:
* the issue is with flow with pkt 1, len 1, 0x15 and pkt 2 as pkt 1
* this flow matches mudfish and IMO dissectors both
* we want this flow to be classified as mudfish
* we don't want to /can't improve the logic on the dissectors
* we don't want to add a special case in the IMO dissector

In other words, we want Mudfish to win againt IMO
If that is correct, the patch (on top of your code) is trivial:
```
--- a/src/lib/ndpi_main.c
+++ b/src/lib/ndpi_main.c
@@ -6490,9 +6490,6 @@ static int dissectors_init(struct ndpi_detection_module_struct *ndpi_str) {
   /* SIP */
   init_sip_dissector(ndpi_str);
 
-  /* IMO */
-  init_imo_dissector(ndpi_str);
-
   /* Teredo */
   init_teredo_dissector(ndpi_str);
 
@@ -7227,6 +7224,9 @@ static int dissectors_init(struct ndpi_detection_module_struct *ndpi_str) {
   /* Mudfish */
   init_mudfish_dissector(ndpi_str);
 
+  /* IMO */
+  init_imo_dissector(ndpi_str);

```
This way Mudfish logic is checked before IMO (instead of the other way around)

utoni:
That solves the issue, but it is not transparent to a developer that the order of the dissectors init functions decides when they get called :confused:

IvanNardi:
@utoni, I think you add some spurious fuzzing pcap by error...

IvanNardi:
> That solves the issue, but it is not transparent to a developer that the order of the dissectors init functions decides when they get called 😕

I honestly don't know how to better handle that list... BTW, it also has a huge impact on performance...

utoni:
Whoops 😬

IvanNardi:
@utoni: merged dev and sync unit tests results :crossed_fingers:

IvanNardi:
The fuzzing errors are unrelated

IvanNardi (approved):
Thank you (for you patience...)

PR #2937, opened 2025-08-07
fuzz: add new fuzzers for bitmask and filter data structures

IvanNardi:


IvanNardi:
Draft only because it depends on #2934

PR #2940, opened 2025-08-10
Implemented new mDNS feature (POC)

Gianla:
Proof of concept for a patch implementing [this feature](https://github.com/ntop/nDPI/issues/2906).

This patch extended `ndpi_mdns_rsp_entry` with additional useful data and uses it to enhance `ndpi_flow_struct`.

Some points:
1. I updated `ndpiReader.c` as well, though the formatting may not be ideal; I’m open to suggestions for improvements.
2. Changes in `dns.c` could potentially apply to standard DNS, but for now all modifications are guarded by the `proto->master_protocol == NDPI_PROTOCOL_MDNS` condition;
3. In particular, I added support for compressed names, as well as SRV and TXT records, which were not handled previously
4. ~~Right now I updated as many tests as I could, except for `/tests/pcap/fuzz-2006-06-26-2594.pcap`. Probably due to the `ndpi_grab_dns_name()` update, it still gives some problems with text format. After a lot of time debugging it, I can't understand what's wrong with it. Any suggestion about it is appreciated.~~ DONE. Looks like opening it to update a test (n. 196) corrupted the entire file, so the `diff` never matched.
5. ~~Duplicate checks are currently not implemented, but they can be easily added above the new `add_to_mdns_metadata()` function if needed.~~ DONE
6. Some new checks can detect malformed packets. From my side, I did not feel fully confident to decide if and which `set_risk` should be used, so I have currently marked them as "TODO". I believe this is an important point that would benefit from further discussion with the core developers to determine the best approach.

If and when all changes for answer records will be approved, I'll be ready to extend it also for additional records since as the approach is basically the same.
If the format is okay, I will also update the serializer.


tests/cfgs/default/result/dns_invert_query.pcapng.out
Gianla, line 1: This 
`UDP 173.147.108.174:18427 <-> 244.187.95.1:53 [proto: 5/DNS]`
is not a mDNS flow. However, the new changes in `search_dns() `correctly detected a DNS record with zero TTL, which is confirmed by inspecting the packet in Wireshark. For this reason, I thought it was appropriate to update this test case accordingly.

src/include/ndpi_typedefs.h
```diff
@@ -1692,6 +1696,11 @@ struct ndpi_flow_struct {
 
   } protos;
 
+  struct {
+    uint8_t num_services;
+    struct ndpi_mdns_rsp_entry services[MAX_NUM_MDNS_ADVERTISED_SERVICES];
+  } mdns_metadata;
+
```
IvanNardi, line 1711: Here....

src/include/ndpi_typedefs.h
```diff
@@ -1749,8 +1758,8 @@ struct ndpi_flow_struct {
 _Static_assert(sizeof(((struct ndpi_flow_struct *)0)->protos) <= 264,
                "Size of the struct member protocols increased to more than 264 bytes, "
                "please check if this change is necessary.");
-_Static_assert(sizeof(struct ndpi_flow_struct) <= 1232,
-               "Size of the flow struct increased to more than 1232 bytes, "
+_Static_assert(sizeof(struct ndpi_flow_struct) <= 1408,
+               "Size of the flow struct increased to more than 1408 bytes, "
```
IvanNardi: ... I let @lucaderi decide (and comment on the entire PR), but I would strongly prefer we try *very* hard to avoid adding ~180 bytes to `struct ndpi_flow_struct`.
Can't we allocate this structure only on demand, like in the previous PR?

Gianla: Yes, it's surely possible. I will edit it. Note that in the previous PR it was just smaller, not dinamically allocated.
May I ask where should I allocate it? I'm afraid (if not sure) that if it is done in `search_dns()`, it will end overwrting itself. Maybe `ndpi_search_dns()`?

IvanNardi: The idea is to have something like;
```
  struct {
    uint8_t num_services;
    struct ndpi_mdns_rsp_entry *services;  <---- pointer
  } mdns_metadata;
```
and ONLY WHEN you want add the FIRST service, you allocate `MAX_NUM_MDNS_ADVERTISED_SERVICES` entries....
Does it sound good?

Gianla: It does, thank you for your support.

src/include/ndpi_typedefs.h
Gianla, line 1: Some tests didn't like this struct being packed. Now that it is dinamically allocated, we can remove the packing without increasing `ndpi_flow_struct`'s size.

example/ndpiReader.c
```diff
@@ -32,6 +32,8 @@
 #include <stdlib.h>
 #include <getopt.h>
 #include <float.h> /* FLT_EPSILON */
+
+#include "../src/include/ndpi_typedefs.h"
```
IvanNardi: Try removing this include....

example/reader_util.c
```diff
@@ -28,6 +28,8 @@
 #include <math.h>
 #include <float.h>
 
+#include "../src/include/ndpi_typedefs.h"
+
```
IvanNardi: ... and this one

Gianla: Thank you, never thought it could be that.

IvanNardi:
@Gianla, I removed another include

Gianla:
@IvanNardi I can't reproduce these bugs, `ndpi_mdns_rsp_entry` is always defined for me even with the artifacts of the fuzzer...
Edit: I think the fuzzer made some allocations fail, in that case there was a memory leak. I hope it's fixed now.

lucaderi:
This is work in progress that needs to be reworked before merging it into the codebase.

PR #2942, opened 2025-08-13
Provide an explicit state for the flow classification process

IvanNardi:
Application should keep calling nDPI until flow state became
`NDPI_STATE_CLASSIFIED`.

The main loop in the application is simplified to something like:
```
res = ndpi_detection_process_packet(...);
if(res->state == NDPI_STATE_CLASSIFIED) {
  /* Done: you can get finale classification and all metadata.
     nDPI doesn't need more packets for this flow */
} else {
  /* nDPI needs more packets for this flow. The provided
     classification is not final and more metadata might be
     extracted.
     If `res->state` is `NDPI_STATE_PARTIAL`, partial/initial
     classification is available in `res->proto`
     as usual but it can be updated later.
  */
}

/*
    Example A (QUIC flow):
     pkt 1: proto QUIC state NDPI_STATE_PARTIAL
     pkt 2: proto QUIC/Youtube  state NDPI_STATE_CLASSIFIED
    Example B (GoogleMeet call):
     pkt 1:   proto STUN state NDPI_STATE_PARTIAL
     pkt N:   proto DTLS state NDPI_STATE_PARTIAL
     pkt N+M: proto DTLS/GoogleCall state NDPI_STATE_CLASSIFIED
    Example C (standard TLS flow):
     pkt 1:   proto Unknown state NDPI_STATE_INSPECTING
     pkt 2:   proto Unknown state NDPI_STATE_INSPECTING
     pkt 3:   proto Unknown state NDPI_STATE_INSPECTING
     pkt 4:   proto TLS/Facebook state NDPI_STATE_PARTIAL
     pkt N:   proto TLS/Facebook state NDPI_STATE_CLASSIFIED
 */
}
```
You can take a look at `ndpiReader` for a slightly more complex example.

API changes:
* remove the third parameter from `ndpi_detection_giveup()`. If you need
to know if the classification flow has been guessed, you can access
`flow->protocol_was_guessed`
* remove `ndpi_extra_dissection_possible()`
* change some prototypes from accepting `ndpi_protocol foo` to
`ndpi_master_app_protocol bar`. The update is trivial: from `foo` to
`foo.proto`

### DIFF
--- a/example/ndpiReader.c
+++ b/example/ndpiReader.c
@@ -563,7 +563,7 @@ void ndpiCheckHostStringMatch(char *testChar) {
            detected_protocol.proto.app_protocol,
            detected_protocol.breed,
            detected_protocol.category,
-           ndpi_protocol2name(ndpi_str, detected_protocol, appBufStr,
+           ndpi_protocol2name(ndpi_str, detected_protocol.proto, appBufStr,
                               sizeof(appBufStr)),
            ndpi_get_proto_breed_name(detected_protocol.breed),
            ndpi_category_get_name(ndpi_str, detected_protocol.category));
@@ -640,7 +640,7 @@ void ndpiCheckHostsFileStringMatch(const char *domains_file) {
 
       printf("Domain [%s] -> %s %s %s\n",
              line,
-             ndpi_protocol2name(ndpi_str, detected_protocol, appBufStr,
+             ndpi_protocol2name(ndpi_str, detected_protocol.proto, appBufStr,
                                 sizeof(appBufStr)),
              ndpi_get_proto_breed_name(detected_protocol.breed),
              ndpi_category_get_name(ndpi_str, detected_protocol.category));
@@ -725,7 +725,7 @@ static void ndpiCheckIPMatch(char *testChar) {
     memset(&detected_protocol, 0, sizeof(ndpi_protocol));
     detected_protocol.proto.app_protocol = ndpi_map_ndpi_id_to_user_proto_id(ndpi_str, ret);
 
-    ndpi_protocol2name(ndpi_str, detected_protocol, appBufStr,
+    ndpi_protocol2name(ndpi_str, detected_protocol.proto, appBufStr,
                        sizeof(appBufStr));
 
     printf("Match Found for IP %s, port %d -> %s (%d)\n",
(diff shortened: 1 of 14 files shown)


mmanoj:
@IvanNardi useful feature, I'm currently developing a Flow State Machine (FSM) with an accumulator mechanism to optimize how packets are sent to nDPI based on protocol characteristics. For example, only a single packet is sent for DNS flows, while TLS flows require at least 6–8 packets before classification.

The goal is to prevent the application from continuously invoking nDPI until the flow state reaches NDPI_STATE_CLASSIFIED, which should significantly reduce CPU cycles and improve performance on hot paths.

If this approach sounds reasonable, I’ll be happy to share the implementation details and results once it’s completed. Please share your thoughts or suggestions.

IvanNardi:
> @IvanNardi useful feature, I'm currently developing a Flow State Machine (FSM) with an accumulator mechanism to optimize how packets are sent to nDPI based on protocol characteristics. For example, only a single packet is sent for DNS flows, while TLS flows require at least 6–8 packets before classification.
> 
> The goal is to prevent the application from continuously invoking nDPI until the flow state reaches NDPI_STATE_CLASSIFIED, which should significantly reduce CPU cycles and improve performance on hot paths.
> 
> If this approach sounds reasonable, I’ll be happy to share the implementation details and results once it’s completed. Please share your thoughts or suggestions.

Let me be clear about nDPI capabilities: libnDPI stops processing flow packets when it gets all the information (classification and metadata) *** required by the application***. That means:
* if you configure nDPI to avoid DNS metadata, processing stops after the first DNS request
* if you are interested only in (sub-)classification, TLS processing (usually) stops after the 4th packets (i.e. after the Client Hello)
* if you are not interested in metadata, STUN (i.e. Whatapp/Signal/Telegram/Messenger calls) stops after (usually) the first packet

In other words, nDPI already processes the minimum number of packets depending on the configuration; with the default configuration we usually want to get everything but you can tune that depending on your usercase.

It is a tradeoff between performance and information extracted.

All of that was true even before this PR has been merged.

Are you doing something different, @mmanoj ?

mmanoj:
@IvanNardi 

Sorry for the delay, I missed the message.We have implemented a lightweight Flow State Machine (FSM) on top of nDPI to manage multi-stage classification for encrypted transports like QUIC and TLS.
nDPI remains the authoritative classifier — we do not modify any internal code.
The FSM adds orchestration logic (buffering, tiered packet feeding, DNS correlation, tentative classification, and flow state transitions) to prevent early leakage before nDPI produces a final APP ID.
We believe this layer belongs in the client application (e.g., NFQUEUE handler or nDPI Reader) rather than nDPI core, since it handles control/decision timing rather than DPI heuristics.
We request your feedback on whether the staging thresholds and tier logic align with nDPI’s recommended practices for QUIC/TLS classification and whether any internal signals (e.g., early TLS hints) should be surfaced to the FSM.

<h1 data-start="628" data-end="667"><strong data-start="633" data-end="667">Why We Introduced an FSM Layer</strong></h1>
<p data-start="668" data-end="789">Modern encrypted transports (especially QUIC and TLS 1.3) require <strong data-start="734" data-end="754">multiple packets</strong> before nDPI can reliably classify:</p>
<ul data-start="791" data-end="906">
<li data-start="791" data-end="847">
<p data-start="793" data-end="847">QUIC: Client Initial → Handshake → encrypted packets</p>
</li>
<li data-start="848" data-end="906">
<p data-start="850" data-end="906">TLS: ClientHello → ServerHello → Cert → encrypted data</p>
</li>
</ul>
<p data-start="908" data-end="962">This leads to the following issues seen in production:</p>
<p>Slow classification	due to App ID unknown for first 4–12 packets</p>
<p>Service leakage	YouTube plays-> for 2–3 seconds before DPI detects it</p>
<p>nDPI must be called repeatedly ->	Client application lacks a state machine</p>
<p>Early packets bypass analysis ->	Incorrect NFQUEUE verdicts for first packets</p>
<p>No pending/hold state ->	Application forces immediate decisions</p>

</body></html>

</div></div>

PR #2960, opened 2025-09-08
Fix JSON value "amount of TLS blocks" in tls2json if dissected protocol is QUIC

utoni:

The value of `flow.l4.tcp.tls.num_tls_blocks` is also serialized if the detected protocol is QUIC. That'll lead to QUIC flow memory interpreted if it was a TLS/DTLS protocol.

FYI: https://github.com/utoni/nDPId/actions/runs/17551820139/job/49845912133#step:18:540

### DIFF
--- a/src/lib/ndpi_utils.c
+++ b/src/lib/ndpi_utils.c
@@ -1206,7 +1206,7 @@ void ndpi_serialize_proto(struct ndpi_detection_module_struct *ndpi_struct,
 
 /* ********************************** */
 
-static void ndpi_tls2json(ndpi_serializer *serializer, struct ndpi_flow_struct *flow)
+static void ndpi_tls2json(ndpi_serializer *serializer, struct ndpi_flow_struct *flow, bool is_tls_proto)
 {
   if(flow->protos.tls_quic.ssl_version)
   {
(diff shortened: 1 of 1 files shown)


src/lib/ndpi_utils.c
```diff
@@ -1766,11 +1769,11 @@ int ndpi_dpi2json(struct ndpi_detection_module_struct *ndpi_struct,
     break;
 
   case NDPI_PROTOCOL_TLS:
-    ndpi_tls2json(serializer, flow);
+    ndpi_tls2json(serializer, flow, true);
     break;
 
   case NDPI_PROTOCOL_DTLS:
-    ndpi_tls2json(serializer, flow);
+    ndpi_tls2json(serializer, flow, true);
```
IvanNardi: false? DTLS is, usually, over UDP...

utoni: But doesn't it use the same flow memory region as TLS?

IvanNardi: The  rule is that we can read/write to `flow.l4.tcp` ONLY if `flow->l4_proto == IPPROTO_TCP`; otherwise, we have a serious bug (as you found out)
DTLS is quite complex; because it can be over TCP too but we shouldn't access to `flow.l4.tcp.tls` in that case: I might double check tomorrow...

utoni: you were right, DTLS also accesses `tls_quic` instead of `tls` .. corrected

IvanNardi (approved):
Thanks for fixing this issue!

IvanNardi:
Slightly updated the title

PR #3012, opened 2025-10-24
Fix broken header install

utoni:
* header files are expected to reside in prefix/includedir/ndpi/ instead of prefix/includedir/

Describe changes:
broken since 9aeb80f902


IvanNardi (approved):
It makes sense. Thanks for fixing that

PR #3053, opened 2025-12-06
test: rework main script

IvanNardi:
The issue about `config.txt` files is that they contains paths:
* to configuration files, which are in the source tree
* to the dynamic plugins, which are in the build tree

Solution:
* copy all configuration files into the build tree
* all those paths are about the build tree
* tests run from the build tree, no from the source tree anymore


Makefile.am
```diff
@@ -72,6 +72,38 @@ pull:
 changelog:
 	cd $(top_srcdir) && git log --since={`curl -s https://github.com/ntop/ndpi/releases | grep datetime | head -n1 | egrep -o "[0-9]+\-[0-9]+\-[0-9]+"`} --name-only --pretty=format:" - %s" | grep  "^ " > CHANGELOG.latest
 
+# For out-of-tree builds: copy required configuration files
+# ndpiReader has some hard-coded paths in some unit tests: in the out-of-tree builds we need
+# some configuration files in the expected path
+all-local:
+	@if test "$(srcdir)" != . ; then \
```
utoni: Those steps are not required when building with: `--with-only-libndpi` right?

IvanNardi: You are right. Let me check if I am able to improve that

utoni (approved):
LGTM

PR #3058, opened 2025-12-08
CI: test `--disable-plugin-support` and add a job on FreeBSD

IvanNardi:


IvanNardi:
Shameless copied from https://github.com/utoni/nDPId/blob/main/.github/workflows/build-freebsd.yml. credits to @utoni !

PR #3093, opened 2026-01-18
Rename `register_dissector()` into `ndpi_register_dissector()`

### LINKED ISSUE #3063: Function name conflict: `register_dissector` conflicts with Wireshark
# Function name conflict: `register_dissector` conflicts with Wireshark

## Problem Description

The function `register_dissector` in nDPI conflicts with Wireshark's API function of the same name. This causes symbol collision when both libraries are linked together in the same project, leading to linking errors or unexpected behavior.

## Affected Code

**nDPI:**
- Declaration: `src/include/ndpi_private.h:641-645` <https://github.com/ntop/nDPI/blob/5.0/src/include/ndpi_private.h#L641>
- Implementation: `src/lib/ndpi_main.c:6602`
- Usage: Called by 256+ protocol dissector files in `src/lib/protocols/`

**Wireshark:**
- Declaration: `epan/packet.h:603-604` <https://gitlab.com/wireshark/wireshark/-/blob/v4.6.2/epan/packet.h?ref_type=tags#L603>
```c
WS_DLL_PUBLIC dissector_handle_t register_dissector(const char *name, dissector_t dissector, const int proto);
```

## Impact

When building a project that uses both nDPI and Wireshark libraries:
- **Linker errors**: Multiple definition errors during linking
- **Symbol collision**: Ambiguous symbol resolution
- **Runtime issues**: Wrong function may be called if symbols are resolved incorrectly

## Proposed Solution

Rename the function to `ndpi_register_dissector` to follow nDPI's naming convention (prefixing internal functions with `ndpi_`) and avoid conflicts with other libraries.

### Benefits
1. ✅ Eliminates symbol collision with Wireshark
2. ✅ Follows nDPI's naming convention (consistent with other nDPI functions)
3. ✅ Makes it clear this is an nDPI-specific function
4. ✅ Better namespace isolation

## Suggested Changes

1. Rename function declaration in `src/include/ndpi_private.h`:
   ```c
   void ndpi_register_dissector(char *dissector_name, ...);
   ```

2. Rename function implementation in `src/lib/ndpi_main.c`:
   ```c
   void ndpi_register_dissector(char *dissector_name, ...) { ... }
   ```

3. Update all call sites in `src/lib/protocols/*.c` files (256+ files):
   ```c
   ndpi_register_dissector("ProtocolName", ndpi_struct, ...);
   ```

## Environment

- **nDPI version**: 5.0
- **Wireshark version**: 4.6.2
- **Issue type**: Symbol collision / Naming conflict

## Additional Notes

This is a common issue when integrating multiple network analysis libraries. Using library-specific prefixes (like `ndpi_`) is a best practice to avoid such conflicts.

IvanNardi:
Try to avoid collisions with similar Wireshark function: see #3063. Note that this function is now used by plugins, so it is somehow part of the public API anyway.

Close #3063

### DIFF
--- a/src/include/ndpi_private.h
+++ b/src/include/ndpi_private.h
@@ -654,7 +654,7 @@ struct ndpi_detection_module_struct {
 int is_proto_enabled(struct ndpi_detection_module_struct *ndpi_str, int protoId);
 int is_flowrisk_enabled(struct ndpi_detection_module_struct *ndpi_str, ndpi_risk_enum flowrisk_id);
 
-void register_dissector(char *dissector_name, struct ndpi_detection_module_struct *ndpi_str,
+void ndpi_register_dissector(char *dissector_name, struct ndpi_detection_module_struct *ndpi_str,
                         void (*func)(struct ndpi_detection_module_struct *,
                                      struct ndpi_flow_struct *flow),
                         const NDPI_SELECTION_BITMASK_PROTOCOL_SIZE ndpi_selection_bitmask,
--- a/src/lib/ndpi_main.c
+++ b/src/lib/ndpi_main.c
@@ -6613,7 +6613,7 @@ int load_protocols_file_fd(struct ndpi_detection_module_struct *ndpi_str, FILE *
 
 /* ******************************************************************** */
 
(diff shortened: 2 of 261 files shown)


lucaderi:
Looks like a good idea

Issue #3021, opened 2025-11-08
Extend Runtime Protocol Definition with TLS Certificate


lucaderi:
Currently in the protocol file (https://github.com/ntop/nDPI/blob/dev/example/protos.txt) it is not possible to binda protocol to a TLS certificate. 

In https://github.com/ntop/nDPI/blob/dev/src/lib/ndpi_content_match.c.inc protocols are defined as follows:

```
static ndpi_tls_cert_name_match tls_certificate_match [] = {
  { "CN=AnyDesk Client",               NDPI_PROTOCOL_ANYDESK     },
  { "O=philandro Software GmbH",       NDPI_PROTOCOL_ANYDESK     },
...
}
```

This is to request to extend protocols file wi enable the creation of new protocols based on the certificate name (i.e. allow definition of new protocols other than using ```tls_certificate_match```)

PR #3099, opened 2026-01-25
adding tls certificate match to proto.txt

kalindafab:

- [ x] I have read the contributing guidelines at https://github.com/ntop/nDPI/blob/dev/CONTRIBUTING.md

Link to the related [issue](https://github.com/ntop/nDPI/issues):https://github.com/ntop/nDPI/issues/3021

**Describe changes:**
Adds support for dynamic TLS certificate–based subprotocol detection using a new @tls_cert: rule syntax. Rules are parsed at runtime, stored in the detection module, and matched against TLS certificate Subject/Issuer fields during inspection.

The implementation follows existing dynamic rule patterns as @ja4:, @ndpifp:), includes proper initialization/cleanup.

**Request for Review:**

@lucaderi  Hi! I'm working on implementing this feature but running into a protocol name parsing issue. The error messages suggest the protocol name is being extracted from the certificate pattern rather than from before the `@tls_cert:` part.

Could you point me to how protocol names are supposed to be extracted for new attribute types? Looking at the existing `@ja4:` and `@ndpifp:` handlers, I tried to follow the same pattern, but something isn't working correctly.

Any guidance would be appreciated! Happy to make any changes needed.


IvanNardi:
@kalindafab, could you add a simple example rule in `protos.txt` with a matching pcap, please? Hopefully, with a new, self-signed certificate, to try to avoid matching with "real" traffic...
In other words, just a unit test for this new feature....

kalindafab:
I added unit test but the parsing issue is so complex :) 
Screenshot from 2026-01-27 13-04-29

lucaderi:
We need to match the certificate hash, not the .pem certificate.

PR #3115, opened 2026-02-14
Fix MinGW pcap check when using --with-only-libndpi

### LINKED ISSUE #3114: configure: --with-only-libndpi triggers pcap check on MinGW instead of skipping it
## Bug

On MinGW/MSYS2, passing `--with-only-libndpi` to configure triggers a `PKG_CHECK_MODULES` check for `libpcap`, causing configure to fail if libpcap-dev is not installed. This is the opposite of the intended behavior — building only the library should not require pcap.

## Root cause

In `configure.ac` line 344, the `AS_IF` condition for the MinGW pcap check is inverted:

```autoconf
AS_IF([test "${enable_npcap+set}" != set && test "${with_only_libndpi+set}" != set],,
      [PKG_CHECK_MODULES([PCAP], [libpcap], [PCAP_LIB="" PCAP_INC="${pkg_cv_PCAP_CFLAGS}"])
       AC_CHECK_LIB([pcap], [pcap_open_live],, [AC_MSG_ERROR([Missing msys2/mingw libpcap library. ...])])])
```

The empty then-body (`,,`) means:
- When **neither** flag is set (default): pcap check **skipped** (correct)
- When `--with-only-libndpi` **is** set: pcap check **runs** (incorrect)

Compare with the non-MinGW path (line 365) which correctly skips pcap when `--with-only-libndpi` is set:

```autoconf
elif test "${with_only_libndpi+set}" != set; then :
   AC_CHECK_LIB([pcap], [pcap_open_live], [PCAP_LIB="-lpcap"])
```

## Error output

```
checking for libpcap... no
configure: error: Package requirements (libpcap) were not met:

Package 'libpcap' not found
```

## Steps to reproduce

On MSYS2/MinGW64:
```bash
./autogen.sh
./configure --with-only-libndpi
```

## Suggested fix

The MinGW pcap check should only run when npcap is explicitly enabled and `--with-only-libndpi` is not set.

drnpkr:
## Summary

On MinGW/MSYS2, `--with-only-libndpi` triggers the pcap check instead of skipping it. This is the opposite of the non-MinGW path behavior (line 355) which corretcly skips pcap when building only the library.

## Change

One-line fix in `configure.ac`: flip the `AS_IF` condition so the pcap check runs only when npcap is explicitly enabled and `--with-only-libndpi` is not set.

Before: empty then-body (`,,`) meant the check ran whenever *either* flag was set
After: check runs in the then-body, only when `enable_npcap` is set AND `with_only_libndpi` is not set

Fixes https://github.com/ntop/nDPI/issues/3114

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -331,7 +331,7 @@ case "$host" in
       LIBS="${LIBS} -lws2_32"
       BUILD_MINGW=1
       EXE_SUFFIX=".exe"
-      AS_IF([test "${enable_npcap+set}" != set && test "${with_only_libndpi+set}" != set],,
+      AS_IF([test "${enable_npcap+set}" = set && test "${with_only_libndpi+set}" != set],
             [PKG_CHECK_MODULES([PCAP], [libpcap], [PCAP_LIB="" PCAP_INC="${pkg_cv_PCAP_CFLAGS}"])
              AC_CHECK_LIB([pcap], [pcap_open_live],, [AC_MSG_ERROR([Missing msys2/mingw libpcap library. Install it with `pacman -S mingw-w64-x86_64-libpcap' (msys2).])])])
       ;;


drnpkr:
Closing in favor of #3116 which handles the --disable-npcap case correctly.

PR #3128, opened 2026-03-06
fuzz: reset flow state and populate headers in fuzz_is_stun

OwenSanzas:
## Summary

Fix three issues in `fuzz/fuzz_is_stun.c` (which also affects the TCP variant via symlink) that reduce fuzzing effectiveness:

1. **Stale flow state**: `ndpi_flow` is static and never reset between calls, so state from previous fuzz inputs leaks into subsequent ones. This makes fuzzing results non-reproducible and prevents proper per-input isolation. Fixed by adding `memset()` at the start and `ndpi_free_flow_data()` at the end of each call.

2. **Invalid IP header**: `iph` is all zeros (version=0, ihl=0, protocol=0), which is not a valid IPv4 header and causes `is_stun()` to skip protocol-specific code paths. Fixed by setting version=4, ihl=5, and protocol=17 (UDP) or 6 (TCP).

3. **Zero transport ports**: `udph`/`tcph` are all zeros (port=0), so STUN port-aware code paths are never exercised. Fixed by setting source and destination ports to 3478 (STUN default port).

## Fuzzing Results (5-minute comparison, ASAN + libFuzzer fork mode)

| Metric | Original | Fixed | Improvement |
|--------|----------|-------|-------------|
| Edge coverage | 2510 | 2976 | **+18.6%** |
| Features | 3640 | 5615 | **+54.3%** |
| Corpus size | 475 | 889 | **+87.2%** |

The fixed version reaches new functions that the original never exercises, including `signal_search_into_cache`, `signal_add_to_cache`, `ndpi_check_dga_name`, `ac_domain_match_handler`, `ndpi_match_bigram`, `ndpi_match_trigram`, and others deep in the STUN and protocol classification code.

## Changes

- `fuzz/fuzz_is_stun.c`: Reset flow, populate IP/transport headers, free flow data
- `fuzz/fuzz_is_stun_tcp.c` is a symlink to `fuzz_is_stun.c`, so the fix applies to both variants via the existing `STUN_TCP` ifdef

### DIFF
--- a/fuzz/fuzz_is_stun.c
+++ b/fuzz/fuzz_is_stun.c
@@ -2,6 +2,8 @@
 #include "ndpi_private.h"
 #include "fuzz_common_code.h"
 
+#include <arpa/inet.h>
+
 static struct ndpi_detection_module_struct *ndpi_struct = NULL;
 static struct ndpi_flow_struct ndpi_flow;
 struct ndpi_iphdr iph;
@@ -27,7 +29,32 @@ int LLVMFuzzerTestOneInput(const uint8_t *data, size_t size) {
 
   if (ndpi_struct == NULL) {
     fuzz_init_detection_module(&ndpi_struct, NULL, path);
+
+    /* Initialize IP/transport headers once — they are never modified */
+    memset(&iph, 0, sizeof(iph));
+    iph.version = 4;
+    iph.ihl = 5;
+#ifdef STUN_TCP
+    iph.protocol = 6; /* TCP */
+#else
+    iph.protocol = 17; /* UDP */
+#endif
+    iph.saddr = htonl(0x0A000001); /* 10.0.0.1 */
+    iph.daddr = htonl(0x0A000002); /* 10.0.0.2 */
+
+#ifdef STUN_TCP
+    memset(&tcph, 0, sizeof(tcph));
+    tcph.source = htons(3478);
+    tcph.dest = htons(3478);
+#else
+    memset(&udph, 0, sizeof(udph));
+    udph.source = htons(3478);
+    udph.dest = htons(3478);
+#endif
   }
+
+  /* Reset flow state to avoid stale data from previous calls */
+  memset(&ndpi_flow, 0, sizeof(ndpi_flow));
 
   packet = &ndpi_struct->packet;
   packet->payload = data;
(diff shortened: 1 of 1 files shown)


IvanNardi:
@OwenSanzas thank you for the contribution. Could you sign the CLA, please?

OwenSanzas:
@IvanNardi Thanks for the approval. I just submitted the CLA.

OwenSanzas:
Thanks\! I've signed the CLA via the Google Form.

fuzz/fuzz_is_stun.c
```diff
@@ -29,6 +31,33 @@ int LLVMFuzzerTestOneInput(const uint8_t *data, size_t size) {
...
+  memset(&tcph, 0, sizeof(tcph));
+  tcph.source = htons(3478);
+  tcph.dest = htons(3478);
+#else
+  /* Populate UDP header with STUN default port */
+  memset(&udph, 0, sizeof(udph));
+  udph.source = htons(3478);
+  udph.dest = htons(3478);
+#endif
+
```
IvanNardi, line 58: Interesting stuff...
Maybe we can initialize iph, tcph and udph only once (like the detection module above) instead of at every iterations? We are never going to update them...
@OwenSanzas what do you think?

OwenSanzas: Good idea — done. Moved the `iph`, `udph`/`tcph` initialization into the one-time `if (ndpi_struct == NULL)` block, since these values are constant across iterations.

Verified with a 5-minute ASAN fork-mode benchmark (original upstream vs. this PR before/after the change):

| Metric | Upstream | PR (per-call init) | PR (one-time init) |
|--------|----------|--------------------|--------------------|
| Edge cov | 2070 | 2218 | **2224** |
| Features | 2661 | 3253 | **3352** |
| Corpus | 353 | 495 | **495** |
| exec/s | 103274 | 92677 | 88578 |

Coverage is equal or slightly better after the move — no regressions.

IvanNardi (approved):
Thank you!

IvanNardi:
@OwenSanzas may I ask you why/how you bumped into nDPI and this specific fuzzer?
Thanks for your contribution

OwenSanzas:
@IvanNardi I am a Phd student in cybersecurity and also a contributor to some big OSS. Everyday I need to review mountains of AI-generated bug reports and many of them are based on oss-fuzz. I am frustrated that most of them are False Positives and really hard to identify. Therefore I want to make current fuzzing suite generate less FP by fixing/updating it. The PR, issue page and the report are generated by a AI-based system with coverage & quality verification. It is scanning all oss-fuzz related repos for problematic fuzzers. I want to save maintainers' time from FP vulnerabilities and make software safer from strengthen current fuzzers. A change in problematic fuzzer may prevent many FP bug reports.

PR #3179, opened 2026-05-06
Add new dissector for iris protocol

BryanProg:

Describe changes:


src/lib/protocols/iris.c
```diff
@@ -0,0 +1,256 @@
...
+                    IS_TWO_FACTOR_ENABLED = 17714,
+                    GATEWAY_INIT = 18759,
+                    MESSAGE_JAVA_OBJECT_CREATED = 14681,
+                    CLOSE_STATEMENT = 21827,
+                    COMPARE_TIMESTAMP = 22083,
+                    TOGGLE_SYNCHRONOUS_COMMIT = 21332,
+                    READ_UNCOMMITTED = 21842,
+                    READ_COMMITTED = 17234,
+                    RESET_CONNECTION = 20050,
+                    OPEN_STREAM = 21327 };
```
IvanNardi, line 118: Do you have any explicit reference for this enum and for the packet format?
If so, please add a link here

src/lib/protocols/iris.c
```diff
@@ -0,0 +1,256 @@
...
+        case GATEWAY_INIT:
+        case MESSAGE_JAVA_OBJECT_CREATED:
+        case CLOSE_STATEMENT:
+        case COMPARE_TIMESTAMP:
+        case TOGGLE_SYNCHRONOUS_COMMIT:
+        case READ_UNCOMMITTED:
+        case READ_COMMITTED:
+        case RESET_CONNECTION:
+        case OPEN_STREAM:                    
+            ndpi_set_detected_protocol(ndpi_struct, flow, NDPI_PROTOCOL_IRIS, NDPI_PROTOCOL_UNKNOWN, NDPI_CONFIDENCE_DPI_PARTIAL);
```
IvanNardi: `NDPI_CONFIDENCE_DPI`
Can you also add something like NDPI_LOG_INFO(ndpi_struct, "Found iris\n"), please?

src/lib/protocols/iris.c
```diff
@@ -0,0 +1,256 @@
...
+            return;
+
+        default:
+            break;
+    }
+
+    if (message_length == (packet->payload_packet_len - 14U)) {
+        ndpi_set_detected_protocol(ndpi_struct, flow, NDPI_PROTOCOL_IRIS, NDPI_PROTOCOL_UNKNOWN, NDPI_CONFIDENCE_DPI_PARTIAL);
```
IvanNardi: `NDPI_CONFIDENCE_DPI`
Can you also add something like `NDPI_LOG_INFO(ndpi_struct, "Found iris\n")`, please?

src/lib/protocols/iris.c
```diff
@@ -0,0 +1,256 @@
...
+    } else {
+        NDPI_EXCLUDE_DISSECTOR(ndpi_struct, flow);
+    }
+}
+
+void init_iris_dissector(struct ndpi_detection_module_struct *ndpi_struct)
+{
+  ndpi_register_dissector("iris", ndpi_struct,
+                     ndpi_search_iris,
+                     NDPI_SELECTION_BITMASK_PROTOCOL_V4_V6_TCP_WITHOUT_RETRANSMISSION,
```
IvanNardi: `NDPI_SELECTION_BITMASK_PROTOCOL_V4_V6_TCP_WITH_PAYLOAD_WITHOUT_RETRANSMISSION`

src/lib/protocols/iris.c
```diff
@@ -0,0 +1,256 @@
...
+}
+
+/* this detection also works asymmetrically */
+static void ndpi_search_iris(struct ndpi_detection_module_struct *ndpi_struct, struct ndpi_flow_struct *flow)
+{
+    struct ndpi_packet_struct *packet = &ndpi_struct->packet;
+
+    NDPI_LOG_DBG(ndpi_struct, "search Iris\n");
+
+    if(packet && packet->tcp && packet->payload_packet_len > 14U){
```
IvanNardi: `if(packet->payload_packet_len > 14U)`: packet and packet->tcp are always valid pointers here

IvanNardi (changes requested):
@BryanProg, thanks for your contribution
Some small adjustments are required; see below.
Can you provide a pcap as example of this new protocol?

src/lib/protocols/iris.c
BryanProg: @IvanNardi Sorry, I don't have any documentation because this is a proprietary protocol and there is no information about it on the network. I performed this analysis using a reverse engineering mechanism. I set up an Iris database, inserted some transactions, and was able to observe this information.

src/lib/protocols/iris.c
BryanProg: Done!

src/lib/protocols/iris.c
BryanProg: Done!

src/lib/protocols/iris.c
BryanProg: Done!

src/lib/protocols/iris.c
BryanProg: Done! Thanks. Good observation.

IvanNardi:
@BryanProg , I rework the detection logic (the protocol is found if the message type is valid AND the length is valid too) to avoid the false positives with SSH traffic, and I simplified the pcaps, keeping only the two iris flows.
Do you agree with that?

BryanProg:
@IvanNardi  I agree with you, but the `message_type` only exists in request packets. When it comes to a response message, the message type is not included. That's why I used the TCP payload length and the Iris header length to check the response messages, hence port 1972. I believe only Iris uses that port by default.

If we continue like this, only request messages will be recognized.

IvanNardi:
@BryanProg, looking closely at the pcap, it seems that:
* length check can be performed on req and respo boths
* the message type is present only in the requests, but in the responses these two bytes are always zero

With this logic we should be able to detect both directions, without false positives (hopefully)
What do you think?

[I fixed the detection on big endian archs]

BryanProg:
@IvanNardi These two bytes in the response that you saw are the request's error code. When it is 0 it is because there was success, but I do not know all the other values ​​of the error code and that is why I did not check with this information.

So I compared the size of the iris payload (tpc payload size minus 14bytes which is the iris header) with the information that the iris header brings (the first 4 bytes of the iris header are the size of the iris payload information) and I only took advantage of this protocol (In main use, this protocol only uses port 1972) using this port 1972 with the TCP method. Don't you think that's all it takes? Do you think there could be false positives?

Even if I define that in these cases the nDPI trust is NDPI_CONFIDENCE_MATCH_BY_PORT or NDPI_CONFIDENCE_DPI_PARTIAL??


Thanks for to fix the detection on big endian archs

IvanNardi:
@BryanProg, let me check if I understood it right.
You are proposing to:
* check only TCP flows on port 1972
* use only the packet length logic
* remove the checks on message types

Is that correct? [BTW, I think that should work...]

BryanProg:
@IvanNardi 

I propose the following:

- Check only TCP flows on port 1972 -> YES

- Use only packet length logic -> More/less. Here we can to use packet length logic AND message type check for REQ messages, but for RESP messages use ONLY packet length logic, as message type check is not available for RESP.

- Remove message type checks -> Yes, But only RESP message. To ignore message type for RESP messages and rely on packet length logic and TCP port == 1972 this cases.

That's it. What do you think?


NOTE: For REQ message to use the message_type lends more confidence to this packet. Because we are certain that it uses the Iris protocol.

IvanNardi:
@BryanProg, what do you think of the latest version?

nicmorais:
> @BryanProg, what do you think of the latest version?

Greetings Ivan!

I'm a coworker of Bryan. He's on vacation.

One thing: I'd rename the `check_msg_type_or_error_code()` function to something like `iris_check_msg_type_or_error_code()`, just to make it clear it is not a common function, but something related only to the Iris protocol. Up to you.


Otherwise, looks good to me.

Also, I think it is worth mentioning the Iris database can be tested for free with the [intersystems/iris-community](https://hub.docker.com/r/intersystems/iris-community) Docker image (in case someone wants to assist in reverse engineering the protocol).

IvanNardi:
> One thing: I'd rename the `check_msg_type_or_error_code()` function to something like `iris_check_msg_type_or_error_code()`, just to make it clear it is not a common function, but something related only to the Iris protocol. Up to you.

Done

> Also, I think it is worth mentioning the Iris database can be tested for free with the [intersystems/iris-community](https://hub.docker.com/r/intersystems/iris-community) Docker image (in case someone wants to assist in reverse engineering the protocol).

Added to the documentation

Thanks!

IvanNardi:
Thanks, everyone!

PR #3193, opened 2026-07-13
Add HBO Max/Max streaming detection

cmonroe:

Describe changes:
Extend NDPI_PROTOCOL_HBO with the domains Warner Bros. Discovery's Max service now uses, so HBO Max video sessions are classified instead of falling back to generic TLS/QUIC:

  * h264.io       WBD video CDN - QUIC video streams
  * max.com       Max app - API and media hosts
  * discomax.com  shared WBD streaming API and image hosts

These domains are WBD-wide.. CNN, Discovery+ and TNT Sports streaming traffic using the same infrastructure is also reported as HBO. Direct cnn.com/cnn.net website traffic still maps to NDPI_PROTOCOL_CNN, since SNI matching takes precedence over certificate-SAN matching.

Add a small sample capture containing the QUIC video flow plus TLS/QUIC control flows, and its expected result.


IvanNardi (approved):
Thank you for your contributions.
I simply merged your pcap into an existing one. As a general rule, if the new traffic is *only* about a "domain match", I prefer to keep it into the existing `pcaps/sitesX.pcap` files

cmonroe:
> Thank you for your contributions. I simply merged your pcap into an existing one. As a general rule, if the new traffic is _only_ about a "domain match", I prefer to keep it into the existing `pcaps/sitesX.pcap` files

Got it.. I was torn on this but it totally makes sense. I'll follow your recommendation for future PRs. Thank you!

PR #3194, opened 2026-07-13
Add Webex STUN sub-classification via SOFTWARE attribute

cmonroe:

Describe changes:
Webex call media uses ICE/STUN. The STUN control flows were reported as STUN/Network because the STUN dissector had no Webex sub-classification hook even though the media servers already fall in the known Webex IP ranges.

Webex (formerly Cisco Spark) ICE agents advertise the STUN SOFTWARE attribute (0x8022) with the value "Cisco-Spark". Parse it in is_stun() and set the app protocol to the existing Webex protocol so the flow classifies as STUN.Webex (VoIP) with DPI confidence.

Add a minimal test capture and its expected result.


IvanNardi (approved):
Thank you

PR #3195, opened 2026-07-13
Add NetMotion Mobility (IMP) protocol detection

cmonroe:

Describe changes:
NetMotion Mobility (Mobility XE) tunnels clients to the Mobility server using the Internet Mobility Protocol (IMP), by default over UDP/5008. The payload is encrypted, but every datagram carries a fixed header: a 0x4c magic, a self-inclusive length prefix, a per-direction connection id plus sequence, and a per-direction session token. Detection verifies these framing invariants are constant within a direction and differ across directions, following the same stateful approach as the Hamachi dissector.

Add NDPI_PROTOCOL_NETMOTION (VPN category) with the netmotion dissector and its unit test. The remaining unit-test results are regenerated because the new protocol id shifts the auto-numbered custom-rule ids and adds one dissector call per UDP flow.

This feels a little hacky to me but I didn't see any other good way.  I'm happy to look into it more if you don't like it.


doc/protocols.rst
```diff
@@ -4327,3 +4327,12 @@ References: `InterSystems IRIS official site: <https://www.intersystems.com/prod
...
 References: `Official site: <https://libp2p.io>`_
+
+
+.. _Proto_477:
+
+`NDPI_PROTOCOL_NETMOTION`
+=========================
+NetMotion Mobility (formerly Mobility XE) is an enterprise mobile VPN that keeps sessions persistent across network changes, widely deployed in public-sector and first-responder fleets. Clients tunnel to the Mobility server using the Internet Mobility Protocol (IMP), by default over UDP/5008.
+
+References: `Official site: <https://www.netmotionsoftware.com>`_
```
IvanNardi, line 4338: Unrelated to your PR, but it is funny that the site of a VPN provider triggers:

Image

src/lib/protocols/netmotion.c
```diff
@@ -0,0 +1,136 @@
...
+ *   [3]     message type/flags (small)
+ *   [4:6]   16-bit connection id, constant per direction, differs by direction
+ *   [6:8]   16-bit per-direction sequence counter
+ *   [8:12]  peer connection id + sequence (mirror of the other direction)
+ *   [12:16] 32-bit session token, constant per direction, differs by direction
+ *
+ * The payload is encrypted; detection keys on the framing invariants only,
+ * verifying that the connection id and session token are constant within a
+ * direction and differ across directions (same approach as hamachi.c).
+ */
```
IvanNardi, line 50: Any reference for this packet format? Or is it reverse engineering?

src/lib/ndpi_main.c
```diff
@@ -7559,6 +7564,7 @@ static int dissectors_init(struct ndpi_detection_module_struct *ndpi_str) {
 
   /* Hamachi */
   init_hamachi_dissector(ndpi_str);
+  init_netmotion_dissector(ndpi_str);
```
IvanNardi: Other than the detection logic, is there any relations between IMP and Hamachi?
Anyway, could you move this initialization to the end of the list, please?

doc/protocols.rst
cmonroe: I found this odd (unfortunate?) as well and dug a bit more.. it seems like NetMotion was maybe acquired by a company called Aboslute so I guess they abandoned the old domain? The only clients I've seen running are all still branded NetMotion and point to this old URL. I can dig more if you want but the [new owner](https://www.absolute.com) has very little useful info on their site.

src/lib/protocols/netmotion.c
cmonroe: Reverse engineering only.. just staring at a few pcap files from different sessions. I very well could have misinterpreted something :)

src/lib/ndpi_main.c
cmonroe: They share nothing in common.. the only reason for the mention of Hamachi is that I followed this as an example. I can remove mention from the comment if you'd like. Either way I moved init to the end of the list.

cmonroe:
Oops, I will fix the error here shortly.

doc/protocols.rst
IvanNardi: It is fine as it is

IvanNardi (approved):
Thanks!
I simply rebased the code

PR #3199, opened 2026-07-16
activision: classify Warzone game session flows

cmonroe:

Describe changes:
Warzone game sessions use 0x46/0x47 direction markers that are protocol-level and unrelated to nDPI's IP-based flow canonicalization, so they must be checked independently of packet_direction.

Also removes a hardcoded client IP check (192.168.0.21) that only matched one specific capture, and accepts retransmitted 0x0c02/0x0d02 initial packets in subsequent-packet checks to prevent false exclusion from server-side application-level retransmission.

Update the activision.pcap golden accordingly: the flows are still classified as Activision by DPI, but detection now happens on packet 5 via the packet_counter check rather than on the first packet, so DPI packets goes 1 -> 5 and FPC reports Unknown. The first-packet match it replaces only ever fired for the one capture whose client was 192.168.0.21.

### DIFF
--- a/src/lib/protocols/activision.c
+++ b/src/lib/protocols/activision.c
@@ -49,43 +49,47 @@ static void ndpi_search_activision(struct ndpi_detection_module_struct *ndpi_str
     return;
   }
 
+  u_int16_t magic = ntohs(get_u_int16_t(packet->payload, 0));
+
+  /*
+   * 0x46/0x47 variant: payload direction markers are protocol-level and
+   * unrelated to nDPI's flow direction assignment, so handle them before
+   * any direction-dependent checks.
+   */
+  if (magic == 0x4600 || magic == 0x4700)
+  {
+    if (flow->packet_counter > 4)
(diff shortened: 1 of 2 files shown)


IvanNardi:
@cmonroe, always reverse engineering or are there some (public) references?
I am not sure that we don't have false positives having a so simple check (but we want multiple packets...). Anyway we will think about that only if we find some issues...

IvanNardi (approved):

cmonroe:
Yes, this one was 100% reverse engineering and the one I'm least certain about. But it's an improvement over what was present at least.

That said, I've had this patch in my tree for a couple of months and never observed a false positive. Not to say it won't exist but it's at least not easy to spot if so.

I'm happy to grab fresh traces and take another look if anything comes up.

IvanNardi:
> I'm happy to grab fresh traces and take another look if anything comes up.

it should be useful since we don't have usually a lot of traces about games... :(

PR #3203, opened 2026-07-21
geoip: support flat MMDB key schemas

cmonroe:

Describe changes:
MaxMind GeoIP databases use a nested key schema (e.g.
country -> iso_code, continent -> names -> en) while other
providers like IPLocate use flat keys (country_code,
continent_code, asn as string, org). Add fallback lookups
so nDPI works with either schema transparently.

Also include a small fix discovered when testing GeoIP.


IvanNardi (approved):
We don't really test the geo code, but the change seems fine.
Thank you

PR #3221, opened 2026-08-13
Add Free Fire protocol

Dulend:
## Summary
- Add `NDPI_PROTOCOL_FREEFIRE` (ID 490)
- UDP gameplay: ports 10011–10015 + payload magic
- TCP control: `vj73p` / `freefiremobile` / `FREE FIRE`
- Hosts: `freefiremobile.com`, `100067.connect.garena.com`, `100067.msdk.garena.com`, `ggpolarbear.com`
- Unit tests: `freefire.pcap`, `freefire_udp.pcap`

## Test plan
- [ ] `make check`
- [ ] ndpiReader on freefire_handshake / gameplay / gameplay2


lucaderi:
@Dulend Please sign (check) the below before submitting the Pull Request:

- [ ] I have signed the ntop Contributor License Agreement at https://github.com/ntop/legal/blob/main/individual-contributor-licence-agreement.md
- [ ] I have read the contributing guidelines at https://github.com/ntop/nDPI/blob/dev/CONTRIBUTING.md

Dulend:
@lucaderi Thanks — confirmed:

- [x] Signed the ntop CLA
- [x] Read CONTRIBUTING.md
- [x] Updated documentation (`doc/protocols.rst`, Proto_490 / Free Fire)

Ready for review.

src/lib/protocols/freefire.c
```diff
@@ -0,0 +1,116 @@
+/*
+ * freefire.c
+ *
+ * Free Fire
+ *
+ * Copyright (C) 2026 - ntop.org
+ * Copyright (C) 2026 - V.G <v.gavrilov@securitycode.ru>
```
IvanNardi: Unrelated. Please remove this line

src/lib/protocols/freefire.c
```diff
@@ -0,0 +1,116 @@
...
+static int is_freefire_gameplay_port(u_int16_t port)
+{
+  return (port >= 10011 && port <= 10015);
+}
+
+static int is_freefire_udp_payload(const struct ndpi_packet_struct *packet)
+{
+  /* Gameplay datagrams commonly start with XX YY a5/a7 ... */
+  return (packet->payload_packet_len >= 8 &&
+          (packet->payload[2] == 0xa5 || packet->payload[2] == 0xa7));
```
IvanNardi: can we assume that this is true for every UDP packets?
Looking at your trace it seems that every packets has 0xa0 || 0xa5 || 0xa7
If that is true (can you check with other traces?)....

src/lib/protocols/freefire.c
```diff
@@ -0,0 +1,116 @@
...
+  } else if (packet->udp != NULL) {
+    u_int16_t sport = ntohs(packet->udp->source);
+    u_int16_t dport = ntohs(packet->udp->dest);
+
+    if ((is_freefire_gameplay_port(sport) || is_freefire_gameplay_port(dport)) &&
+        is_freefire_udp_payload(packet)) {
+      NDPI_LOG_INFO(ndpi_struct, "found Free Fire (UDP)\n");
+      ndpi_int_freefire_add_connection(ndpi_struct, flow);
+      return;
+    }
```
IvanNardi, line 124: ... here we can directly exclude the dissector

src/lib/protocols/freefire.c
```diff
@@ -0,0 +1,116 @@
...
+  if (packet->tcp != NULL) {
+    if (is_freefire_tcp_payload(packet)) {
+      NDPI_LOG_INFO(ndpi_struct, "found Free Fire (TCP)\n");
+      ndpi_int_freefire_add_connection(ndpi_struct, flow);
+      return;
+    }
+
+    if (flow->packet_counter >= 8)
+      NDPI_EXCLUDE_DISSECTOR(ndpi_struct, flow);
+  } else if (packet->udp != NULL) {
```
IvanNardi: No need to check `packet->udp` here: it is always non NULL here

src/lib/protocols/freefire.c
```diff
@@ -0,0 +1,116 @@
...
+
+  if (ndpi_memmem(packet->payload, packet->payload_packet_len,
+                  "freefiremobile", NDPI_STATICSTRING_LEN("freefiremobile")))
+    return 1;
+
+  if (ndpi_memmem(packet->payload, packet->payload_packet_len,
+                  "FREE FIRE", NDPI_STATICSTRING_LEN("FREE FIRE")))
+    return 1;
+
+  return 0;
```
IvanNardi, line 85: Can we find a faster logic, in particular a faster way to exclude the dissector on TCP, instead of waiting 8 packets pretty much for every TCP flows?
For example, it seems that:
* the first pkt C->S is always 770 bytes long
* on the first pkt C->S we have bytes 2,3,4,5 always 0
Can you check that on other traces, please?

The goal is to have a faster way to exclude the protocol for the general traffic, not necessarily a faster way to identify it...

src/lib/ndpi_main.c
```diff
@@ -7530,6 +7535,9 @@ static int dissectors_init(struct ndpi_detection_module_struct *ndpi_str) {
   /* Call of Duty: Mobile */
   init_cod_mobile_dissector(ndpi_str);
 
+  /* Free Fire */
+  init_freefire_dissector(ndpi_str);
+
```
IvanNardi: can you move it at the end of the list, please?

IvanNardi (changes requested):
Thanks for your contribution.
Take a look at the comment below, please

Dulend:
@IvanNardi Thanks — checked all three captures (handshake, gameplay, gameplay2).

UDP: every datagram on 10011–10015 has payload[2] in {0xa0, 0xa5, 0xa7} (no exceptions). Current code missed 0xa0. I’ll accept all three and exclude the UDP dissector on the first non-matching packet.

TCP: at session start (handshake pcap), first C→S is always 770 bytes with bytes 2–5 zero (both 39698 and 39800). gameplay*.pcap are mid-flow keepalives (`02 07`), so they don’t have that first packet. I’ll use the 770 + zeros check only to exclude other TCP on packet 1, and keep the keepalive / string match afterwards. Also removing the V.G copyright line and moving init_freefire_dissector to the end of the list.

Dulend:
@IvanNardi Addressed the review comments.

- Removed the unrelated V.G copyright line
- UDP: payload[2] is 0xa0 || 0xa5 || 0xa7 on every datagram in the handshake/gameplay traces; exclude on first miss
- Dropped the redundant packet->udp check
- TCP: at session start, first C→S is 770 bytes with zeros at [2..5] (handshake pcap, both 39698 and 39800). Used that to exclude other TCP on packet 1. Gameplay captures are mid-flow (02 07 keepalives), so they don’t have that first packet.
- Moved init_freefire_dissector to the end of the list

src/lib/protocols/freefire.c
```diff
@@ -0,0 +1,141 @@
...
+      return;
+    }
+
+    /* Fast-exclude generic TCP: first C->S is not the Free Fire setup frame */
+    if (packet->packet_direction == 0 &&
+        flow->packet_direction_counter[0] == 1) {
+      if (!is_freefire_tcp_setup(packet))
+        NDPI_EXCLUDE_DISSECTOR(ndpi_struct, flow);
+      return;
+    }
```
IvanNardi, line 105: Can we try to be smarter here? Something like (untested)
```
if(ndpi_seen_flow_beginning(flow)) {
  if(flow->packet_counter[0] == 1 && is_freefire_tcp_setup(packet)) {
    // FOUND
  } else {
    NDPI_EXCLUDE_DISSECTOR(ndpi_struct, flow);
  }
  return;
}
```
The idea is to check for the logic about the first packet if we are analyzing the entire session and to check for the more-complex logic only if we missed the initial part of the flow.
Does it make sense to you?

Dulend: @IvanNardi Thanks for the suggestion. Before I update the code, I want to make sure I understand the intent correctly.

My plan:

Keep the vj73p keepalive check outside (runs regardless of flow start)
Wrap the 770-byte fast-exclude inside if (ndpi_seen_flow_beginning(flow)) so it only fires when we saw the full 3-way handshake
For mid-flow captures (gameplay pcaps), fall through to the string matching (freefiremobile, FREE FIRE) instead of excluding immediately.

Thank you

IvanNardi: yes, that's the idea

src/lib/protocols/freefire.c
```diff
@@ -0,0 +1,144 @@
...
+  NDPI_LOG_DBG(ndpi_struct, "search Free Fire\n");
+
+  if (packet->tcp != NULL) {
+    if (packet->payload_packet_len == 10 &&
+        packet->payload[0] == 0xa0 &&
+        memcmp(&packet->payload[5], "vj73p", 5) == 0) {
+      NDPI_LOG_INFO(ndpi_struct, "found Free Fire (TCP)\n");
+      ndpi_int_freefire_add_connection(ndpi_struct, flow);
+      return;
+    }
```
IvanNardi: Can we remove this first block?
With:
* complete/initial session, we have dedicated path and we can't have a keepalive as firs pkt
* mid-flow: we already check for the exact same condition on is_freefire_tcp_payload

IvanNardi (commented):
Do you agree with the proposed change?
If so, I can take care of it while merging....

src/lib/protocols/freefire.c
Dulend: Yes, makes sense — that first block is redundant with is_freefire_tcp_payload(). Fine to remove it while merging. Thanks.

IvanNardi (approved):
Thank you!

Issue #3062, opened 2025-12-12
Ambiguous data in IP address lists


vel21ripn:
The main issue is with the AMAZON_AWS and AWS_EC2 addresses.
2284 addresses are simultaneously in different lists.
```
grep 3.4.16.0/20 *
ndpi_amazon_aws_ec2_match.c.inc: { 0x03041000 /* 3.4.16.0/20 */, 20, NDPI_PROTOCOL_AWS_EC2 },
ndpi_amazon_aws_match.c.inc: { 0x03041000 /* 3.4.16.0/20 */, 20, NDPI_PROTOCOL_AMAZON_AWS },
```

It's unclear how to resolve this.
The second issue is TOR and WHATSAPP_CALL. 1009 addresses are simultaneously in different lists.
IMHO, addresses from the TOR list should not be included in other lists.
The full list is attached.

[ip_dup_list.txt](https://github.com/user-attachments/files/24124460/ip_dup_list.txt)

IvanNardi:
@vel21ripn, sorry for the delay

> The main issue is with the AMAZON_AWS and AWS_EC2 addresses. 2284 addresses are simultaneously in different lists.

We had a similar issue in the past with Microsoft addresses; I will try to come up with a similar solution

> 
> It's unclear how to resolve this. The second issue is TOR and WHATSAPP_CALL. 1009 addresses are simultaneously in different lists. IMHO, addresses from the TOR list should not be included in other lists. The full list is attached.

We don't have a list of WHATSAPP_CALL addresses!!
I think the confusion is caused by the fact that we have two types of lists:
* the list of ips for protocol classification (list of ip/id pairs)
* the list of ips for setting flow risks (list of ip/flow_risk pairs)

I think you tried to load the second type of lists as if they were the first type....

vel21ripn:
I found a bug in the address list parsing scripts.
There was an error in the script that caused all address lists to be merged.
So far, the issue only persists with the AMAZON_AWS and AWS_EC2 addresses.

IvanNardi:
> There was an error in the script that caused all address lists to be merged.

If the issue is still present upstrem, could you push a PR with the fix, please?

vel21ripn:
This script is not part of nDPI.
If you're interested, check out my fork: https://github.com/vel21ripn/nDPI/tree/flow_info-4/utils/ndpi_network_list_compile*

To compile, run ndpi_network_list_compile.sh.
[...]

PR #3226, opened 2026-08-16
Issue 3062 aws list dedup

RajaMuhammadAwais:

What was wrong

1.The upstream ip-ranges.json double tags 143 prefixes as both AMAZON and a service such as EC2, S3 or CloudFront, so the same address range ends up in two generated lists (ndpi_amazon_aws_match.c.inc and the service matcher).

2.mergeipaddrlist.py merges adjacent prefixes into wider supernets inside each list. Merged supernets from a higher-precedence service can then fully cover prefixes kept in a lower-precedence list. For example, CloudFront rows merge into 52.82.128.0/19 while the generic AMAZON_AWS list still keeps its own 52.82.128.0/23, which is completely contained in the /19.

How it is fixed

1. ndpi_aws_ip_dedup.py (new): assigns every AWS prefix to exactly one service using a fixed precedence chain (EC2_INSTANCE_CONNECT > EC2 > S3 > CLOUDFRONT > DYNAMODB > API_GATEWAY > KINESIS_VIDEO_STREAMS > AMAZON_AWS generic), so the same address can never appear in two final lists.

2. ndpi_aws_prune_overlaps.py (new): post-processes the merged list files and removes any lower-precedence entry that is covered by a net in a higher-precedence list, catching the overlaps that cidr merging introduces.

3. aws_ip_addresses_download.sh (updated): wires the two helpers into the existing pipeline without touching anything else.

Two real bugs were found and fixed along the way: the merge call for EC2_INSTANCE_CONNECT + EC2 silently dropped the EC2 list (724 entries) because mergeipaddrlist.py reads only its first argument, and the pruner's merge-file paths were missing a dot so the prune pass never ran.

Independent verification: the regenerated include files contain 0 duplicate entries across all seven AWS matchers, API_GATEWAY and KINESIS raw prefixes are genuinely nested inside S3/EC2 supernets in the upstream data (214/214 and 105/105 prefixes respectively), the full unit, pcap and DGA test suites pass locally with the same flags as CI (cc, -g -O2, -Werror -DNDPI_EXTENDED_SANITY_CHECKS), and the python helpers pass py_compile. The SonarCloud pythonsecurity:S8707 annotations raised on earlier revisions of this PR were also resolved: the suffix argument is now validated against the fixed set of known values and all constructed paths are resolved and checked before any open() call.

Link to the related issue: https://github.com/ntop/nDPI/issues/3062


IvanNardi:
@RajaMuhammadAwais, could you provide a clean PR only about the AWS stuff, removing the DGA feature, please?

RajaMuhammadAwais:
done new PR you can check

IvanNardi:
See #3227

Issue #2231, opened 2023-12-26
Classifying RDP in http packets


frkn4129:
## Describe the bug
When I run ndpiReader with a pcap, it shows that an RDP session is created, but I see that there is http traffic from port 80 in the packets inside the pcap. Could there be an error here?

## nDPI Environment (please complete the following information):
* OS name: [e.g. Ubuntu].
* OS version: [e.g. 18.04]
* Architecture: [e.g. arm64]
* nDPI version or commit hash: Using nDPI (4.9.0-4439-99d48383)

``` shell
ndpiReader -i ../http_traffic.pcap 



nDPI Memory statistics:
	nDPI Memory (once):      37.30 KB     
	Flow Memory (per flow):  976 B        
	Actual Memory:           5.02 MB      
	Peak Memory:             5.02 MB      
	Setup Time:              9 msec
	Packet Processing Time:  0 msec

Traffic statistics:
	Ethernet bytes:        2000          (includes ethernet CRC/IFC/trailer)
	Discarded bytes:       0            
	IP packets:            20            of 20 packets total
	IP bytes:              1520          (avg pkt size 76 bytes)
	Unique flows:          2            
	TCP Packets:           20           
	UDP Packets:           0            
	VLAN Packets:          20           
	MPLS Packets:          0            
	PPPoE Packets:         0            
	Fragmented Packets:    0            
	Max Packet size:       67           
	Packet Len < 64:       18           
	Packet Len 64-128:     2            
	Packet Len 128-256:    0            
	Packet Len 256-1024:   0            
	Packet Len 1024-1500:  0            
	Packet Len > 1500:     0            
	nDPI throughput:       54.20 K pps / 41.35 Mb/sec
	Analysis begin:        22/Dec/2023 15:09:00
	Analysis end:          22/Dec/2023 15:09:00
	Traffic throughput:    538.62 pps / 420.80 Kb/sec
	Traffic duration:      0.037 sec
	DPI Packets (TCP):     8             (4.00 pkts/flow)
	Confidence: DPI        2             (flows)


Detected protocols:
	RDP                  packets: 20            bytes: 1520          flows: 2            


Protocol statistics:
	Acceptable                    1520 bytes

Risk stats [found 2 (100.0 %) flows with risks]:
	Known Proto on Non Std Port                  2 [50.0 %]
	Desktop/File Sharing                         2 [50.0 %]

	NOTE: as one flow can have multiple risks set, the sum of the
[...]

PR #3230, opened 2026-08-23
fix: avoid classifying HTTP responses as RDP

RajaMuhammadAwais:
- [x] I have updated the documentation (in doc/ ) to reflect the changes made (if applicable)

Link to the related [issue](https://github.com/ntop/nDPI/issues/2231 ):

#2231

Describe changes:

This pull request fixes a false-positive RDP classification reported in issue #2231.

The problem occurs when nDPI receives a TCP flow whose destination is an HTTP service on port 80. In the official reproduction capture, the client sends a 19-byte payload that begins with values matching the RDP TPKT and COTP negotiation structure. Because the payload satisfies the current RDP checks, the RDP dissector marks the flow as tentatively detected RDP by setting `rdp_protocol_detected`.

However, the payload is not followed by a valid RDP negotiation response. The server sends an HTTP response instead:

HTTP/1.1 400 Bad Request

The previous RDP logic treated any subsequent non-matching server packet as evidence that the flow was RDP. It then promoted the tentative state to a confirmed RDP classification and added RDP-specific risk information. This caused the complete HTTP flow to be incorrectly reported as RDP, even though the server clearly returned an HTTP response.

The issue was reproduced on the latest upstream `dev` branch using the official capture attached to #2231. Before the fix, the result was:

Detected protocols:
    RDP packets: 20 bytes: 1520 flows: 2

The implementation now handles this verified case explicitly. When a tentative RDP request is followed by a server-to-client payload beginning with the HTTP response signature `HTTP/`, and neither TCP endpoint uses the standard RDP port 3389, the RDP dissector is excluded instead of confirming RDP. This allows the HTTP dissector to classify the flow correctly.

The standard RDP port is intentionally excluded from this guard. This preserves legitimate RDP behavior on port 3389, including the existing RDP-over-HTTP scenario where the application payload can contain an HTTP response while the underlying service is still RDP.

After the fix, the official #2231 capture is classified as HTTP:

Detected protocols:
    HTTP packets: 20 bytes: 1520 flows: 2

Regression validation:

* The official #2231 HTTP false-positive capture changed from RDP to HTTP.
* The existing `rdp3.pcap` positive test continues to classify the traffic as RDP.
* The existing `rdp_over_http.pcapng` positive test continues to classify the traffic as RDP.
* The full nDPI regression suite passed with `NDPI_DISABLE_FUZZY=1 NDPI_FAIL_FAST=1 ./tests/do.sh`.
* The nDPI unit tests passed.
* `git diff --check` passed.

Files changed:

* `src/lib/protocols/rdp.c`: prevent promotion of a tentative RDP flow when the server clearly returns HTTP on a non-RDP port.
* `tests/cfgs/default/pcap/rdp_http_false_positive.pcap`: add the official issue capture as a negative regression test.
* `tests/cfgs/default/result/rdp_http_false_positive.pcap.out`: record the expected HTTP classification.


IvanNardi (approved):
Thank you

Issue #3213, opened 2026-08-04
UBSan "Type mismatch in operation" trap (SIGILL) on unaligned pointer-cast reads (DNS get16, ndpi_serializer)


1820893135-pixel:
## Describe the bug

nDPI performs 16/32-bit field reads by casting a byte pointer directly and dereferencing it (`*(u_int16_t*)&payload[*i]` / `ntohs(*((u_int16_t *)&buf[offset]))`). When the byte offset is odd the pointer is not 2-byte aligned, which is undefined behavior (CWE-843). A build with `-fsanitize=address,undefined -fno-sanitize-recover=all` traps on it (UBSan reports `Type mismatch in operation`) and the process dies with SIGILL.

Two concrete sites are reachable from untrusted input and were confirmed to crash:

1. **DNS dissector `get16()`** (`src/lib/protocols/dns.c`) — an attacker-controlled DNS message whose query/answer/additional walk leaves an odd offset before a 16-bit field is read.
2. **`ndpi_serializer.c`** deserializer helpers — `ndpi_init_serializer_ll` reads `*((u_int16_t*)&deserializer->buffer.data[offset])` on an unaligned offset (reached via `get_ndpi_flow_info` in `example/reader_util.c`).

The same pointer-cast pattern appears library-wide (`dns.c`, `ndpi_serializer.c`, `capwap.c`, `hpvirtgrp.c`, `tls.c`).

### Expected behavior

nDPI should read multi-byte fields from a byte buffer without relying on host alignment: unaligned loads must be performed with `memcpy` or byte assembly, so the library is portable to strict-alignment CPUs and never triggers UBSan.

### Obtained behavior

UBSan aborts the process (SIGILL / exit 134) on a single 45-byte DNS packet, both through the DNS dissector (`fuzz_dns_parse`) and through the full packet-processing path (`fuzz_ndpi_reader`). Under ASan/UBSan builds this is a remote DoS for any nDPI consumer that parses untrusted traffic.

## nDPI Environment

- **OS**: Ubuntu (OSS-Fuzz `base-builder` image, x86-64)
- **OS version**: Debian 12 bookworm based
- **Architecture**: x86-64 (the crash is the UBSan trap; on strict-alignment CPUs ARM/MIPS/RISC-V the same misaligned load faults natively)
[...]

PR #3231, opened 2026-08-23
fix: avoid unaligned reads in DNS and serializers

RajaMuhammadAwais:
- [x] Documentation update is not applicable for this low-level portability and sanitizer bug fix. The changed behavior is documented in this pull request and in the related issue.

Link to the related [issue](https://github.com/ntop/nDPI/issues/3213 ):

#3213

Describe changes:

## Summary

This pull request fixes undefined behavior and sanitizer failures caused by unaligned multi-byte reads in nDPI's DNS parsing and serializer code.

The issue is described in #3213. nDPI was reading 16-bit, 32-bit, and 64-bit values from byte buffers by casting byte pointers directly to integer pointers and dereferencing them. Such a cast requires the resulting address to satisfy the alignment requirement of the target integer type. Packet data and serialized data do not guarantee this alignment, particularly when a parser reaches an odd byte offset.

On x86-64, these reads may appear to work because the processor commonly permits unaligned access. However, the code still violates C alignment rules and UBSan correctly reports the operation as undefined behavior. On strict-alignment architectures such as ARM, MIPS, and some RISC-V configurations, the same access can fault natively. When nDPI is built with address and undefined-behavior sanitizers using `-fno-sanitize-recover=all`, the process terminates instead of continuing to process the packet.

## Reproduced DNS failure

The official issue contains three 45-byte malformed DNS payloads. Each payload causes the DNS parser to walk to an odd offset before reading a 16-bit field.

The first trigger contains an invalid query count and causes the DNS query walker to continue after the only valid query. The second trigger reaches the EDNS OPT additional-record fields at an odd offset. The third trigger uses a compression-pointer byte pattern that causes the name walk to reach an odd offset.

The triggers were wrapped in UDP PCAP files and replayed through an independently built nDPI reader with address and undefined-behavior sanitizers enabled.

Before the fix, the exact observed diagnostic was:

protocols/dns.c:151:13: runtime error: load of misaligned address ... for type 'u_int16_t', which requires 2 byte alignment

All three official trigger PCAPs reproduced the failure on the clean latest upstream `dev` checkout. This confirms that the issue is deterministic and not based on an assumed code path.

## Root cause in DNS parsing

The DNS helper `get16()` previously performed a direct cast and dereference similar to:

```c
u_int16_t v = *(u_int16_t *)&payload[*i];
The DNS parser also contained direct pointer-cast reads for EDNS payload size, record data length, option code, option length, and the packet cache key. These accesses can receive odd offsets or otherwise unaligned addresses when processing attacker-controlled packet data.
The fix routes these reads through alignment-safe helpers. The existing network-byte-order conversion behavior is preserved. Only the way the bytes are loaded has changed.
**Root cause in serializer handling**
The serializer deserialization helpers previously read integer values using direct casts such as:
*s = ntohs(*((u_int16_t *) &deserializer->buffer.data[offset]));
The serializer object itself was exposed through an opaque character-only structure. The implementation then viewed that storage as ndpi_private_serializer. Because the public storage type did not preserve the alignment of the private implementation type, the full reader path could also trigger an alignment diagnostic when the object was accessed.
The fix makes the opaque serializer storage an alignment-preserving union while keeping its storage size and byte-buffer layout compatible with the implementation. Integer deserialization now uses the same alignment-safe raw-byte helpers for 16-bit, 32-bit, and 64-bit values.
Implementation details
The following changes are included:
The raw integer access helpers in src/include/ndpi_define.h.in now load 16-bit, 32-bit, and 64-bit values with memcpy. This preserves the existing raw host-endian semantics without requiring the source address to be aligned.
The get_u_int8_t helper remains a byte access because an 8-bit value has no alignment requirement.
The DNS dissector in src/lib/protocols/dns.c now uses the safe accessors for get16(), EDNS fields, and the packet cache key.
The serializer implementation in src/lib/ndpi_serializer.c now uses safe accessors for unsigned and signed 16-bit, 32-bit, and 64-bit deserialization.
The public ndpi_serializer storage in src/include/ndpi_typedefs.h now preserves the alignment required by ndpi_private_serializer while retaining the existing storage buffer.
The three official issue triggers were added to fuzz/corpus/fuzz_dns_parse/ so future DNS fuzzing retains coverage for the malformed inputs that originally exposed the problem.

### CODE REVIEW

**Comment by IvanNardi** [comment]
@RajaMuhammadAwais, I need to think a little bit about these changes...

@utoni, what do you think, especially about using `memcpy` also on x86-64?

**Comment by RajaMuhammadAwais** [comment]
Thanks @IvanNardi I understand the concern about applying memcpy-based loads on x86-64 as well. I checked the generated code with GCC -O2 on x86-64: the fixed-size 16-bit and 32-bit memcpy helpers are fully inlined as a single movzx/mov load, with no library call, so the generated code is equivalent to the direct load in this configuration. The main reason for using memcpy is to preserve defined C behavior for unaligned packet and serialized data, especially on strict-alignment architectures.

I agree that the final decision should be based on an nDPI-specific benchmark rather than this small compiler check. I can add a benchmark comparison for the relevant DNS and serializer paths, or the change can be restricted if the project prefers architecture-specific accessors. I will not modify the PR until you and utoni agree on the preferred approach.

File: src/include/ndpi_typedefs.h [hunk 1]
```diff
@@ -1544,9 +1544,9 @@ typedef struct {
 #define ndpi_private_deserializer ndpi_private_serializer
 
 #ifdef NDPI_CFFI_PREPROCESSING
-typedef struct { char c[72]; } ndpi_serializer;
+typedef union { u_int64_t _alignment; char c[72]; } ndpi_serializer;
 #else
-typedef struct { char c[sizeof(ndpi_private_serializer)]; } ndpi_serializer;
+typedef union { ndpi_private_serializer _alignment; char c[sizeof(ndpi_private_serializer)]; } ndpi_serializer;
```
utoni, line 1556: Are you sure that this line enforces correct alignment?

utoni:
> @RajaMuhammadAwais, I need to think a little bit about these changes...
> 
> @utoni, what do you think, especially about using `memcpy` also on x86-64?

Fine, will be most likely optimized away on x64, and it's the right way to fix unaligned memory access

src/include/ndpi_typedefs.h
RajaMuhammadAwais: Yes, in the normal C build the union should enforce the required alignment because ndpi_private_serializer is a complete union member, so the union alignment is at least the alignment of that type. The char array preserves the existing storage size. I agree that this should be confirmed with sizeof/_Alignof checks on the supported architectures before finalizing the change. No CNAME or unrelated DNS logic is included in this PR.

utoni: Thanks, adding a static assertion seems like a good idea.

IvanNardi:
@RajaMuhammadAwais, could you take a look at the failures with mingw and window, please?

Anyway, this  PR will be merged **after** the next major release (expected this week)

src/include/ndpi_typedefs.h
utoni: @RajaMuhammadAwais 
Is there actually any reason why you did not choose to use the `alignas(N)` keyword?

RajaMuhammadAwais: @utoni  Thanks for the question. I considered alignas, but there are two separate alignment issues here.
For the opaque ndpi_serializer storage, the required alignment is type-dependent. The union includes ndpi_private_serializer as an alignment-bearing member, so the storage inherits the alignment of the actual implementation type without hard-coding an architecture-specific N. The character array preserves the existing storage size.
For the packet and serialized-data reads, however, alignas(N) would not solve the main problem. It could align a destination object, but it cannot change the alignment of an arbitrary source buffer when a field starts at an odd offset. A direct dereference of that source as u_int16_t, u_int32_t, or u_int64_t would still rely on invalid alignment.
That is why I used memcpy for the actual byte loads: it safely copies from a potentially unaligned source into an aligned local integer object while preserving the existing byte-order handling. In other words, the alignment-preserving union addresses the opaque serializer object, while memcpy addresses unaligned fields inside packet and serialized buffers.
If the project prefers explicit alignas or Alignas for the opaque storage and the supported C standard allows it, I can consider adding it together with sizeof and alignment assertions. It would complement the memcpy reads, but it would not replace them.

RajaMuhammadAwais:
@IvanNardi I investigated the three failing jobs and reproduced the exact TLS commands locally using a clean regenerated build. The parent commit produce identical results: tls-appdata.pcap reports 38 DPI packets and tls_verylong_certificate.pcap reports 32 DPI packets in both builds. The CI values of 86 and 48 therefore do not reproduce in a clean local PR build. I do not think the expected outputs should be updated. Could you please check whether the masan, s390x, and Ubuntu 26.04 jobs are reusing generated headers or stale build artifacts, or whether they are running a different build configuration? The source changes in this PR are limited to alignment-safe helpers and the CFFI serializer representation. No further code or test-output changes have been pushed.

src/include/ndpi_typedefs.h
utoni: Thanks, @RajaMuhammadAwais, I think it is important to document the alignment fix within opaque data structures. If possible via static assertions.

RajaMuhammadAwais: @utoni Thanks for the suggestion. I have documented the alignment contract for the opaque serializer storage and added portable static assertions to verify its size and alignment at compile time.

IvanNardi (approved):
LGTM

IvanNardi:
@utoni, are you fine with the final version?

utoni (approved):
LGTM

Issue #2575, opened 2024-09-30
DNS dissector do not proceed next paket when receice CNAME


didgaudin:
## Describe the bug
When we receive 2 responses packets, if the first is cname, et second who contain de A record is not processed, and then  flow->info do not contain protos.dns.rsp_addr.ipv4 value

### Obtained behavior
A clear and concise description of what happening.

## nDPI Environment (please complete the following information):
* OS name: fedora
* OS version: 41
* Architecture: x86_64
* nDPI version or commit hash: Using nDPI (4.11.0-4871-ddd08f9) [1 thread(s)]
* nDPI compilation flags used:--with-pcre2 --with-local-libgcrypt]
* Attach the `config.log` file generated after `./configure` ran (if you are building from source).

### Reproducible using ndpiReader?
- [x] The reported bug is reproducible using ndpiReader.
- [x] The reported bug is not reproducible using ndpiReader.

### If your bug is reproducible using a pcap, please attach a pcap file (or a valid link to download it)
[dns.pcap.gz](https://github.com/user-attachments/files/17187363/dns.pcap.gz)

Example: dns.pcap.gz

### Steps to reproduce the behavior:
1. Run '...'
2. Set '....'
3. Do '....'

IvanNardi:
I acknowledge the problem.
However it is a little bit more complex that "skip any answers up to A/AAAA record" because we must be sure to return request metadata and answer metadata of the SAME transaction.

PR #3232, opened 2026-08-23
dns: support an opt-in custom detection port

### LINKED ISSUE #1788: nDPI fails to recognize custom DNS ports
I am using DNS server on a custom port 1234. nDPI fails to recognize the packets as DNS due to hardcoded port 53 in detection.

I've found [this](https://github.com/ntop/nDPI/issues/725) issue, but it did solve the problem for the OpenDNS protocol exclusively. 
I understand that lack of port checking might create an overhead and it is quite unusual to use ports different than 53. It happens though.

It would be helpful if ndpi allowed specifying whether to use some particular ports for the sake of detection of DNS or just accept anything as a possible DNS packet with some optional compile time flag.

Let's say that adding -D FLEXIBLE_DNS_PORTS would allow to detect ports different than 53.

What do you think?

RajaMuhammadAwais:
## Related issue

Fixes #1788

## Summary

DNS detection is currently restricted to the standard DNS, mDNS, and LLMNR ports. This prevents nDPI from recognizing valid DNS traffic deployed on a user-selected nonstandard port. The behavior was confirmed in the issue discussion and reproduced with the official `dns_on_port_80.pcapng` capture.

This pull request adds an explicit, disabled-by-default configuration parameter that enables DNS detection on one additional exact TCP or UDP port.

## Configuration

```text
--cfg=dns,custom_port,80
```

The default value is `0`, which keeps existing behavior unchanged. The accepted range is `0` through `65535`. Only the configured port is added to the DNS entry-point gate. Standard DNS on port 53, mDNS on port 5353, and LLMNR on port 5355 retain their existing handling.

## Implementation details

The DNS port classification helper now receives the nDPI configuration context and returns `NDPI_PROTOCOL_DNS` only when the configured custom port matches either flow endpoint. Existing DNS structural validation remains mandatory, so enabling a custom port does not classify arbitrary traffic solely because it uses that port.

The configuration parameter is registered in `src/lib/ndpi_config.c`, stored in the private configuration structure, and documented in `doc/configuration_parameters.rst`.

## Evidence and regression testing

The official issue capture is included as the positive regression fixture:

```text
PCAP: tests/cfgs/dns_custom_port/pcap/dns_on_port_80.pcapng
Config: --cfg=dns,custom_port,80
Result: DNS, 2 packets, 1 flow
```

The same capture without the configuration remains `Unknown`, confirming that the option is disabled by default. Standard DNS traffic remains classified as DNS when the option is enabled. An unrelated HTTP capture with port 80 produced no DNS classification. An invalid value of `70000` is rejected by the configuration API.

The following checks passed locally:

```text
make -j2
./tests/unit/unit
Focused positive and negative ndpiReader checks
Standard DNS compatibility check
Port-80 HTTP false-positive check
Invalid configuration range check
Address, Undefined, and Leak Sanitizer build and fixture run
git diff --check
```

The sanitizer run completed without an AddressSanitizer, UndefinedBehaviorSanitizer, or LeakSanitizer report.

## Files changed

| File | Change |
|---|---|
| `src/lib/protocols/dns.c` | Add exact custom-port matching while preserving standard-port logic |
| `src/lib/ndpi_config.c` | Register `dns.custom_port` with default `0` and range `0` to `65535` |
| `src/include/ndpi_private.h` | Add the private configuration field |
| `doc/configuration_parameters.rst` | Document the parameter and its default behavior |
| `tests/cfgs/dns_custom_port/` | Add the official PCAP fixture, configuration, and expected output |

## Checklist

- [x] I have signed the ntop Contributor License Agreement.
- [x] I have read the nDPI contributing guidelines.
- [x] I have updated the documentation where applicable.
- [x] I have added a focused regression fixture.
- [x] I have kept this change isolated from PR #3231 and issue #2575.

## References

1. [nDPI issue #1788](https://github.com/ntop/nDPI/issues/1788)
2. [nDPI contributing guidelines](https://github.com/ntop/nDPI/blob/dev/CONTRIBUTING.md)
3. [Fork branch](https://github.com/RajaMuhammadAwais/nDPI/tree/fix/1788-custom-dns-port)


lucaderi:
Thank you

PR #3234, opened 2026-08-24
fix: exclude TCP retransmissions from reader goodput

### LINKED ISSUE #1853: Goodput Ratio Calculation Accuracy ?
## Describe the bug
Questioning how goodput ratio is calculated when pcap includes retransmissions.  

### Expected behavior
I would expect goodput is simply the cumulative non-zero payload packet size sent from A->B not counting retransmissions and duplicate packets.   
### Obtained behavior
Attached is a sample pcap showing how the client is retransmitting every packet.  My goodput calculation is different then what ndpIReader shows.  ndpiReader ratio calc is 65 out and 94 in.  My calc is 33 out and 94 in.  

* OS name: Ubuntu.
* OS version: 20.04
* Architecture: [amd
* nDPI version: nDPI 4.5.0-3751-2c3f56d4

``` shell
ndpiReader -v 2 -i  ~/projects/agentslim/pcaps/CF_TLS_21_2.pcap

[CF_TLS_21_2.zip](https://github.com/ntop/nDPI/files/10422479/CF_TLS_21_2.zip)

### PR DESCRIPTION
**Pull request by RajaMuhammadAwais**
## Related issue

Fixes #1853

## Problem

`ndpiReader` previously added every captured TCP payload byte to the Goodput counters. When a TCP segment retransmitted bytes already seen in the same direction, those bytes were counted again. This made the reported Goodput ratio depend on wire retransmissions instead of unique TCP sequence space.

The official issue reproducer contains a duplicated client segment. On the existing implementation, the client Goodput ratio was reported as `65/94`. Packet-level sequence analysis shows that the unique client payload is 759 bytes while the captured payload includes retransmitted bytes.

## Solution

This change adds per-direction TCP sequence interval tracking in `ndpiReader`. For each TCP payload segment, the implementation calculates the portion of the half-open sequence interval that has not already been observed in that direction. Only that uncovered portion is added to the Goodput counter.

The implementation handles exact duplicates, partial overlaps, out-of-order non-overlapping segments, SYN and FIN packets without payload, and zero-payload ACKs. TCP wire-byte counters, packet counters, DPI counters, protocol detection, and packet analysis behavior remain unchanged. The tracker is kept in the example reader and does not change the nDPI core API or protocol dissectors.

## Verification

The official #1853 PCAP now reports:

```text
Goodput ratio: 33/94
```

The result matches the unique payload accounting derived from the packet sequence numbers.

The complete repository regression result files were regenerated through the existing test harness update mechanism. A fixture audit covered 111 changed result blocks. The changes are limited to Goodput ratio values or are unchanged after comparison against the exact upstream base commit.

Validation completed:

* Native build passed.
* Unit tests passed.
* Official #1853 PCAP passed with the corrected `33/94` Goodput ratio.
* Targeted regression suite passed: 3 tests passed, 0 failed.
* `git diff --check` passed.
* No Markdown or research files are included.
* No changes were made to nDPI protocol dissectors or public library APIs.

## Files changed

* `example/reader_util.c`
* `example/reader_util.h`
* Existing tracked regression result files under `tests/cfgs/**/result/` to reflect the verified unique-payload Goodput semantics.

## Notes

The sequence interval state is maintained per flow and direction. Total captured-byte counters retain their previous wire-byte meaning. Only Goodput accounting excludes duplicate sequence coverage.

I have signed the nDPI Contributor License Agreement and have read the contributing guidelines.


IvanNardi:
Rebased

IvanNardi (approved):
Thank you!

PR #3240, opened 2026-08-31
dcerpc: reassemble fragmented TCP messages

### LINKED ISSUE #3048: Wrong classification of DCERPC PCAP data
## Describe the bug
NDP failed to decode DCERPC PCAP data

### Expected behavior
Classify the PCAP as DCERPC.
### Obtained behavior
Classify as Unknown 

* Ubuntu 22.04 
* OS22.04
* Architecture: amd64
* nDPI 4.14

./ndpiReader -i /home/ubuntu/dcerpc_drs_op.pcap

-----------------------------------------------------------
* NOTE: This is demo app to show *some* nDPI features.
* In this demo we have implemented only some basic features
* just to show you what you can do with the library. Feel
* free to extend it and send us the patches for inclusion
------------------------------------------------------------

Using nDPI (4.14.0-5235-f2a9087) [1 thread(s)]
Using libgcrypt version 1.8.6internal
Reading packets from pcap file /home/ubuntu/dcerpc_drs_op.pcap...
Running thread 0...

nDPI Memory statistics:
        nDPI Memory (once):      42.82 KB
        Flow Memory (per flow):  1.15 KB
        Actual Memory:           11.06 MB
        Peak Memory:             11.06 MB
        Setup Time:              40 msec
        Packet Processing Time:  0 msec

Traffic statistics:
        Ethernet bytes:        6161          (includes ethernet CRC/IFC/trailer)
        Discarded bytes:       0
        IP packets:            19            of 19 packets total
        IP bytes:              5705          (avg pkt size 300 bytes)
        Unique flows:          1
        TCP Packets:           19
        UDP Packets:           0
        VLAN Packets:          0
        MPLS Packets:          0
        PPPoE Packets:         0
        Fragmented Packets:    0
        Max Packet size:       1280
        Packet Len < 64:       5
        Packet Len 64-128:     1
        Packet Len 128-256:    9
        Packet Len 256-1024:   2
        Packet Len 1024-1500:  2
        Packet Len > 1500:     0
        nDPI throughput:       27.38 K pps / 67.73 Mb/sec
        Analysis begin:        20/Dec/2012 21:19:59
        Analysis end:          20/Dec/2012 21:20:00
        Traffic throughput:    41.77 pps / 105.81 Kb/sec
        Traffic duration:      0.455 sec
        DPI Packets (TCP):     16            (16.00 pkts/flow)
        Confidence: Unknown    1             (flows)

Detected protocols:
        Unknown              packets: 19            bytes: 5705          flows: 1

Protocol statistics:
        Unrated              packets: 19            bytes: 5705          flows: 1

Risk stats [found 1 (100.0 %) flows with risks]:
        Susp Entropy                                 1 [100.0 %]

        NOTE: as one flow can have multiple risks set, the sum of the
              last column can exceed the number of flows with risks.

Please remove the txt extension to get the PCAP

[dcerpc_drs_oppcap.txt](https://github.com/user-attachments/files/23853909/dcerpc_drs_oppcap.txt)

RajaMuhammadAwais:
Fixes the TCP-layer fragmentation reported in ntop/nDPI#3048 using a bounded per-direction buffer and add the issue PCAP as a regression test.

Link to the related Fixes #3048 

Describe changes:
# Fix DCERPC detection for fragmented TCP messages

## Problem

The DCERPC dissector currently validates connection-oriented DCERPC only when the complete message is available in a single TCP payload. When a valid DCERPC fragment is split across multiple TCP segments, the dissector evaluates each segment independently. The first segment does not contain a complete DCERPC PDU, so the dissector is excluded before the remaining segments can be inspected.

As a result, the DCERPC flow from the issue-provided capture is reported as `Unknown` instead of `DCERPC`.

The attached issue capture demonstrates a valid DCERPC message distributed across multiple TCP packets, including the DCERPC header and body being received in separate segments.

## Root Cause

The previous implementation had no TCP reassembly state for DCERPC and excluded the dissector when an individual payload did not satisfy the complete-message checks.

The existing connection-oriented validation also expects the payload length to match the DCERPC fragment length exactly, which is not true for an incomplete TCP segment.

## Solution

This change adds a bounded, per-direction TCP reassembly path for connection-oriented DCERPC messages:

* Maintains separate reassembly state for each TCP direction.
* Buffers candidate DCERPC TCP payloads until the minimum header fields are available.
* Reads the DCERPC fragment length using the protocol byte-order flag.
* Validates the fragment length before allocating or extending the buffer.
* Limits the reassembly buffer to 65,535 bytes.
* Passes the reconstructed complete PDU through the existing DCERPC classifier.
* Frees all reassembly buffers during normal flow cleanup.
* Preserves the existing UDP connectionless and complete single-packet TCP detection paths.
* Avoids treating TCP retransmissions as new fragment data.

## Regression Test

The PCAP supplied in issue #3048 was added as a regression fixture:

```text
tests/cfgs/default/pcap/dcerpc-fragmented.pcap
tests/cfgs/default/result/dcerpc-fragmented.pcap.out
```

The fixture verifies that the fragmented flow is detected as `DCERPC` instead of remaining `Unknown`.

## Verification

The following checks were completed:

* Issue #3048 fragmented PCAP: detected as `DCERPC` with 19 packets and 1 flow.
* Existing `dcerpc.pcap` fixture: passes with no output difference.
* New regression fixture using the same command format as `tests/do.sh`: passes with no output difference.
* `./tests/do-unit.sh`: passes.
* `./example/ndpiReader --run-tests`: exits successfully.
* `git diff --check`: passes.
* Project compilation with `make`: passes.

## Scope

This is a focused fix for DCERPC connection-oriented messages fragmented across TCP segments.

It does not change the existing connectionless UDP detection logic or the normal complete-PDU TCP fast path.

## Related Issue

Closes #3048


IvanNardi:
@RajaMuhammadAwais, could you rebase please?

src/lib/ndpi_main.c
```diff
@@ -8225,6 +8225,27 @@ void ndpi_free_flow_data(struct ndpi_flow_struct* flow) {
...
+
+      for(i = 0; i < 2; i++) {
+        if(flow->core.dns_tcp_reasm->dir[i].buf)
+          ndpi_free(flow->core.dns_tcp_reasm->dir[i].buf);
+      }
+
+      ndpi_free(flow->core.dns_tcp_reasm);
+      flow->core.dns_tcp_reasm = NULL;
+    }
+
```
IvanNardi: Same code block twice. Remove it

src/lib/protocols/dcerpc.c
```diff
@@ -89,7 +199,15 @@ static void ndpi_search_dcerpc(struct ndpi_detection_module_struct *ndpi_struct,
...
   }
 
-  if(packet->payload_packet_len>1)
+  if(packet->tcp != NULL && (flow->core.dcerpc_tcp_reasm != NULL || dcerpc_tcp_start_candidate(packet))) {
+    if(dcerpc_tcp_process(ndpi_struct, flow) >= 0)
+      return;
+  }
+
+  /* The issue capture has a short ACK zero-padding segment before DCERPC. */
+  if(dcerpc_tcp_ack_padding(packet))
```
IvanNardi: could you elaborate about what this function does and why we need it, please?
Keep in mind that in this dissector (with TCP packets):
* packet->tcp_retransmission is always false
* packet->payload_packet_len is always > 0

IvanNardi (changes requested):

src/lib/protocols/dcerpc.c
RajaMuhammadAwais: This helper handles a capture-specific TCP segment that appears immediately before the first DCERPC fragment.

In the issue capture, packet #3 has ACK set, PSH unset, a 6-byte payload, and the payload is entirely zero:
00 00 00 00 00 00

The following packet starts the fragmented DCERPC message. This zero-filled segment is not a DCERPC PDU, so it must not be appended to the DCERPC reassembly buffer. We simply keep the DCERPC dissector active; otherwise this segment could cause the dissector to be excluded before the actual DCERPC fragment arrives. The same 6-byte zero-filled ACK segment also appears at the end of the capture.

The helper therefore recognizes an ACK segment without PSH, with a 2–7 byte payload, where every payload byte is zero. The lower bound matches nDPI’s existing TCP ACK-padding heuristic and avoids treating a one-byte TCP keep-alive payload as DCERPC padding.

`packet->tcp_retransmission` and `packet->payload_packet_len > 0` do not need to be checked here because the TCP dissector path guarantees those invariants for packets reaching this dissector. The `packet->tcp == NULL` check is retained because the DCERPC dissector is also registered for UDP.

IvanNardi: > In the issue capture, packet [#3](https://github.com/ntop/nDPI/pull/3) has ACK set, PSH unset, a 6-byte payload, and the payload is entirely zero:
> 00 00 00 00 00 00

I don't get it, packet 3 has packet->payload_packet_len = 0; it is not processed by any dissector. Am I missing something obvious?

In other word, is `dcerpc_tcp_ack_padding` really needed?

RajaMuhammadAwais:
DONE

IvanNardi (changes requested):

RajaMuhammadAwais:
@IvanNardi You are right. I misinterpreted the packet path in the capture.
The ACK zero-padding packet is handled before the protocol dissectors by nDPI's TCP processing. By the time the DCERPC dissector is considered, that packet has packet->payload_packet_len == 0 / is excluded by the TCP selection conditions, so dcerpc_tcp_ack_padding() cannot and does not need to handle it.
Therefore, dcerpc_tcp_ack_padding() is unnecessary for this fix. I will remove the helper and its call. The actual fix is the TCP reassembly of the DCERPC PDU split across packets 4, 5, and 7.

src/lib/protocols/dcerpc.c
```diff
@@ -78,6 +82,98 @@ static bool is_connectionless_dcerpc(struct ndpi_packet_struct *packet)
...
+                              struct ndpi_flow_struct *flow)
+{
+  struct ndpi_packet_struct *packet = &ndpi_struct->packet;
+  struct ndpi_dcerpc_tcp_reasm *reasm;
+  const u_int8_t *original_payload = packet->payload;
+  u_int16_t original_len = packet->payload_packet_len;
+  u_int32_t seq = ntohl(packet->tcp->seq);
+  u_int16_t fragment_len;
+
+  if(flow->core.dcerpc_tcp_reasm == NULL) {
```
IvanNardi, line 138: The general idea is fine, now, but we need to be smarter here, and allocate all those structures only if really needed.
For example: we should check if we already have a complete message: in that (common
) case we don't need the reassembler and we shouldn't allocate anything.
In case of a real fragment, we also should check if the "part of the message that we already have" is valid before allocating anything
This is the same logic we recently used for DNS reassembling

IvanNardi (changes requested):

src/lib/protocols/dcerpc.c
RajaMuhammadAwais: Updated accordingly.

The TCP reassembly path now follows the same allocation strategy as the recent DNS reassembly changes:

- A complete DCERPC PDU is handled by the existing fast path, so no reassembly state is allocated.
- Before allocating reassembly state, the available DCERPC header prefix is validated.
- If the payload contains a complete header, the declared fragment length is checked first; reassembly is allocated only when the declared PDU is larger than the current TCP payload.
- Invalid header prefixes and invalid fragment lengths are rejected before allocation.
- Subsequent fragments are still appended only when their TCP sequence number matches the expected next sequence number.

IvanNardi (approved):
Thank you!

We do need a generic reassembler code, but this PR seems fine

PR #3254, opened 2026-09-09
fix: ndpi_network_ptree6_match uses v6 maxbits for prefix fill

### LINKED ISSUE #3253: ndpi_network_ptree6_match uses max_bits from v4 ptree
## Describe the bug
I was looking at the 5.0 branch to bring in the patch for CVE-2026-86098 and I noticed a problem in another commit delivered after 5.0. Commit 70ca5b82b1 ("Typo fix") fixes a problem in ndpi_network_ptree6_match where it access the v4 ptrees instead of v6, it fixed the call to ndpi_patricia_search_best but not ndpi_fill_prefix_v6. This is still the case in the dev branch.

This call to ndpi_fill_prefix_v6 will always return -1 (bits > maxbits) because bits is 128 and ndpi_str->protocols->v4->maxbits is 32. 

### Expected behavior

IPv6 addresses should always be matched properly.

### Obtained behavior

I wrote a test in tests/unit/unit.c.
```
networkPtreeUnitTest: ERROR: 2606:4700:100::1: expected proto 300, got 0 — exit 255
```

Fixing the other case of v4 to v6, it now passes
```
networkPtreeUnitTest OK — exit 0
```

Other test suites do not seem to detect the problem. ndpi_internal_detection_process_packet will eventually call ndpi_network_port_ptree6_match which uses the correct ptrees.

* OS name: Ubuntu
* OS version: 22.04.5
* Architecture: x86
* nDPI version or commit hash: 5.0-stable (e7b413ace) AND dev (06bd2303)
* nDPI compilation flags used: if you are building from source [e.g. --with-pcre2 --with-local-libgcrypt].
* Attach the `config.log` file generated after `./configure` ran (if you are building from source).

## How to reproduce the reported bug

Unsure. ndpi_network_ptree6_match is only called from LLVMFuzzerTestOneInput and the return proto is not checked. The fuzztests currently all pass.

- [x] The reported bug is not reproducible using ndpiReader.

## Additional context
I'll have a pull request for this soon.

wahajahmed010:
## Summary

 passes  to  for a 128-bit IPv6 prefix. The function rejects this with  because , so the subsequent  against the v6 tree never runs against a valid prefix and the lookup always misses.

Commit 5d941ca7 ("Typo fix") already corrected the  argument to use , but missed the matching cast in the  argument right above it.

## Fix

Use  so the prefix is filled, matching the symmetry already present in  (which uses  with ).

## Reproduction

Per #3253,  reports:

because  returns  (bits=128 > maxbits=32) before the search runs.

## Checklist

- [x] I have signed the ntop Contributor License Agreement
- [x] I have read the contributing guidelines

Closes #3253

### DIFF
--- a/src/lib/ndpi_main.c
+++ b/src/lib/ndpi_main.c
@@ -3345,7 +3345,7 @@ u_int16_t ndpi_network_ptree6_match(struct ndpi_detection_module_struct *ndpi_st
 
   /* Make sure all in network byte order otherwise compares wont work */
   ndpi_fill_prefix_v6(&prefix, pin, 128,
-		      ((ndpi_patricia_tree_t *) ndpi_str->protocols->v4)->maxbits);
+		      ((ndpi_patricia_tree_t *) ndpi_str->protocols->v6)->maxbits);
   node = ndpi_patricia_search_best(ndpi_str->protocols->v6, &prefix);
 
   return(node ? node->value.u.uv16[0].user_value : NDPI_PROTOCOL_UNKNOWN);


IvanNardi:
Fixed via another PR. Thank you anyway for your PR

